PR #7, opened 2019-11-15
fixed #3 (three-argument project_point!)

### LINKED ISSUE #3: project_point!
At the moment `project_point!` is a little out of line because it operates on its only argument in-place while other functions generally assume that mutated output is separate. Should we change it? Or maybe provide both variants?

mateuszbaran:
This is the change discussed in #3.


kellertuer:
Looks good. If this one is merged, I will propose a new PR after this one, removing Euclidean and adding more tests (wrote a few in the old one already).

PR #8, opened 2019-11-16
Delete Euclidean

kellertuer:
This PR deletes the Euclidean Manifold and adds a (more minimal) Default Manifold, that is as short as possible, such that a new user can get started.

I also extended the tests to also cover geodesics.

I noticed, that the non mutating `exp`, `log`,... face the challenge, that `similar_result` will not work for the `Point`/`TVector` subtypes (see extended tests for the `empty_manifold`. Any ideas how we could unify array and typed approaches to manifolds with respect to the `similar_result` function?


README.md
```diff
@@ -4,4 +4,8 @@
 
 Basic interface for manifolds in Julia.
 
-The main part of the project is [`Manifolds.jl`](https://github.com/JuliaNLSolvers/Manifolds.jl).
+This interface includes an easy `DefaultManifold`, which is a reduced version
```
mateuszbaran:
```suggestion
This interface includes a simple `DefaultManifold`, which is a reduced version
```

src/DefaultManifold.jl
```diff
@@ -0,0 +1,25 @@
+struct DefaultManifold{T<:Tuple} <: Manifold where {T} end
```
mateuszbaran: It would be a good idea to add a docstring to `DefaultManifold`.

src/ManifoldsBase.jl
```diff
@@ -384,69 +384,110 @@ function shortest_geodesic(M::Manifold, x, y, T::AbstractVector)
...
+struct ParallelTransport <: AbstractVectorTransportMethod end
+
+"""
+    ProjectTangent <: AbstractVectorTransportMethod
+
+Specify to use projection onto tangent space as vector transport method within
+[`vector_transport_to`](@ref), [`vector_transport_direction`](@ref) or
+[`vector_transport_along`](@ref). See [`project_tangent`](@ref) for details.
+"""
+struct ProjectTangent <: AbstractVectorTransportMethod end
```
mateuszbaran: I don't really like this name. Firstly it does not specify that it's a transport method and secondly, the same struct could in the future be used for projection-based transport of cotangent vectors (for example). Maybe `ProjectionTransport`, `ProjectTransport` or `ProjectionVectorTransport`?

test/empty_manifold.jl
```diff
@@ -23,42 +28,42 @@ struct NonManifold <: Manifold end
 
     exp_retr = ManifoldsBase.ExponentialRetraction()
 
-    @test_throws ErrorException retract!(m, [0], [0], [0])
```
mateuszbaran: Why did you replace these tests with `MPoint`-based ones instead of keeping both?

src/ManifoldsBase.jl
kellertuer: Oh, good remark. I think I am in favour of `ProjectionTransport`, that would then also apply for `CoTVectors`.

kellertuer:
> `similar_result` should work as long as `MPoint`/`TVector` subtypes have `eltype` and `similar` defined (which I think is a good idea anyway).
Ah, yes, you're right, at least if one encapsulates properly, I think, so

````
struct MyMPoint <: MPoint
    data::Vector{Float64}
end
````
woudl require and additional line `eltype(::MyPoint) = Float64` for example, and similar would return a similar `MyPoint` then, right? That is a good idea, meybe to be even included into the `DefaultManifold` as an illustration.

File: test/empty_manifold.jl (hunk 4 above)
Comment (kellertuer) [reply test/empty_manifold.jl]: Oh, indeed, one can even keep both – also a good idea. There was no specific reason to “delete”, I merely edited them. I will bring the old ones back, too.

**Comment by mateuszbaran** [comment]
Yes, a full example would be something like
```julia
struct MyMPoint{T} <: MPoint
    data::Vector{T}
end
Base.eltype(p::MyMPoint) = eltype(p.data)
Base.similar(p::MyMPoint, ::Type{EltypeNew}) where EltypeNew = MyMPoint(similar(p.data, EltypeNew))
Base.similar(p::MyMPoint) = MyMPoint(similar(p.data))
```

It could definitely be included in `DefaultManifold`.

kellertuer:
We could maybe still extend the example on the new Types by test cases. And elaborate on that in the documentation later on, how such types are meant to be used. I only fear that the inplace methods might not work, since `.=` does not work for such types? Or can that be “passed down” to the `.data` field?

kellertuer:
I think when a few tests for the typed data, are finished, we can merge this.

mateuszbaran:
> We could maybe still extend the example on the new Types by test cases. And elaborate on that in the documentation later on, how such types are meant to be used. I only fear that the inplace methods might not work, since `.=` does not work for such types? Or can that be “passed down” to the `.data` field?

This is a bit complicated. On one hand, broadcasting can be passed into the `.data` field without much work in typical cases. The `ArrayAndChar` example from [documentation](https://docs.julialang.org/en/v1/manual/interfaces/) shows how to do it. On the other hand, sometimes this won't work (when the internal representation is more complex). Another thing is that Zygote compatibility https://github.com/JuliaNLSolvers/Manifolds.jl/issues/42 essentially requires a separate implementation of non-mutating operations.


> I think when a few tests for the typed data, are finished, we can merge this.

Sure. So, you want to add broadcasting support for typed data? I think we can't generally expect that all typed data will be broadcastable. `copyto!` should _usually_ work though. You can also take a look at `ProductArray` and `ProductRepr` from Manifolds.jl: https://github.com/JuliaNLSolvers/Manifolds.jl/blob/master/src/ProductRepresentations.jl

kellertuer:
For wuite some cases I am fine with another implementation. One thing I want to try is, provide types for SPD, that store U,S,V from the SVD and check whether that can be faster. So with that I am not saying we should _in general_ pass down broadcasting (and I haven't read much about Zygote yet), I was just wondering, what the nicest way is to show the user of this interface in the `DefaultManifold` how the typed data should be used.

I am not sure whether I want to add broadcasting for type data in general, though. I was just thinking about – as I said – what the nicest way is to equip the small example here with typed functions (exp,log,distance,norm,inner, and vector_transport I think?); and shortly document that Idea (in the readme as long as our doc is not that long)

mateuszbaran:
I'm not sure if there is one correct way of using typed points and vectors. `Manifolds.jl` has `ArrayManifold`, `FVector`, `ProductRepr` and `ProductArray`, each one addresses a different problem. Maybe it would be enough to refer readers to these examples and describe their design? I could make such description. Probably `ArrayManifold` is the closes to what you want for `DefaultManifold`?



> One thing I want to try is, provide types for SPD, that store U,S,V from the SVD and check whether that can be faster.

This will heavily depend on the size of matrices. Up to 3x3 I wouldn't expect much improvement in most cases since `StaticArrays` has really fast eigendecomposition of such small matrices. For larger matrices is seems like a good idea though.

kellertuer:
I might have been a little tired tomorrow; actually what I had in mind for `MatrixManifold`is, what `ArrayManifold` already does and `DefaultManifold` just would have to subtype from that. So then … do we move `ArrayManifold` here or do we keep it in `Manifolds`? I am not yet completely sure how much the interface should take over at that point.

Concerning SPDs – yes for 3x3 I also don't expect much of a gain.

mateuszbaran:
> So then … do we move `ArrayManifold` here or do we keep it in `Manifolds`? I am not yet completely sure how much the interface should take over at that point.

I think we can keep `ArrayManifold` in `Manifolds.jl` then. At least for now.

kellertuer:
I just shortly checked locally – extending `DefaultManifold` to be similar to an `ArrayManifold` means copying a lot of stuff from that while copying `ArrayManifold` reduces the `DefaultManifold` even further (no need for the point/tangent/cotangent nor the similar/copy functions). That's why I would opt for transferring `ArrayManifold` but I'll wait until we have discussed that in #10.

kellertuer:
With the addition of `ArrayManifold` to illustrate semantic types (`MPoint`, `TVector`, and `CoTVector`) I think this PR is finished.

mateuszbaran:
Great, I think this PR looks fine except `ArrayManifold` should be updated to the new vector transport interface.

kellertuer:
Good observation, added the method, added the `along` variant and noticed the non-mutating one could be omitted since it should work directly with the original interface.

kellertuer:
I carefully implemented more tests for the ArrayManifold but _now_ I really think this PR is finished and I would like to merge, let's say tomorrow.

PR #11, opened 2019-11-19
Implements a THTT for decorator manifolds.

kellertuer:
This PR introduces a THTT for decorator manifolds.

### DIFF
--- a/src/ManifoldsBase.jl
+++ b/src/ManifoldsBase.jl
@@ -64,37 +64,52 @@ representations of cotangent vectors and their types on a manifold.
 abstract type CoTVector end
 
 """
+    is_decorator_manifold(M)
+
+indicate whether a manifold is a decorator manifold, i.e. whether it encapsulates
+a [`Manifold`](@ref) with additional features, and stores internally the original
+manifold instance. An example is the [`ArrayManifold`](@ref).
+
+Using Tim Holys Traits Trick (THTT), certain functions are just calling themselves
+on the internal manifold, and hence have not to be implemented for decorators
+again, for example [`manifold_dimension`](@ref) and especially [`base_manifold`](@ref).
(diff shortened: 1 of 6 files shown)


src/ManifoldsBase.jl
```diff
@@ -64,37 +64,52 @@ representations of cotangent vectors and their types on a manifold.
...
+    is_decorator_manifold(M)
+
+indicate whether a manifold is a decorator manifold, i.e. whether it encapsulates
+a [`Manifold`](@ref) with additional features, and stores internally the original
+manifold instance. An example is the [`ArrayManifold`](@ref).
+
+Using Tim Holys Traits Trick (THTT), certain functions are just calling themselves
+on the internal manifold, and hence have not to be implemented for decorators
+again, for example [`manifold_dimension`](@ref) and especially [`base_manifold`](@ref).
+This also assumes, that the undecoratord (base) manifold is stored in `M.manifold`.
```
mateuszbaran: Perhaps it could be 'decorated' instead of 'undecoratord'?

kellertuer: Or one could leave that just out, because within might be a decorated or a non decorated manifold; we might encapsulate quite a few decorators.

kellertuer:
I will add the remaining tests this evening, I already saw what's todo.

src/ManifoldsBase.jl
mateuszbaran: Yes, that's also an option. By 'decorated' I meant earlier the encapsulated manifold but it can be somewhat confusing.

mateuszbaran:
With this merged, I think we could register `ManifoldsBase.jl` and update `Manifolds.jl` to use this interface, right?

kellertuer:
Yes, we could register this one after this PR.

Transferring `Manifolds.jl` would also mean to adapt documentation and stuff, but yes we could then start that in a PR over there.

PR #16, opened 2019-12-21
Update docstrings, format code, and a few additions

sethaxen:
This PR makes mostly formatting changes to docstrings and code:
- Enforce emerging conventions for docstrings and variable names and reduce signature size
- Fix various typos
- Add missing docstrings
- Run [JuliaFormatter](https://github.com/domluna/JuliaFormatter.jl). This to ensure that consistent style is used across the package. It's not the only way, but it works quite well.

It also makes a few content changes/additions:
- https://github.com/JuliaNLSolvers/ManifoldsBase.jl/commit/b43bd93614573d6b8e958a02cb8c0e32bcb612f7 adds a new fallback from local injectivity radius to global. Since the global is the infimum of the local, this is a good conservative choice
- https://github.com/JuliaNLSolvers/ManifoldsBase.jl/pull/16/commits/82fed9947288679a15d0b7f8c95cecc073a228d3 changes default `injectivity_radius` to error instead of return `Inf`
- https://github.com/JuliaNLSolvers/ManifoldsBase.jl/commit/2225974984521d08f5ca33fb20a268e7ba237f5f replaces some broadcast copies with `copyto!`. The performance of these two patterns is the same, but it's more explicit, and since broadcasting in Julia is complicated, it's more likely to play with AD. (e.g. `Zygote.Buffer` supports `copyto!` but not broadcast).
- https://github.com/JuliaNLSolvers/ManifoldsBase.jl/commit/cecf019d421fc759233023a14813bacb733e3ef6 adds `AbstractEstimationMethod` from Manifolds.


src/ManifoldsBase.jl
```diff
@@ -1,89 +1,74 @@
...
 using LinearAlgebra
 
 """
     Manifold
 
-A manifold type. The `Manifold` is used to dispatch to different exponential
-and logarithmic maps as well as other function on manifold.
+A manifold type. The `Manifold` is used to dispatch to different
+[`exp`](@ref)onential and [`log`](@ref)arithmic maps as well as other functions
+on the manifold.
```
kellertuer: I think it would be nicer to turn this sentence around:

```
A manifold the. The `Manifold` is used to dispatch to different functions on a manifold.
It should usually be the first argument. Examples are the , [`exp`](@ref)onential and
[`log`](@ref)arithmic maps as well as more general functions that are built on them
like the [`geodesic`](@ref).
```

src/ManifoldsBase.jl
```diff
@@ -511,41 +575,53 @@ end
 @doc doc"""
     injectivity_radius(M::Manifold, x)
 
-Distance $d$ such that `exp(M, x, v)` is injective for all tangent
-vectors shorter than $d$ (has a left inverse).
+Distance $d$ such that [`exp(M, x, v)`](@ref exp(::Manifold, ::Any, ::Any)) is
+injective for all tangent vectors shorter than $d$ (i.e. has a left inverse).
 """
-injectivity_radius(M::Manifold, x) = Inf
+injectivity_radius(M::Manifold, x) = injectivity_radius(M)
```
kellertuer: I like this conservative choice.

src/ManifoldsBase.jl
```diff
@@ -511,41 +575,53 @@ end
...
 injectivity_radius(M::Manifold, x, ::AbstractRetractionMethod) = injectivity_radius(M, x)
 
 """
     injectivity_radius(M::Manifold)
 
-Infimum of the injectivity radii of all manifold points.
+Infimum of the [`injectivity_radius`](@ref injectivity_radius(::Manifold, ::Any))
+of all manifold points.
 """
 injectivity_radius(M::Manifold) = Inf
```
kellertuer: If the above local version defaults to this conservative choice, this should not be `Inf` but return an error, that its not implemented. `Inf` is far too large (for the `Sphere` for example).

kellertuer:
Since your at a meta review, could you try to include also tests for the last 3 untested lines?

src/ManifoldsBase.jl
sethaxen: Good point, I'll change.

sethaxen: Technically, that's a breaking change, so per semver we'll need to bump the version number to 0.3.

sethaxen:
Which 3 lines?

kellertuer:
See https://codecov.io/gh/JuliaNLSolvers/ManifoldsBase.jl/src/docsfixes/src/ManifoldsBase.jl, its the lines 75, 83 & 95, fallbacks to base or nonfallbacks (for false) for `base_manifold`, `represenation_size` as well as `manifold_dimension`.

src/ManifoldsBase.jl
kellertuer: Yes, well. But it's important, I think, since `Inf` might really have unwanted side effects.

sethaxen:
Weird, each of those already has a corresponding (indirect) test. I'm not sure why codecov doesn't count them, but I've seen it miss tests before.

src/ManifoldsBase.jl
sethaxen: Agreed

sethaxen:
No problem! I know this PR gets really into the details, but since Manifolds and ManifoldsBase in particular are base packages, I believe they should be easily readable and understandable as much as that's possible. For me, implementing these packages has been very useful in understanding manifolds, and my hope is that others find it educational by reading the docs and even moreso the code.

Btw, I'm roughly following the [Invenia BlueStyle](https://github.com/invenia/BlueStyle) conventions.

Once appveyor passes, I'll merge and tag a new release. There's a much larger companion PR I'm working on for Manifolds that I'll push up with a compat lower bound once this is released.

kellertuer:
Actually, when I started my first package – one of my goals was also to provide a good documentation for people to get started, so yes that should be readable and consistent. I'll try to follow that Style then now, too.

sethaxen:
I appreciate how detailed Manopt's docs are, and I've consulted them several times. 🙂

kellertuer:
I adopted that from my approaches in MVIRT, also because I can look up quite a few formulae when I need them :)

sethaxen:
I think Appveyor may have bugged out here. Usually it cancels all previous commits in the queue. This just shows all the most recent pushes as queueing, but nothing is running. Going to try to jump start it with a close and reopen.

PR #18, opened 2019-12-28
Resolve ArrayManifold ambiguities and fix functions for complex manifolds

sethaxen:
I used [Aqua.jl](https://github.com/tkf/Aqua.jl) to test for method ambiguities (we should probably do this for `Manifolds.jl` as well). It found that `vector_transport_to!` and `injectivity_radius` had several ambiguities that were missed by our tests, so I added failing tests and resolutions.

Several of our defaults did not work correctly for complex manifolds. The Hermitian positive definite inner product on a complex manifold produces a complex number whose real part is a Riemannian inner product on the underlying real manifold (i.e. by identifying `ℂⁿ`  with `ℝ²ⁿ`) and whose imaginary part is a complex differential form.

As I understand it, the concepts of norm, distance, and angle only apply by considering the real part, i.e. the part that arises due to the Riemannian metric. `norm` should always produce a complex number whose imaginary part is 0, so this just discards the imaginary part. `angle` requires discarding the imaginary part, which may be non-zero.

### DIFF
--- a/src/ArrayManifold.jl
+++ b/src/ArrayManifold.jl
@@ -218,6 +218,29 @@ function vector_transport_to!(
     return vto
 end
 
+function vector_transport_to!(
+    M::ArrayManifold,
+    vto,
+    x,
+    v,
+    y,
+    m::ProjectionTransport;
+    kwargs...,
+)
+    is_manifold_point(M, y, true; kwargs...)
+    is_tangent_vector(M, x, v, true; kwargs...)
+    vector_transport_to!(
+        M.manifold,
+        array_value(vto),
+        array_value(x),
+        array_value(v),
+        array_value(y),
+        m,
+    )
+    is_tangent_vector(M, y, vto, true; kwargs...)
+    return vto
+end
+
 function vector_transport_along!(
     M::ArrayManifold,
     vto,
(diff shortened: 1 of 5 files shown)


src/ArrayManifold.jl
```diff
@@ -218,6 +218,29 @@ function vector_transport_to!(
...
+        M.manifold,
+        array_value(vto),
+        array_value(x),
+        array_value(v),
+        array_value(y),
+        m,
+    )
+    is_tangent_vector(M, y, vto, true; kwargs...)
+    return vto
+end
```
kellertuer: An and even, this also adds the tests that we document `ArrayManifold` does (but did not do until now), so also semantically I think this is a good idea.

src/ArrayManifold.jl
```diff
@@ -241,9 +264,28 @@ function vector_transport_along!(
...
+    is_manifold_point(M, x, true; kwargs...)
+    return injectivity_radius(M.manifold, array_value(x))
+end
+function injectivity_radius(
+    M::ArrayManifold,
+    x,
+    method::AbstractRetractionMethod;
+    kwargs...,
+)
+    is_manifold_point(M, x, true; kwargs...)
```
kellertuer: same as above, good point and notice.

sethaxen:
> Thanks also for noticing the flaw in the complex case, we should check that with `Euclidean`, too. You comments are right, especially for the `inner` I noticed the same on the complex circle already.

I checked `Euclidean`, and with these fixes, everything should be good, since its `norm` defaults to `LinearAlgebra.norm` and it doesn't overload `Base.angle`.

But on that note, we need to make sure our complex manifolds consistently interpret the inner product. Here I've adopted the usual interpretation of a Hermitian inner product, which `Euclidean` and I think `FixedRankMatrices` do as well, but `Circle` and `Grassmann` with complex type both constrain the inner product to be real, i.e. they only return the Riemannian part. I propose that unless those manifolds are interpreted as real manifolds with complex representations (as opposed to complex manifolds), then they should return the full complex inner product. For `Circle`, this means replacing `complex_dot` with `dot`. If one really just wants just the real inner product, perhaps we need another method to distinguish the two.

kellertuer:
The `Circle` is really just a real manifold with a complex representation. It's one-dimensional and you can think of the compley representationas if it's `Sphrer(1)`int the Gaussian plane. So it's just Grassmann we have to check.

sethaxen:
Also `Stiefel`, `FixedRankMatrices`, and `Symmetric`. `Symmetric` looks fine. I don't know enough about `FixedRankMatrices` to tell. `Stiefel` and `Grassmann` both use the metric of the embedding `g(v,w) = dot(v, w)`. But they currently only return the real part, so that needs to be fixed.

kellertuer:
I can check `FixedRank` the coming days, I think it might be similar to `Circle`, i.e. a complex representation of a real-valued manifold. `Stiefel` and `Grassmann` should then be fixed, yes.

sethaxen:
Okay, I'll submit a PR for `Stiefel` and `Grassmann` tomorrow.

kellertuer:
I checked `FixedRank`, it also falls back to `dot` and should hence be fine, so with your PR this should be resolved.

PR #19, opened 2020-01-03
initial ONB implementation

mateuszbaran:
I've implemented the key parts of ONBs discussed in https://github.com/JuliaNLSolvers/Manifolds.jl/issues/66 . I'll probably add `ProjectedONB` here as well, and then make a PR to `Manifolds.jl` to add the ONB functionality.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "ManifoldsBase"
 uuid = "3362f125-f0bb-47a3-aa74-596ffd7ef2fb"
 authors = ["Seth Axen <seth.axen@gmail.com>", "Mateusz Baran <mateuszbaran89@gmail.com>", "Ronny Bergmann <manopt@ronnybergmann.net>", "Antoine Levitt <antoine.levitt@gmail.com>"]
-version = "0.3.2"
+version = "0.3.3"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
--- a/src/ArrayManifold.jl
+++ b/src/ArrayManifold.jl
@@ -263,6 +263,23 @@ function vector_transport_along!(
     return vto
 end
 
+function represent_in_basis(M::ArrayManifold, x, v, B::AbstractBasis; kwargs...)
+    is_tangent_vector(M, x, v, true; kwargs...)
+    return represent_in_basis(M.manifold, x, v, B)
+end
+
+function inverse_represent_in_basis(M::ArrayManifold, x, v, B::AbstractBasis; kwargs...)
+    is_manifold_point(M, x; kwargs...)
+    vout = inverse_represent_in_basis(M.manifold, x, v, B)
+    is_tangent_vector(M, x, v, true; kwargs...)
+    return vout
+end
+
+function basis(M::ArrayManifold, x, B::AbstractBasis; kwargs...)
+    is_manifold_point(M, x; kwargs...)
+    return basis(M.manifold, x, B)
+end
+
 injectivity_radius(M::ArrayManifold) = injectivity_radius(M.manifold)
 function injectivity_radius(M::ArrayManifold, method::AbstractRetractionMethod)
     return injectivity_radius(M.manifold, method)
--- a/src/DefaultManifold.jl
+++ b/src/DefaultManifold.jl
@@ -23,6 +23,24 @@ zero_tangent_vector!(::DefaultManifold, v, x) = fill!(v, 0)
 project_point!(::DefaultManifold, y, x) = copyto!(y, x)
 project_tangent!(::DefaultManifold, w, x, v) = copyto!(w, v)
 function vector_transport_to!(::DefaultManifold, vto, x, v, y, ::ParallelTransport)
-    return copyto!(vto, v)
+    copyto!(vto, v)
+    return vto
 end
+
+function represent_in_basis(M::DefaultManifold, x, v, ::ArbitraryONB)
+    return reshape(v, manifold_dimension(M))
+end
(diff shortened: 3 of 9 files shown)


mateuszbaran:
I could change the default implementations of `represent_in_basis` and `inverse_represent_in_basis` so that they try to use `basis`. What do you think? Or maybe have a `PrecalculatedBasis` struct that inherits from `AbstractBasis` that holds basis vectors and then implement `represent_in_basis` and `inverse_represent_in_basis` for this type?

mateuszbaran:
I've added `ProjectedONB`.

kellertuer:
Thanks for starting this. While I did not have time to check your interface idea, could you also include the diagonalizing one that includes the eigenvalues? And – should we write out `ONB`? Or is `ProjectedOrthoNormalBasis` too long?

I have to check in what sence the projected one is orthonormal (in the embedding before projecting? We could do Gram Schmidt in a tangent space?).

mateuszbaran:
> could you also include the diagonalizing one that includes the eigenvalues?

Sure, I can add a struct for that basis here. How would you call that struct? I think I need to tweak the return type of `basis` to let this function return kappas in a reasonable way.

> And – should we write out `ONB`? Or is `ProjectedOrthoNormalBasis` too long?

I don't have a preference here, I can expand all `ONB`s to `OrthonormalBasis` if you prefer it.

> I have to check in what sence the projected one is orthonormal (in the embedding before projecting? We could do Gram Schmidt in a tangent space?).

It assumes that the tangent space is a linear subspace of the ambient space of size specified by `representation_size`. It returns an ONB of the tangent space at a given point. We could use Gram-Schmidt instead of SVD but as far as I know SVD is more numerically stable.

kellertuer:
> > could you also include the diagonalizing one that includes the eigenvalues?
> 
> Sure, I can add a struct for that basis here. How would you call that struct? I think I need to tweak the return type of `basis` to let this function return kappas in a reasonable way.

Yes, that would be nice. Maybe since we return a `Basis` anyways, the one with the kappas might just have fields for that?


> I don't have a preference here, I can expand all `ONB`s to `OrthonormalBasis` if you prefer it.
I would prefer that for readability, yes.

> 
> > I have to check in what sence the projected one is orthonormal (in the embedding before projecting? We could do Gram Schmidt in a tangent space?).
> 
> It assumes that the tangent space is a linear subspace of the ambient space of size specified by `representation_size`. It returns an ONB of the tangent space at a given point. We could use Gram-Schmidt instead of SVD but as far as I know SVD is more numerically stable.

That would also imply that all tangent spaces get their metric by restricting the inner from the ambient space to the tangent? Otherwise you loose normality. By projection you will loose orthogonality anyways (in general) I think.

mateuszbaran:
> Yes, that would be nice. Maybe since we return a `Basis` anyways, the one with the kappas might just have fields for that?

Yes, I'll do something like that.

mateuszbaran:
> I would prefer that for readability, yes.

OK, I'll change it then.

> That would also imply that all tangent spaces get their metric by restricting the inner from the ambient space to the tangent? Otherwise you loose normality. By projection you will loose orthogonality anyways (in general) I think.

Projection both destroys orthogonality and changes the norm. I restore orthogonality using SVD. This assumes that orthogonalization in ambient space coincides with orthogonalization in the tangent space, which doesn't have to be true for all embedded manifolds but for many it is. Probably a `MetricManifold` with a non-diagonal metric wouldn't work here. I could add Gram-Schmidt for that case. Finally I normalize all vectors by dividing by their norm.

The SVD-based approach won't work for all manifolds but I still think it works for sufficiently many to put this method here.

kellertuer:
Great and thanks for the clarification. It sounds reasonable using SVD, the advantage of Gram schmidt (using the inner product on the tangent) would be, that it always returns orthonormals.

mateuszbaran:
BTW, could you write a docstring for that diagonalizing basis with kappas? I've read the one you have in Manopt.jl and the symbols aren't properly explained. What's the first vector that's mentioned there? What's `\Xi`? Why does it say that `\kappa_1` is zero but for sphere all kappas are equal to 1?

kellertuer:
Can you link that? The first one should be zero. As I linked the paper a few times, it's the curvature tensors diagonalization along the geodesic pointing towards the first tangent vectors direction. Since that's along a geodesic, that direction is not getting any (additional) curvature and is hence zero. I can't find the one you referring to, but Manopt.jl also does not have it that often, the matlab code does, see [here](https://github.com/kellertuer/MVIRT/blob/master/manifolds/Sn.m#L227).

\Xi is  the set of vectors (\xi_1,...\xi_n), i.e. the whole ONB, \kappa_1, ..., \kappa_n are above eigen values of the diagonalized curvature tensor along the geodesic indicated by the first, so \xi_1.

sethaxen:
Thanks for getting started on this. I haven't had a chance to look at the code yet, but I wonder why this PR is in our base package and not in Manifolds.jl. In our past discussions, we agreed new features would appear in Manifolds first and only after the interfaces settle would the interface perhaps end up in Manifolds.jl.

mateuszbaran:
Right, I forgot. I'll move it to `Manifolds.jl`.

kellertuer:
I just noticed, that if you provide a direction (x not y) then the function is actually correct, see a few lines after where the first component is set to zero (sadly, kind of, since I hoped I had found a bug why my bezier curves to not yet work).

mateuszbaran:
Right, I see, when the tangent vector is non-zero then it works.

kellertuer:
I would phrase that differently: if you have the zero vector, you do not get a direction for the geodesic and the resulting ONB is a classical one (hence all curvature Eigenvalues stay one). So then also it works :)

PR #20, opened 2020-01-18
Reworking allocation

mateuszbaran:
Representing points on a power manifold with nested arrays of arrays (discussed here: https://github.com/JuliaNLSolvers/Manifolds.jl/pull/92#issuecomment-575062069 for example) forces us to change a bit the way points are allocated. I've introduced here a new pattern of `scalar_eltype` (that unwraps all arrays and tuples until it reaches a scalar `eltype`) and `allocate` which is like `similar` but for nested arrays it actually calls similar at the lowest level only and at higher levels of nesting just `map`s through the structure.

Unfortunately, it's a breaking change that can't be done in `Manifolds.jl` first, so I've opened a PR here. I'll open a matching PR in `Manifolds.jl` later when we agree on the direction it should take.

Along with this changes, I'd like to tackle https://github.com/JuliaNLSolvers/Manifolds.jl/issues/95 in the PR I'll open in `Manifolds.jl`.

@sethaxen What do you think about these changes?

### DIFF
--- a/src/ManifoldsBase.jl
+++ b/src/ManifoldsBase.jl
@@ -167,13 +167,13 @@ Retraction method can be specified by the last argument, defaulting to
 methods.
 """
 function retract(M::Manifold, x, v)
-    xr = similar_result(M, retract, x, v)
+    xr = allocate_result(M, retract, x, v)
     retract!(M, xr, x, v)
     return xr
 end
 retract(M::Manifold, x, v, t::Real) = retract(M, x, t * v)
 function retract(M::Manifold, x, v, method::AbstractRetractionMethod)
-    xr = similar_result(M, retract, x, v)
+    xr = allocate_result(M, retract, x, v)
     retract!(M, xr, x, v, method)
     return xr
 end
@@ -224,12 +224,12 @@ Inverse retraction method can be specified by the last argument, defaulting to
 for available methods.
 """
 function inverse_retract(M::Manifold, x, y)
-    vr = similar_result(M, inverse_retract, x, y)
+    vr = allocate_result(M, inverse_retract, x, y)
     inverse_retract!(M, vr, x, y)
     return vr
 end
 function inverse_retract(M::Manifold, x, y, method::AbstractInverseRetractionMethod)
-    vr = similar_result(M, inverse_retract, x, y)
+    vr = allocate_result(M, inverse_retract, x, y)
     inverse_retract!(M, vr, x, y, method)
     return vr
 end
@@ -253,7 +253,7 @@ The function works only for selected embedded manifolds and is *not* required to
 closest point.
 """
 function project_point(M::Manifold, x)
-    y = similar_result(M, project_point, x)
+    y = allocate_result(M, project_point, x)
     project_point!(M, y, x)
     return y
 end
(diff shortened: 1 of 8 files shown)


mateuszbaran:
> I like the approach – would that also allow arbitrary manifolds to use a nested data representation without any extra implementation? I am not saying that that's reasonable, but if its included thats nice. For the Fixed Rank for example I overwrote the functions since I have an internal data structure for those.

Actually only `AbstractArray`-based and tuple-based nesting will be supported by default. Each `struct` used to represent points would have to be accommodated more or less like before this change.

> Testing and Docs should be done, but of course we should first agree on the approach and I like this approach 👍

Sure, I'll add more tests and docs when we agree on the right approach.

kellertuer:
Oh sorry, I never meant to imply, that this approach would do structs (that was just the comparisor to what I did once).

sethaxen:
In general I like this approach. My only issue is with the naming. `scalar_eltype` might be better named to `number_eltype` for consistency with `Manifolds.jl`'s `AbstractNumbers`.

When the coverage is complete and the methods are documented, I'm happy to see this merged.

mateuszbaran:
Right, `number_eltype` sounds better 🙂 .

mateuszbaran:
I think, I should probably also rename `similar_result` to `allocate_result` and `similar_result_type` to `allocate_result_type`?

src/ManifoldsBase.jl
```diff
@@ -739,7 +739,7 @@ export allocate,
        representation_size,
        retract,
        retract!,
-       scalar_eltype,
+       number_eltype,
```
sethaxen: Needs a reorder now

sethaxen:
> I think, I should probably also rename `similar_result` to `allocate_result` and `similar_result_type` to `allocate_result_type`?

Yes, that's a good idea.

mateuszbaran:
I think this PR is more or less complete now, including tests and documentation. I'd like to merge it if CI passes and tag 0.4.0 release then. I'll then update https://github.com/JuliaNLSolvers/Manifolds.jl/pull/98 to use the tagged version 0.4.0 of `ManifoldsBase.jl`.

kellertuer:
Thanks for the docstrings. Yes, let's wait for the CI and if that's fine, we should up this to `0.4.0`.

mateuszbaran:
I've added the docs because it's a relatively tricky part of the code and I think it's better if someone else also has some idea how it works :slightly_smiling_face: .

kellertuer:
For Manifolds we could preview branch-docs, I have to check how I get those here, easiest might be to check the branch out locally – will do that after my upcoming exercise.

mateuszbaran:
I don't really understand why the coverage decreased. Is it OK to merge this?

kellertuer:
In line 547 of `ManifoldsBase.jl` we are also using the old style of first inplace-modification and then returning in a separate line; that was unified on `manifolds.jl` to one line. And maybe we don't have function that gets the returned value to do something with it (i.e. no `v = ...` line with a test afterwards?)

I think for the same reason no test runs through lines 262 and 263 in `ArrayManifold`. Can you maybe just add a test similar to the existing ones but using a return-value-style `v = ...` and a `@test v`, such that both methods have to actually return something in one case (and that does not get optimised away then)?

mateuszbaran:
That's weird. I don't see how these lines could have been covered on master branch. It's probably some improvement to coverage algorithms.

kellertuer:
I also don't know, might also be, that Julia is more efficient and has separate functions for returning and non-returning calls? However, now I think this PR can be merged and the new version be triggered.

PR #23, opened 2020-02-12
[WIP] Decorator improvements

mateuszbaran:
I'm opening this PR to implement some changes prompted by https://github.com/JuliaNLSolvers/Manifolds.jl/pull/114 .

The first part is making `depth` argument of `base_manifold` a `Val` to make `base_manifold` type stable.


kellertuer:
We can also put the macro already here and decorate most basic functions. I can develop the PR on Manifolds also with this PR checked out for Base then.

mateuszbaran:
OK, I'll commit my changes in a moment and you can continue then. I'll put the macro here.

kellertuer:
Since you just mentioned that in the commit: Can we do default values? I haven't checked what happens currently for the default values from `exp` for example.

mateuszbaran:
I can add support for default values, yes. I didn't do that yet because it's a bit complicated and no functions from `Decorator.jl` required it.

kellertuer:
It would be nice, as said, for `exp`. Further

* `is_transparent_function`
* `default_decorator`
* `_acts_transparently`

need to be ported

and we could call `val_is_transparent_function` just `val_transparency`?

And, since your porting, the error message function could be put here an used.

mateuszbaran:
OK, I'll port it tomorrow and add support for default argument values. I have to prepare a few lectures now.

kellertuer:
After test coverage (and checking documentation again) I think this is ready to merge. In other words: It's feature complete.

src/DecoratorManifold.jl
```diff
@@ -0,0 +1,475 @@
...
+values.
+
+This yields the following advantange: For a manifold one usually implicitly assumes for
+example a metric. To avoid reimplementation of this metric when introducing a second metric,
+the first metric can be set to be the default, i.e. its implementaion is already given by
+the undecorated case.
+
+to change this value, see [`default_decorator_dispatch`](@ref).
+"""
+is_default_decorator(M::Manifold) = _extract_val(default_decorator_dispatch(M))
```
mateuszbaran: `_extract_val` needs to be moved to ManifoldsBase.jl.

src/DecoratorManifold.jl
```diff
@@ -0,0 +1,475 @@
...
+This yields that _all_ functions are passed through to the decorated [`Manifold`](@ref)
+if `M` is indicated as default. This overwrites all [`is_decorator_transparent`](@ref)
+values.
+
+This yields the following advantange: For a manifold one usually implicitly assumes for
+example a metric. To avoid reimplementation of this metric when introducing a second metric,
+the first metric can be set to be the default, i.e. its implementaion is already given by
+the undecorated case.
+
+to change this value, see [`default_decorator_dispatch`](@ref).
```
mateuszbaran:
```suggestion
Value returned by this function is determined by [`default_decorator_dispatch`](@ref),
which returns a `Val`-wrapped boolean for type stability of certain functions.
```

src/DecoratorManifold.jl
```diff
@@ -0,0 +1,475 @@
...
+just passes through down to the internally stored manifold.
+Only decorator manifolds can be transparent and their default is, to be transparent.
+A function that is affected by the decorator indicates this by returning `false`. To change
+this behaviour, see [`decorator_transparent_dispatch`](@ref).
+
+If a decorator manifold is not in general transparent, it might still pass down
+for the case that a decorator is the default decorator, see [`is_default_decorator`](@ref).
+"""
+function is_decorator_transparent(f, M::Manifold, args...)
+    return decorator_transparent_dispatch(f, M, args...) == Val(:transparent)
```
mateuszbaran:
```suggestion
    return decorator_transparent_dispatch(f, M, args...) === Val(:transparent)
```
This should also work and could be faster.

src/DecoratorManifold.jl
```diff
@@ -0,0 +1,475 @@
...
+A function to decide whether by default to dispatch th the inner manifold of
+a decorator (`Val(true)`) or not (`Val(false`). For more details see
+[`is_decorator_transparent`](@ref).
+"""
+default_decorator_dispatch(M::Manifold) = Val(false)
+
+"""
+    is_decorator_transparent(f, M, args...)
+
+Given a [`Manifold`](@ref) `M` and a function `f(M,arge...)`, indicate, whether a
```
mateuszbaran:
```suggestion
Given a [`Manifold`](@ref) `M` and a function `f(M, args...)`, indicate, whether an
```

src/DecoratorManifold.jl
```diff
@@ -0,0 +1,475 @@
...
+[`AbstractDecoratorManifold`](@ref) acts transparent for `f`. This means, it
```
mateuszbaran:
```suggestion
[`AbstractDecoratorManifold`](@ref) acts transparently for `f`. This means, it
```

src/DecoratorManifold.jl
```diff
@@ -0,0 +1,475 @@
...
+just passes through down to the internally stored manifold.
+Only decorator manifolds can be transparent and their default is, to be transparent.
```
mateuszbaran:
```suggestion
Transparency is only defined for decorator manifolds and by default all decorators are transparent.
```
Does that sound better to you? I think the meaning of "Only decorator manifolds can be transparent" is quite unclear.

src/DecoratorManifold.jl
```diff
@@ -0,0 +1,475 @@
...
+    return decorator_transparent_dispatch(f, M, args...) == Val(:transparent)
+end
+
+"""
+    decorator_transparent_dispatch(f, M, arge...)
+
+Given a [`Manifold`](@ref) `M` and a function `f(M,arge...)`, indicate, whether a
+function is `Val(:transparent)` or `Val(:intransparent)` for the (decorated)
+[`Manifold`](@ref) `M`. Another possibility is, that for `M` and given `args...`
+the function `f` should invoke `M`s `Val(:parent)` implementation.
```
mateuszbaran: I'd add a reference to `@decorator_transparent_function` for more details.

src/DecoratorManifold.jl
```diff
@@ -0,0 +1,475 @@
...
+
+@decorator_transparent_signature log!(M::AbstractDecoratorManifold, X, p, q)
+
+@decorator_transparent_signature manifold_dimension(M::AbstractDecoratorManifold)
+
+@decorator_transparent_signature project_point!(M::AbstractDecoratorManifold, q, p)
+
+@decorator_transparent_signature project_tangent!(M::AbstractDecoratorManifold, Y, p, X)
+
+@decorator_transparent_signature projected_distribution(M::AbstractDecoratorManifold, d, p)
```
mateuszbaran: I think we should leave distribution-related stuff in Manifolds.jl for now.

src/DecoratorManifold.jl
```diff
@@ -0,0 +1,475 @@
+#
+# Type
+#
+"""
+    AbstractDecoratorManifold <: Manifold
+
+An `AbstractDecoratorManifold` indicates that to some extend a manifold subtype
+decorates another manifold in the sense that
```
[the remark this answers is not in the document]
kellertuer: no that's fine, I'll just remove the “it” from the items then because now its already upfront.

src/DecoratorManifold.jl
kellertuer: Oh, yes; I even added them over there to distributions, but forgot to delete them here.

mateuszbaran:
I've slightly improved the test coverage and made some minor edits. I think the design is nice but we still need better coverage.

mateuszbaran:
I think it's worth adding `ManifoldsBase.` in macros for calls to internal `ManifoldsBase` functions to not pollute the global namespace in other packages (I did it in a recent commit).

kellertuer:
I increased the cases for `injectivity_radius`.

mateuszbaran:
I've added a few more tests but I'm having a bit of trouble understanding the correct usage of `@decorator_transparent_fallback`. It seems like it's used in Manifolds.jl for a few random functions and I don't see a clear pattern there. Some of its options aren't even used as far as I can see.

kellertuer:
The idea of the fallback is to introduce an implementation on an abstract level and for a specific case. For example if

```julia
function exp!(MyAbstractInBetweenDecorator, q, p, X)
```
should be interoduced as a new default for `: intransparent` you can just
```julia
@decoratr_transparent_fallback :intransparent function exp!(MyAbstractInBetweenDecorator, q, p, X)
```
I think this is necessary for abstract classes, when you want to overwrite either `:intransparent` or `:parent` cases. I had hoped the docs where specific enough, sorry.

kellertuer:
Oh and I just noticed, the fallback does not work for inline functions (yet) as does the function definition macro. I think one has to tweak the body for that, so I for now removed it from function – we should remove it from fallback, too.

mateuszbaran:
I think I'm finished for now tweaking this PR. I'll think if the docs for these three macros could be improved though.

> Oh and I just noticed, the fallback does not work for inline functions (yet) as does the function definition macro. I think one has to tweak the body for that, so I for now removed it from function – we should remove it from fallback, too.

It's possible that it'd be enough to `@macroexpand` the expression with function to support `@inline`. The `@inline` macro just injects an expression into function body.

kellertuer:
Well but currently the implementations are still different, the fallback still “reacts” on inline functions while the other macro does not.

What would you improve on them?

mateuszbaran:
> Well but currently the implementations are still different, the fallback still “reacts” on inline functions while the other macro does not.

It depends on what you consider by "reacts". `macroexpand` would make sure that `@inline` is accepted and forwarded to the method with `$body`. Wouldn't that be fine? That should work for both `_function` and `_fallback` macros.

> What would you improve on them?

Probably a bit more discussion of examples, I'm not sure yet.

kellertuer:
> > Well but currently the implementations are still different, the fallback still “reacts” on inline functions while the other macro does not.
> 
> It depends on what you consider by "reacts". `macroexpand` would make sure that `@inline` is accepted and forwarded to the method with `$body`. Wouldn't that be fine? That should work for both `_function` and `_fallback` macros.

That would be perfectly fine. We should adapt `_fallback` to work the same as `_function` then (i.e.  not accepting inlines directly).

mateuszbaran:
Now we can accept `@inline` pretty directly 🙂 .

src/DecoratorManifold.jl
```diff
@@ -69,7 +69,8 @@ end
 macro decorator_transparent_fallback(ex)
     return esc(quote @decorator_transparent_fallback :intransparent ($ex) end)
 end
-macro decorator_transparent_fallback(fallback_case, ex)
+macro decorator_transparent_fallback(fallback_case, input_ex)
+    ex = macroexpand(ManifoldsBase, input_ex)
     if ex.head == :function || ex.head == :(=) #complete or inline function
```
kellertuer: With the expansion, the `||` case can be omitted.
```suggestion
    if ex.head == :function
```

kellertuer: Oh and the following error message has to be adapted accordingly, too (same for the `_function` case.

mateuszbaran: I think there is was a misunderstanding. `@inline` functions and functions defined using `=` instead of the `function` keyword are two different things. `macroexpand` doesn't touch `=`.

kellertuer: Oh, then I had the wrong term in mind. Still, for functions defined with `=` `_fallback` does not work (neither does `_function`, but there I removed that and extended the error message).

mateuszbaran: I think not supporting `=` functions is OK there.

kellertuer: I think so, too. Just also adapt the error message :)

mateuszbaran:
While increasing coverage I've spotted a bug with forwarding of keyword arguments. It's fixed now.

kellertuer:
Hm, are you sure you fixed it? I get several `unsupported keywords` errors now over in `Manifolds.jl` with this version of commit `43936c4`. Did you maybe loose some functions keywords somewhere now?

mateuszbaran:
Sorry, I've fixed one case but broken another. I'll fix this in a moment.

kellertuer:
Nevermind, I wouldn't have noticed the first case. It's working again.

kellertuer:
While most remaining test cases should be doable (define a function, use `signature` and set it to `:parent` is missing for example I think), the two error messages after the ` if` checking for head or call ... I am not sure how to trigger those?

mateuszbaran:
I don't really understand some of these coverage failures. Some of these uncovered lines are definitely executed. I think it would be OK to just cover all methods generated by these macros (for example the methods with `invoke` aren't covered yet).

kellertuer:
Yes, the invoke with parent should be covered and are not yet; for one remaining I have an idea, for the two errors (neither where nor call) I don'T know how to get them, if a call would issue them you can see in my tests how to check that (for the other messages I did that).

kellertuer:
I made a change to the decorators (and unified exp and exp!): Not also non-mutating functions can be transparent and are so by default. This fixes non-mutating manifolds to work the same way. I just have to check for metricmanifold whether I thought of all changes.

mateuszbaran:
I've extracted some parts of these macros (including these errors) into separate functions. This should make testing easier.

> I made a change to the decorators (and unified exp and exp!): Not also non-mutating functions can be transparent and are so by default. This fixes non-mutating manifolds to work the same way. I just have to check for metricmanifold whether I thought of all changes.

I think it's a good decision, decorators should be able to handle special non-mutating variants.

kellertuer:
And I removed `exp(M,p,X,T)` with a vector of `T`s, since we have `geodesic` for that; interestingly we also just did that for the non-mutating one; now that's unified, too.

Great simplification that you did, the code was a little repetitive, yes.

src/DecoratorManifold.jl
```diff
@@ -0,0 +1,616 @@
...
+        end
+    end)
+end
+
+#
+# Functions
+#
+
+"""
+    is_default_decorator(M)
```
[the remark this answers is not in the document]
mateuszbaran: `is_default_decorator` returns a `Bool`. `default_decorator_dispatch` returns a `Val`.

kellertuer:
> This is a really nice unification. I haven't tried out the code, but it looks like it will simplify a lot of our implementations.
Yes, that's what I noticed when I started the `EmbeddedManifold` (that's where it escalated). Thanks for taking the time to work through this approach.

> 
> RE the macros, it really would be ideal to have shorter, yet more descriptive names for these.
> From the names alone, I have no idea what they do even though they're quite long.
> Moreover, it's confusing that they all have `transparent` in the name, yet can also take
> `:intransparent` and `:parent` as arguments.
Yes the naming scheme grew into that and we should adapt the names.
I think it can even just be one macro if done carefully:
* symbol + function is the `_fallback`
* function is the `_function`
* signature is the `_signature`

(though I am sure it would take me a long time to unify that and distinguish these expressions)

and call that macro maybe `@decorate_manifold`?

> I'd favor having utility macros that eliminate the need for the symbol argument, even if internally they just call another macro.
I actually like the symbol (compared to true/false) because it enables to later add another path to dispatch on. 

> Also, I think this code needs to be processed by JuliaFormatter.
Yes, after naming the macros.

src/DecoratorManifold.jl
```diff
@@ -0,0 +1,616 @@
...
+This can especially be used if for an already transparent function and an abstract
+intermediate type a change in the default is required.
+For implementing a concrete type, neither this nor any other trick is necessary. One
+just implements the function as before. Note that a decorator that [`is_default_decorator`](@ref)
+still dispatches to the transparent case.
+
+
+* `:transparent` states, that the function is transparently passed on to the manifold that
+is decorated by the [`AbstractDecoratorManifold`](@ref) `M`, which by default is assumed to
+be stored in `M.manifold`.
```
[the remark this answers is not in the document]
kellertuer: That would be good for my `EmbeddedManifold` where for technical reasons currently the `embedding` is stored in `.manifold` 🙄

src/DecoratorManifold.jl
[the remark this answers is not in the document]
sethaxen: Ah, I see, then perhaps replace the `Val` with a `Bool`, just to be clear.

sethaxen:
One more question, do these macros also let us append docstrings to the functions or does that need to be done separately?

kellertuer:
> One more question, do these macros also let us append docstrings to the functions or does that need to be done separately?

Currently this has to be done separately before (but we kind of do that for the mutating ones, documenting the nonmutating, anyways in most places). If that could be handled, that would be quite neat, you're right, but a minor issue I think.

src/DecoratorManifold.jl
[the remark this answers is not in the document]
mateuszbaran: Should it be different from `base_manifold(M, Val(1))`?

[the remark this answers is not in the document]
kellertuer: It has the same functionality but maybe should be a separate function (default implementation exactly your line) to be easier overwritten?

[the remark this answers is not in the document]
mateuszbaran: OK, this makes sense.

src/DecoratorManifold.jl
```diff
@@ -0,0 +1,616 @@
...
+The cases of transparency are
+
+* `:transparent` states, that the function is transparently passed on to the manifold that
+is decorated by the [`AbstractDecoratorManifold`](@ref) `M`, which by default is assumed to
+be stored in `M.manifold`.
+* `:intransparent` states that an implementation for this decorator is required, and if
+none of the types provides one, an error is issued. Since this macro provides such an
+implementation, this is the default.
+* `:parent` states, that this function passes on to the supertype instead of to the
+decorated manifold.
```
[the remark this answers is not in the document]
mateuszbaran: I've added a note here.

src/DecoratorManifold.jl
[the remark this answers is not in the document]
mateuszbaran: I've introduced the `decorated_manifold` function here. I'll adapt Manifolds.jl PR in a moment.

kellertuer:
Shall we try to unify the macros or leave that for a later PR? Should we rename the macros to something shorter?
Despite that I think this is finished.

mateuszbaran:
I think having these three macros is fine. I'm not sure about names but we can keep them as well.

We could remove exporting of the three macro names, `is_decorator_transparent` and `is_default_decorator`. These are sort of internal, an average user would never use any of them. Other than this, I think we can merge this PR.

sethaxen:
I haven't had a chance to review the recent changes, but I trust you both and don't want to hold up progress. Feel free to merge when ready. 🙂

mateuszbaran:
I've removed these exports. If there are no objections, I'll merge this tomorrow and tag a new release.

mateuszbaran:
OK, since you approve I'll merge it now :slightly_smiling_face: .

mateuszbaran:
That is, once PR passes.

PR #26, opened 2020-03-11
Moving Bases to Base

mateuszbaran:
In this PR I'll move basis interface as it is in https://github.com/JuliaNLSolvers/Manifolds.jl/pull/115 . For a start, a small change that fixes special euclidean tests for me.

### DIFF
--- a/src/ArrayManifold.jl
+++ b/src/ArrayManifold.jl
@@ -139,18 +139,71 @@ end
 function get_basis(M::ArrayManifold, p, B::AbstractBasis; kwargs...)
     is_manifold_point(M, p, true; kwargs...)
     Ξ = get_basis(M.manifold, array_value(p), B)
-    nV = length(get_vectors(M.manifold, array_value(p), Ξ))
-    if nV != manifold_dimension(M.manifold)
-        return ErrorException(
-            "For a basis of the tangent space at $(p) of $(M.manifold), $(manifold_dimension(M)) vectors are required, but get_basis $(B) computed $(nV)"
-        )
+    bvectors = get_vectors(M, p, Ξ)
+    N = length(bvectors)
+    if N != manifold_dimension(M.manifold)
+        throw(ErrorException(
+            "For a basis of the tangent space at $(p) of $(M.manifold), $(manifold_dimension(M)) vectors are required, but get_basis $(B) computed $(N)"
+        ))
     end
-    map(X -> is_tangent_vector(M, p, X, true; kwargs...), get_vectors(M.manifold, array_value(p), Ξ))
+    # check that the vectors are linearly independent\
+    bv_rank = rank(reduce(hcat, bvectors))
+    if N != bv_rank
+        throw(ErrorException(
+            "For a basis of the tangent space at $(p) of $(M.manifold), $(manifold_dimension(M)) linearly independent vectors are required, but get_basis $(B) computed $(bv_rank)"
+        ))
+    end
+    map(X -> is_tangent_vector(M, p, X, true; kwargs...), bvectors)
+    return Ξ
+end
+function get_basis(
+    M::ArrayManifold,
+    p,
+    B::Union{AbstractOrthogonalBasis,CachedBasis{<:AbstractOrthogonalBasis}};
+    kwargs...,
+)
+    is_manifold_point(M, p, true; kwargs...)
+    Ξ = invoke(get_basis, Tuple{ArrayManifold,Any,AbstractBasis}, M, p, B; kwargs...)
+    bvectors = get_vectors(M, p, Ξ)
+    N = length(bvectors)
+    for i = 1:N
+        for j = i+1:N
+            dot_val = real(inner(M, p, bvectors[i], bvectors[j]))
+            if !isapprox(dot_val, 0; atol = eps(eltype(p)))
+                throw(ArgumentError("vectors number $i and $j are not orthonormal (inner product = $dot_val)"))
+            end
+        end
+    end
+    return Ξ
+end
+function get_basis(
+    M::ArrayManifold,
+    p,
+    B::Union{AbstractOrthonormalBasis,CachedBasis{<:AbstractOrthonormalBasis}};
+    kwargs...,
+)
+    is_manifold_point(M, p, true; kwargs...)
+    Ξ = invoke(get_basis, Tuple{ArrayManifold,Any,AbstractOrthogonalBasis}, M, p, B; kwargs...)
+    bvectors = get_vectors(M, p, Ξ)
+    N = length(bvectors)
+    for i = 1:N
+        Xi_norm = norm(M, p, bvectors[i])
+        if !isapprox(Xi_norm, 1)
+            throw(ArgumentError("vector number $i is not normalized (norm = $Xi_norm)"))
+        end
+    end
     return Ξ
 end
 for BT in DISAMBIGUATION_BASIS_TYPES
+    if BT <: Union{AbstractOrthonormalBasis,CachedBasis{<:AbstractOrthonormalBasis}}
+        CT = AbstractOrthonormalBasis
+    elseif BT <: Union{AbstractOrthogonalBasis,CachedBasis{<:AbstractOrthogonalBasis}}
+        CT = AbstractOrthogonalBasis
+    else
+        CT = AbstractBasis
+    end
     eval(quote
-        @invoke_maker 3 AbstractBasis get_basis(M::ArrayManifold, p, B::$BT; kwargs...)
+        @invoke_maker 3 $CT get_basis(M::ArrayManifold, p, B::$BT; kwargs...)
     end)
 end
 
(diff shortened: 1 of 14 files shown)


kellertuer:
What's your feeling using the new dispatch mode using specific functions?

mateuszbaran:
The transition was quite smooth, the only problem is slightly non-intuitive behavior when defining new fallbacks (if the __parent/__transparent/__intransparent function was defined in ManifoldsBase, it needs to be specifically imported). Maybe it could be fixed in the fallback macro, I didn't check that yet.

> Looking at the coverage I think all this collecting dispatch and calling `_get_vector` is not necessary anymore? We should also check for base code coverage.

I didn't check that either, there is still quite a bit of stylistic improvements to be done here. Ambiguities need to be checked and possibly resolved and the coverage needs to be improved. At least I do not anticipate any major changes in this PR and the Manifolds.jl one.

src/DecoratorManifold.jl
```diff
@@ -221,29 +224,30 @@ macro decorator_transparent_function(fallback_case, input_ex)
...
             ) where {$(where_exprs...)}
-                return ($fname)(
-                    $(argnames[1]),
-                    ManifoldsBase._acts_transparently($fname, $(argnames...)),
-                    $(argnames[2:end]...),
-                    ;
-                    $(kwargs_call...),
-                )
+                transparency = ManifoldsBase._acts_transparently($fname, $(argnames...))
+                if transparency === Val(:parent)
```
kellertuer: Can this maybe be done with a dispatch? Maybe the function is not the nicest (it also need `parts`, `argnames` and `kwargs_call`), but it would be nicer and a little more Julia-style-ish?

mateuszbaran: Hmm... maybe? I'll try turning that `if` into dispatch.

mateuszbaran: As far as I can see it would require a lot of methods like
```julia
get_dispatch_function(::typeof(exp!), ::Val{:transparent}) = exp!__transparent
```
One method for each function and transparency type. Do you think it's worth it?

kellertuer: I had hoped it would suffice to do one function per mode, like
````
get_dispatch_function(f, ::Val{:transparent}, parts, kwarg-names) = ...
````
where this is before you quote it again so you can build up the function name and pass it back? But I also did not try, that's why I carefully asked.

Comment (mateuszbaran) [reply src/DecoratorManifold.jl]: So, you mean dispatch at macroexpansion? That wouldn't give us anything since the macro wouldn't be re-run each time `get_dispatch_function` gets new methods.

Comment (kellertuer) [reply src/DecoratorManifold.jl]: You're right. Then we can also stick with the ifs, though currently that's doubled, right?

Comment (mateuszbaran) [reply src/DecoratorManifold.jl]: Yes, it's duplicated but there is enough differences around the duplicated block that removing this duplication would almost certainly add complexity.

Comment (kellertuer) [reply src/DecoratorManifold.jl]: Ok, then it's fine. I only saw the block added in a quite similar manner and did not yet check their surrounding code.

**Comment by mateuszbaran** [comment]
@kellertuer  could you take a look at my last commit? I was trying to improve coverage and I got a bit stuck on weird errors related to the ArrayManifold. Some basis tests weren't working because DefaultManifold didn't have check functions, so I've added them but then other things got broken and I don't really understand why this fails now.

File: src/ArrayManifold.jl [hunk 2]
```diff
@@ -130,6 +130,80 @@ function exp!(M::ArrayManifold, y, x, v; kwargs...)
...
+        end
+    end
+    return B
+end
+
+function get_coordinates(M::ArrayManifold, p, X, B::AbstractBasis; kwargs...)
+    is_tangent_vector(M, p, X, true; kwargs...)
+    return get_coordinates(M.manifold, p, X, B)
+end
+for BT in DISAMBIGUATION_BASIS_TYPES
```
kellertuer: I thought the new scheme might make it possible not to have these? Though I first have to check carefully ;)

mateuszbaran: No, this is still necessary. But the new scheme removed some other method ambiguities, so it has some benefits.

kellertuer:
> Could you take a look at my last commit?

Interesting error. The first one I could fix (will submit in a second), we decorated a little too early, or in other words the `check_tangent_vector` is transparent – calls the default manifolds one instead of the arraymanifold, that one was called before and can handle `TVector`s. While this one could easily be fixed by decorating only afterwards, the line after that fails, too, because `check_manifold_point` got transparent by default or in other words the one in `ArrayManifold` does not get called anymore, the manifold is undecorcates (array->default) and the defaults check is called. That one, howver, is of course not able to handle `MPoint`s. 

Did you accidentially make transparency ignore actual implementations? It seems `ArrayManifold` is here treated as `AbstractDecoratorManifold` and gets undecorated, while in fact a more specific implementation is provided in `ArrayManifold` actually.

kellertuer:
One could of course change line 119 in `ArrayManifold` from

```julia
    is_tangent_vector(M, x, v, true; kwargs...)
```
to
```julia
    is_tangent_vector(M, array_value(x), array_value(v), true; kwargs...)
```
or even
```julia
    is_tangent_vector(M.manifold, array_value(x), array_value(v), true; kwargs...)
```
but I had hoped either `is_tangent_vector` (which should not be transparent but parent) or `check_tangent_vector` (whcih does not seem to be called for array manifold as mentioned) would take care of the undecoration (they did before the switch to functions instead of vals). So – what did we miss?

mateuszbaran:
I'm not sure what went wrong, all tests were passing on the "reducing ambiguities" commit and then the changes in the next one revealed this problem. I was trying to keep the exact same behavior of the decorators. Are you sure this undecoration worked before? It was revealed by one of the new tests I've added, the old ones are still passing.

kellertuer:
Ah I see, you implemented the `DefaultManifold`s checks. They now get called instead of the array ones. I am not sure why (so I am also not sure whether it was correct before ;)), but it seems they are transparent (passing from array->default) and the original arraymanifold function is no longer called (that would take care of stripping away MPoint and TVector before passing to default). So maybe the checks are transparent but should not be?

kellertuer:
I found the culprit: The checks where implemented for `MPoint`/`TVector` only, though using `array_value` internally. I am not taking a look at the errors with bases.

mateuszbaran:
Thanks! It's quite strange that this only surfaced now.

kellertuer:
I had in mind I never added those types, maybe you added them to reduce ambiguities? I just added a second one with `MPoint` to reduce that again. I have implemented `get_basis` now just coordinates/vector are left to fix.

mateuszbaran:
> I had in mind I never added those types, maybe you added them to reduce ambiguities?

Which types do you mean? `MPoint` and `TVector`? IIRC we don't actually rely on them right now.

kellertuer:
Yeah, we actually don't but removing those from the checks resolved the errors. I fixed the remaining bases errors so we now just have to check for ambiguitites, which raised to 20 during my fixing and was already down to 12.

mateuszbaran:
I think I've finally *actually* fixed ArrayManifold-bases interactions, with proper checks of linear independence etc. Part of the complexity of my solution is caused by the fact that cached orthogonal and orthonormal bases are not actually subtypes of, respectively, `AbstractOrthogonalBasis` and `AbstractOrthonormalBasis`.

kellertuer:
Thanks for those checks, I think they are very helpful when looking for errors involving orthogonality or orthonormality! And I thought about extending the const, but wasn't sure, so good that you did that (and yay! only 10 disambiguities left!).

mateuszbaran:
As far as I can see some ambiguities are caused by this: https://github.com/JuliaNLSolvers/ManifoldsBase.jl/blob/88cf3536c7f51478062faf5a66dfeb300191affe/src/ManifoldsBase.jl#L246 . This method can't really work properly since it would only get triggered when no checks were implemented for a manifold, but when some are but not for `MPoint`, then we'll only get ambiguity errors when passing `MPoint`s.

kellertuer:
Indeed, that does not look reasonable.

mateuszbaran:
So maybe we should just remove that method? But then `MPoint` wouldn't ensure checking as initially planned (only `ArrayManifold` would do that).

kellertuer:
Yes, that was the reason for that function, hm, and yes, this way you could start “anywhere” to switch to testing. Maybe in the end it's only reasonable to start changing the manifold then. We can still keep MPoint and TVector because between computations you can still get information about the data that you would not get from a “pure” array.

kellertuer:
I think we never test for the deault-transparency yet and somehow the parent calls might get optimized way – despite that decorator is now thoroughly tested – for basis there are still some cases left to do, especially with wrong length of some values (in cached basis I think?)

mateuszbaran:
I think the main remaining points to cover are:

- [ ] complex coefficient bases
- [ ] non-broadcasting `get_vector!` on a cached basis
- [ ] `get_vectors` on a diagonalizing basis

I wouldn't worry about other uncovered lines. I'll add tests for (2) and (3) today.

kellertuer:
I noticed that my actual approaches to get to `:parent` dispatches were wrong, and we don't have a default-case yet, which should be the three main lines missing in decorator.

mateuszbaran:
> I noticed that my actual approaches to get to `:parent` dispatches were wrong, and we don't have a default-case yet, which should be the three main lines missing in decorator.

OK, I see. Could you also add a test for complex coefficient bases? After that, this PR will be complete.

kellertuer:
I am busy the afternoon, but tomorrow morning could work, unless you have time and want to do that.

PR #27, opened 2020-03-17
Unify project and embed

kellertuer:
This PR unifies both projection and embedding of points and tangents, following the same general approach as we do for metrics: Implicitly assumed ambient spaces may be used for a manifold.

* shorten `project_tangent` and `project_point` to `project`
* introduce manifold fallbacks for `embed` similar to `project`.


mateuszbaran:
Since you also add `embed` in this PR, maybe we could also move the entire `EmbeddedManifold` decorator here?

kellertuer:
Oh I did not think of that. Do you think that's reasonable? We could also move the other decorators here, MetricManifold and Group. If we move decorators we should move one. Then I would do it in a separate PR.

Here I mainly wanted to unify where these functions are defined and where the fallbacks are.

mateuszbaran:
I think moving `EmbeddedManifold` is very natural with both `project` and `embed` here. It's also relatively simple. `MetricManifold` is much more complicated since it provides generic fallbacks and relies on differentiation support (which I'm reworking in https://github.com/JuliaNLSolvers/Manifolds.jl/pull/110 ). `GroupManifold` would be easier to move here but it's quite complicated anyway and `Identity` still requires some work. I think it'd be fine to just move `EmbeddedManifold` for now.

kellertuer:
We are down (again) to just 3 lines missing, which look like they are optimised away, since the `:parent`-invoke missing is called if one looks at the ifs upfront – and the two lines/variables in show of a basis are used in the following line. When this one is merged (and after the basis rework) we should look how to adapt/rearrange the docs.

src/ManifoldsBase.jl
```diff
@@ -264,6 +264,54 @@ Shortest distance between the points `p` and `q` on the [`Manifold`](@ref) `M`.
...
 distance(M::Manifold, p, q) = norm(M, p, log(M, p, q))
 
+"""
+    embed(M::Manifold, p)
+
+Embed point `p` from the [`Manifold`](@ref) `M` into the ambient space.
+The function works only for selected embedded manifolds.
+"""
+function embed(M::Manifold, p)
+    q = allocate_result(M, project, p)
```
mateuszbaran:
```suggestion
    q = allocate_result(M, embed, p)
```

src/ManifoldsBase.jl
```diff
@@ -264,6 +264,54 @@ Shortest distance between the points `p` and `q` on the [`Manifold`](@ref) `M`.
...
+end
+
+"""
+    project!(M::Manifold, q, p)
+
+Embed point `p` from the [`Manifold`](@ref) `M` into the ambient space. The point `q` is
+overwritten by the embedded point. The function works only for selected embedded manifolds.
+"""
+function embed!(M::Manifold, q, p)
+    error(manifold_function_not_implemented_message(M, project!, q, p))
```
mateuszbaran:
```suggestion
    error(manifold_function_not_implemented_message(M, embed!, q, p))
```

src/ManifoldsBase.jl
```diff
@@ -264,6 +264,54 @@ Shortest distance between the points `p` and `q` on the [`Manifold`](@ref) `M`.
...
 
+"""
+    embed(M::Manifold, p)
+
+Embed point `p` from the [`Manifold`](@ref) `M` into the ambient space.
+The function works only for selected embedded manifolds.
+"""
+function embed(M::Manifold, p)
+    q = allocate_result(M, project, p)
+    project!(M, q, p)
```
mateuszbaran:
```suggestion
    embed!(M, q, p)
```

src/ManifoldsBase.jl
```diff
@@ -264,6 +264,54 @@ Shortest distance between the points `p` and `q` on the [`Manifold`](@ref) `M`.
...
+"""
+    embed(M::Manifold, p, X)
+
+Embed a tangent vector `X` at a point `p` on the [`Manifold`](@ref) `M` into the manifolds
+ambient space.
+
+The function works only for selected embedded manifolds.
+"""
+function embed(M::Manifold, p, X)
+    Y = allocate_result(M, project, X, p)
```
mateuszbaran:
```suggestion
    Y = allocate_result(M, embed, X, p)
```

src/ManifoldsBase.jl
```diff
@@ -264,6 +264,54 @@ Shortest distance between the points `p` and `q` on the [`Manifold`](@ref) `M`.
...
+    project!(M, Y, p, X)
```
mateuszbaran:
```suggestion
    embed!(M, Y, p, X)
```

src/ManifoldsBase.jl
```diff
@@ -264,6 +264,54 @@ Shortest distance between the points `p` and `q` on the [`Manifold`](@ref) `M`.
...
+"""
+    embed!(M::Manifold, Y, p, X)
+
+Embed a tangent vector `X` at a point `p` on the [`Manifold`](@ref) `M` into the manifolds
+ambient space. The result is saved in vector `Y`.
+
+The function works only for selected embedded manifolds.
+"""
+function embed!(M::Manifold, Y, p, X)
+    error(manifold_function_not_implemented_message(M, project!, Y, p, X))
```
mateuszbaran:
```suggestion
    error(manifold_function_not_implemented_message(M, embed!, Y, p, X))
```

kellertuer:
Thanks for noticing, sorry that I was too imprecise when copying interface functions.

mateuszbaran:
No problem!

PR #28, opened 2020-03-22
Clarify the real- and complex-valued basis difference.

kellertuer:
This also introduces a number_type for manifolds.

### DIFF
--- a/src/numbers.jl
+++ b/src/numbers.jl
@@ -52,3 +52,11 @@ end
 real_dimension(::RealNumbers) = 1
 real_dimension(::ComplexNumbers) = 2
 real_dimension(::QuaternionNumbers) = 4
+
+@doc raw"""
+    number_system(M::Manifold)
+
+return the number system a manifold is based on. The default of the number system is
+real-valued, i.e. `number_system(M) = ℝ`.
+"""
+number_system(M::Manifold) = ℝ
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "ManifoldsBase"
 uuid = "3362f125-f0bb-47a3-aa74-596ffd7ef2fb"
 authors = ["Seth Axen <seth.axen@gmail.com>", "Mateusz Baran <mateuszbaran89@gmail.com>", "Ronny Bergmann <manopt@ronnybergmann.net>", "Antoine Levitt <antoine.levitt@gmail.com>"]
-version = "0.7"
+version = "0.7.1"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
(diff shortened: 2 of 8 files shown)


src/numbers.jl
```diff
@@ -52,3 +52,11 @@ end
...
 real_dimension(::ComplexNumbers) = 2
 real_dimension(::QuaternionNumbers) = 4
+
+@doc raw"""
+    number_system(M::Manifold)
+
+return the number system a manifold is based on. The default of the number system is
+real-valued, i.e. `number_system(M) = ℝ`.
+"""
+number_system(M::Manifold) = ℝ
```
mateuszbaran: I'm not sure if this default is a good idea, you've already made an error due to this (`DefaultManifold` can be a complex manifold, depending on its number system argument).

src/numbers.jl
```diff
@@ -52,3 +52,11 @@ end
 real_dimension(::RealNumbers) = 1
 real_dimension(::ComplexNumbers) = 2
 real_dimension(::QuaternionNumbers) = 4
+
+@doc raw"""
+    number_system(M::Manifold)
+
+return the number system a manifold is based on. The default of the number system is
```
mateuszbaran:
```suggestion
Return the number system the manifold `M` is based on. The default of the number system is
```

src/numbers.jl
kellertuer: Thanks for noticing, that I missed implementing it for `DefaultManifold`. The idea for the default is, that manifolds, that do not have such a field are (implicitly) assumed to be real-valued. Though looking for an example I think this is only applicable for the `Torus`. I would like to make it easy for a user to have this by a default, since I think this function might be used quite often in the end.

I will add the function for `DefaultManifold` – but what do you think about the default? Is it useful or too dangerous?

mateuszbaran: OK, let's keep the default but also let's be careful about testing this method for each manifold.

By the way, what should one do if they want a real-coefficient basis of a product of a real and a complex manifold?

kellertuer: That depends on what coefficients you mean? The basis vectors? Then this should be an error since the second basis is complex.

But does the Product Basis have a field? It does not need one I think? It should be inferred automatically.

mateuszbaran: I mean real coefficients as in coefficients of linear combination of basis vectors. Could we assume that for a real Euclidean space the complex basis is the same as the real basis?

Currently product bases assume that all multiplied bases are over the same number system.

kellertuer: Well it is not that reasonable to have real euclidean with complex base, since it might yield contingent points as soon as you produce non real entries.

But despite that I think it is ok to assume that especially in a product manifold. So in that sense we can keep the current assumption (and not be too restrictive).

PR #29, opened 2020-03-23
Fixes a bug, where if both number_systems (manifold and base) are equal

kellertuer:
in this case non-real coefficients occured, which is not allowed.

### DIFF
--- a/src/EmbeddedManifold.jl
+++ b/src/EmbeddedManifold.jl
@@ -266,7 +266,21 @@ function decorator_transparent_dispatch(
     return Val(:parent)
 end
 function decorator_transparent_dispatch(
+    ::typeof(get_coordinates!),
+    ::AbstractEmbeddedManifold,
+    args...,
+)
+    return Val(:parent)
+end
+function decorator_transparent_dispatch(
     ::typeof(get_vector),
+    ::AbstractEmbeddedManifold,
+    args...,
+)
+    return Val(:parent)
+end
+function decorator_transparent_dispatch(
+    ::typeof(get_vector!),
     ::AbstractEmbeddedManifold,
     args...,
 )
--- a/src/EmbeddedManifold.jl
+++ b/src/EmbeddedManifold.jl
@@ -147,18 +147,6 @@ function check_tangent_vector(M::AbstractEmbeddedManifold, p, X; check_base_poin
         kwargs...
     )
 end
-"""
-    embed(M::AbstractEmbeddedManifold, p)
-
-return the embedded representation of a point `p` on the [`AbstractEmbeddedManifold`](@ref)
-`M`.
-
-    embed(M::AbstractEmbeddedManifold, p, X)
-
-return the embedded representation of a tangent vector `X` at point `p` on the
-[`AbstractEmbeddedManifold`](@ref) `M`.
-"""
-embed(::AbstractEmbeddedManifold, ::Any...)
 
 decorated_manifold(M::AbstractEmbeddedManifold) = M.embedding
 
--- a/src/ManifoldsBase.jl
+++ b/src/ManifoldsBase.jl
@@ -268,7 +268,11 @@ distance(M::Manifold, p, q) = norm(M, p, log(M, p, q))
     embed(M::Manifold, p)
 
 Embed point `p` from the [`Manifold`](@ref) `M` into the ambient space.
-The function works only for selected embedded manifolds.
+This method is only available for manifolds where implicitly an embedding or ambient space
+is given.
+Additionally, `embed` includes changing data representation, if applicable, i.e.
+if the points on `M` are not represented in the embeddings array data,
+the representation is changed accordingly.
 """
 function embed(M::Manifold, p)
     q = allocate_result(M, embed, p)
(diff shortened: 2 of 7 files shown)


kellertuer:
This has broken more things than I thought, I have to carefully look over the complex case again.

src/EmbeddedManifold.jl
```diff
@@ -265,13 +265,27 @@ function decorator_transparent_dispatch(
 )
     return Val(:parent)
 end
+function decorator_transparent_dispatch(
+    ::typeof(get_coordinates!),
+    ::AbstractEmbeddedManifold,
+    args...,
+)
+    return Val(:parent)
```
mateuszbaran: Why is this `:parent` instead of `:intransparent`? I think we will usually want people to implement this function for all embedded manifolds.

test/runtests.jl
```diff
@@ -3,7 +3,7 @@ using ManifoldsBase
 
 @testset "ManifoldsBase" begin
     # This should remain at 0
-    @test length(Test.detect_ambiguities(ManifoldsBase)) == 0
+    @test length(Test.detect_ambiguities(ManifoldsBase)) == 4
```
mateuszbaran: This needs to be reduced back to 0 before merging.

src/EmbeddedManifold.jl
kellertuer: I am nor yet 100% sure, the problem is, that you want to resolve the basis first.

test/runtests.jl
kellertuer: Yes, of course, it's currently ambiguities related to `get_coordinates`, since the bases types are now dispatched per number_type…

src/ManifoldsBase.jl
```diff
@@ -268,7 +268,11 @@ distance(M::Manifold, p, q) = norm(M, p, log(M, p, q))
     embed(M::Manifold, p)
 
 Embed point `p` from the [`Manifold`](@ref) `M` into the ambient space.
-The function works only for selected embedded manifolds.
+This method is only available for manifolds where implicitly an embedding or ambient space
```
mateuszbaran: Explicit embedding should be OK as well (through `AbstractEmbeddedManifold`).

src/ManifoldsBase.jl
```diff
@@ -268,7 +268,11 @@ distance(M::Manifold, p, q) = norm(M, p, log(M, p, q))
...
+is given.
+Additionally, `embed` includes changing data representation, if applicable, i.e.
+if the points on `M` are not represented in the embeddings array data,
```
mateuszbaran:
```suggestion
if the points on `M` are not represented in the same way as points on the embedding,
```

mateuszbaran: Similarly below.

src/ManifoldsBase.jl
mateuszbaran: Similarly below.

src/EmbeddedManifold.jl
mateuszbaran: What do you mean by resolving the basis first? We already have lots of disambiguation methods to make sure it's correctly resolved.

kellertuer: Well, the error was a `VeeOrthogonalBasis` that was not further “passed down” the hierarchy of basis (to orhogonal, then to orhonormal) for the mutating case. This resolving happens usually on the `Manifold` level (i.e. at parent). Of course we can also track down the case where something passes to mutating _before_ resolving the bases, but this way also works.

src/ManifoldsBase.jl
kellertuer: Yes, of course an explicit embedding is fine, too; this is the documentation of `Manifold` so I did not want to get into more specific types. For an general manifold, only the case of an implicitly assumed embedding is the explanation for this function.

src/EmbeddedManifold.jl
kellertuer: to be precise, it happens when you call `vee!` (or `vee` since that calls `vee!`) so we have to then “look” for the next basis (from `get_coordinates!`) that fits.

mateuszbaran: OK, that seems reasonable.

src/ManifoldsBase.jl
mateuszbaran: I understand now. I was a little confused because documentation for the non-mutating method on `Manifold` usually describes the entire function. It may be beneficial to point here to that section in `interface.md` for the case of an explicit embedding.

kellertuer: The entire function (like for `inner`) does not cover decorators. We could of course refer from inner to the `MetricManifold` as well, but how should such a link work in REPL?

mateuszbaran: Hm, right, REPL is not as nice here. I'd say then let's just add
```
See also: [`EmbeddedManifold`](@ref), [`project`](@ref)
```
at the end of the docstring.

kellertuer: Will do so (for embed, project, but also for inner) then.

mateuszbaran: Thanks!

Issue #30, opened 2020-03-23
Parametrize Manifold by its eltype/number system


kellertuer:
I would like to add a parameter to `Manifold`, i.e. its number_system. This would

* unify `Manifold` with `Bases`
* make a general implelemtation of ` numbersystem(M::Manifold{F}) = F` possible
* simplify subaccordingy a few implementations here and for Manifolds later on

What do you think?

mateuszbaran:
One reason why it'd be problematic is the product manifold. I'm already not quite sure how well product bases will handle bases of mixed real-complex manifolds. `ProductManifold` implements a lot of basis-related functionality that should make this work but maybe we should have a "Mixed number system" or something to indicate that no default fallback that relies on the number system will work?

kellertuer:
The `ProductManifold` should take the parent type (i.e. complex as soon as a complex manifold is involved) for example, but also “keep track” of its children's types and dispatch thereon correctly. 
For the elements of product it should still work fine. But basically we already have the same problem then for any basis of a `ProductManifold`? The basis is already parametrized by its number system.

Another possibility would of course be to forbid mixed manifold and say: If you take a complex one, please make them all complex. I think I'd prefere the first, mixed one.

mateuszbaran:
Yes, bases have the same problem and there it currently works by having to explicitly give the number system, and the number system is not reverted to real numbers for real manifolds. Product manifolds however store the component manifolds and parametrizing manifolds by number system would fix this problem.

I wouldn't like to restrict product manifolds to the have the same number system.

So, with that in mind, I think adding the number system to manifolds is a good idea 🙂 .

kellertuer:
I think that is a good restriction. So we should add the number system as a parameter to manifolds then.

mateuszbaran:
Resolved by #32

PR #32, opened 2020-03-28
Parametrize Manifold by number field

kellertuer:
This PR solves #30, though there is a point of discussion left.

Since the inner manifold of for example `EmbeddedManifold` should be the same as the embedded manifold (similar for array manifold), we have to type as follows

```julia
struct EmbeddedManifold{𝔽,MT<:Manifold{𝔽},NT<:Manifold,ET} <: AbstractEmbeddedManifold{ET,𝔽}
```

i.e. the field has to be placed before the manifold, while in all other cases (especially the abstract types, but also for the basic manifolds over at `Manifolds.jl`) the field is always the last parameter.
Note that eben for the abstract case, I went for the putting the field last.
 I feel this is a little inconsistent and should still be documented properly somewhere, if we do this change. 
What do you think? Is this inconcistency worth this (rather fundamental) change of the type?

Good news is, the main change despite this was, that all test manifolds had to be subtype to `Manifold{ℝ}` instead of just `Manifold`, but I feel that even makes the code a little nicer, since the definition directly helps to see whether the manifold is real or complex.

### DIFF
--- a/test/validation_manifold.jl
+++ b/test/validation_manifold.jl
@@ -18,15 +18,15 @@ function ManifoldsBase.injectivity_radius(
     return 11.0
 end
 
-@testset "Array manifold" begin
+@testset "Validation manifold" begin
     M = ManifoldsBase.DefaultManifold(3)
-    A = ArrayManifold(M)
+    A = ValidationManifold(M)
     x = [1.0, 0.0, 0.0]
     y = 1 / sqrt(2) * [1.0, 1.0, 0.0]
     z = [0.0, 1.0, 0.0]
     v = log(M, x, y)
-    x2 = ArrayMPoint(x)
-    y2 = ArrayMPoint(y)
+    x2 = ValidationMPoint(x)
+    y2 = ValidationMPoint(y)
     v2 = log(A, x, y) # auto convert
     y2 = exp(A, x, v2)
     w = log(M, x, z)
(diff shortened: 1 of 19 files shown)


mateuszbaran:
I think if we adapt the convention of putting size argument first, then number system, then everything else, there wouldn't be that much to change and it would be consistent.

kellertuer:
That's a good rule to follow. Basic (i.e. really implemented) manifolds will most commonly only depend on size and number system. If we unify this, we should also change the
`DiagonalizingOrthonormalBasis{B,ℝ}` to `DiagonalizingOrthonormalBasis{ℝ,B}`
and in the current approach I took only requires changes in the abstract cases, i.e. the
`AbstractEmbeddedManifold{T<:AbstractEmbeddingType,𝔽} <: AbstractDecoratorManifold{𝔽}`
should read
`AbstractEmbeddedManifold{𝔽,T<:AbstractEmbeddingType} <: AbstractDecoratorManifold{𝔽}`
but that's still nice, if it is unified afterwards.

kellertuer:
With full code coverage again, this is ready to review/merge.

kellertuer:
Why should we make such a release of Base? My plan was to add this to Manifolds once Sphere is done, and its also not that much to change, just copy more Numberfields to Manifold and remove parameters from where-clauses.

mateuszbaran:
That would be just in case of an urgent bug in Base. I think it's quite unlikely so it's also OK to merge it and tag 0.8 first.

kellertuer:
We can also leave this open until the sphere is really finished so I can merge& tag this when I start working on reworking Manifolds, maybe that's safe.

test/validation_manifold.jl
```diff
@@ -118,7 +118,7 @@ end
         @test injectivity_radius(A, x, CustomArrayManifoldRetraction()) == 11
```
mateuszbaran, line 118: Renaming `CustomArrayManifoldRetraction` to `CustomValidationManifoldRetraction` would make it a bit more consistent.

Issue #35, opened 2020-04-18
Remove non-implemented-errors?


kellertuer:
I recently stumbled upon this article (in preparation): https://github.com/oxinabox/oxinabox.github.io/pull/14/files#diff-e18e780852d70bbdfc8ecbe3aca97a23R22 on anti patterns.

The author writes, that writing own error messages is not useful, so instead of doing ( https://github.com/JuliaManifolds/ManifoldsBase.jl/blob/master/src/ManifoldsBase.jl#L369-L371 )

````julia
function exp!(M::Manifold, q, p, X)
    error(manifold_function_not_implemented_message(M, exp!, q, p, X))
end
````

we should just to

````julia
function exp! end
````

What I do not really like in this, is that you seem to not be able to provide arguments, so it is not so easy to document functions that we extend and that do not have default implementations (do we have such)?

The advantage is, that instead of `ErrorExceptions` `MethodErrors` are thrown, with the advantage, that the suggestions are provided in case you just missed a type. I even ran into this independently yesterday, so I wanted to discuss whether to do that.

The only point where I see a useful case for error messages is within the decorator dispatch.

What do you think?

mateuszbaran:
> What I do not really like in this, is that you seem to not be able to provide arguments, so it is not so easy to document functions that we extend and that do not have default implementations (do we have such)?

`manifold_dimension` is probably the only important example.

I don't have strong feelings about it either way. We avoid the main problem with error messages by making them quite explicit about types of arguments but they indeed don't seem to provide much value.

> The only point where I see a useful case for error messages is within the decorator dispatch.

Yes, and in that case they even could be longer, referring the user to documentation about decorators for help about figuring out what went wrong.

kellertuer:
I thought about this and indeed we might require the error messages due to our decorator dispatch.

PR #37, opened 2020-04-19
Introduce manifold_features and reduce ErrorMessages where MethodError is enough

kellertuer:
This PR works on #35 and uses the new freedom to implement a manifolds_features function.

While currently it returns tuples of the form
`(exp,[],1)` – if exp is implemented on the manifold or
`(log,[],0)` – if log is not implemented

further details can be provided in the `[]`, for example different retraction types or basis types.

Actually, just for the basis functions I am not yet 100% sure the checks are correct.

__Added paragraph__
One application I have in mind is, turning this into a table within the docs to get a fast overview which properties are available. One could to the same for decorators for manifolds, or something like whether its embedded or not.

Within this check, I removed a few manifold errors, where Method Errors should be enough. Here, I am not sure whether this is too much removed, for a few of above tests one gets Methods error due to ambiguities now.

Since this is breaking, we might have to increase more than just a patch.


mateuszbaran:
That's an interesting approach, I'll have to think about it.

kellertuer:
I think a good thing is, that if you just switch order of arguments somewhere you get hints, with the types, where your mistake was due to the `!Matched` signs in the print of similar signatures.

Though for one of the basis I checked you just might get an “is ambiguous” error from which you have to infer it is not implemented, so maybe then we should still provide an error?

mateuszbaran:
I have many minor issues with this code but I try to fix them at a later point. The main question right now, IMHO, is how many keyword arguments of `test_manifold` could be turned into calls of `manifold_feature`? BTW, I don't quite get the `specs` thing. Why is that specified only for `embed` and `project`? How do you imagine using it?

> Though for one of the basis I checked you just might get an “is ambiguous” error from which you have to infer it is not implemented, so maybe then we should still provide an error?

Which case is that?

mateuszbaran:
> Actually, the `manifold_feature`(`s`). is not meant to replace `test_manifold`, since it only checks whether a function exists; but if you mean we could automatically determine whether to do a test or not based on a check for the feature – that's a neat idea.

Yes, that's what I meant: automatic (or semi-automatic) determination of which tests should be done. Except then we'd have to be more careful since a test could fail because a method is incorrect (and we'd like to catch that). Probably inspection of the list generated by `methods(f)` would work better than just calling and looking if it errors.

> `specs` (maybe its not the best of names) is meant to distinguish different specialisations of a method. I misused that a little bit (but I think that's ok) to distinguish embedding/projecting a point or a tangent vector, since for example some manifolds only have functions to project tangents.
> The specs originally came to my mind to distinguish different retractions, so the vector of specs always carries the type of retraction; similarly for the basis, where I noticed that the `DiagonalizingOrthonormalBasis` could be added using `X` as a direction – though if `X` is zero that might not always work.

OK, I see. It's now a bit inconsistent between different functions. Probably tuples of argument types would work just fine for all functions.

> > > Though for one of the basis I checked you just might get an “is ambiguous” error from which you have to infer it is not implemented, so maybe then we should still provide an error?
> 
> > Which case is that?
> 
> I think that was either the `DefaultBasis` or the `DefaultOrthonormalBasis` for the `Sphere(2)`, we might also have lost a default-pass-down somewhere with basis?

I wouldn't be surprised if we actually lost something there, it's quite complex.

kellertuer:
I split that because In the result I just wanted to report on functions and their specialisations and not always return the full signature, but we can of course also return the full signature, if you prefer.

mateuszbaran:
I don't really see how to make this approach robust. Just because a function didn't throw an error doesn't mean we think it's correct. It'd have to be coupled to to the test suite but then the approach of checking for errors wouldn't work. Neither would `hasmethod`. Why not simply gather keyword arguments to `test_manifold` in a single feature list?

kellertuer:
`manifold_feature` does not say, a certain feature (function + type, like a retraction method) is _correct_ it just says “it exists”. For the docs I want to document what exists.

I like the idea to automatically test features that exist and I think with my idea that can be done. But this was first not meant for that (I still think it can be used for that) but it should also not be mixed with testing or include testing. the feature is just meant to test existence.

The disadvantage of knolling together the keyword arguments is, that that's done in the test suite and its done manually.

What do you think will not be doable robustly?

mateuszbaran:
Hmm... what would be the intended usage of that "it exists" information? I can see that this approach would do fine for this. Is it just a small helpful thing for us to see which things don't throw an error but should? Or which tests are missing?

kellertuer:
1) (my main intention) generate an automated markdown table (my idea is right aligned atop the page of a manifold) which features are available. One could extend that (later) do display properties (injectivity radius, maybe?)

2) (from your comments) remove most of the kwargs of `test_manifold` and switch to a scheme that does “test it if it exists”

and yes 3) we could also test a little bit which features are reported to exist but maybe somehow actually do not exist ;)

mateuszbaran:
As I realized, (2) won't work unfortunately. Now I think the right approach would be to put in `test_manifold` tests that check that `manifold_feature` returns correct information. Then we would be fairly sure that documentation generated based on `manifold_features` contains correct information.

kellertuer:
What do you mean by “right information”? What if a certain feature is (for what reason ever) not yet covered by tests? It is still a feature then. So for me `manifold_feature` (which should maybe really just return true/false) being true does not mean all tests (or even only one test) is run on that function. Just that it exists.

Why would 2 not work?

For retractions, for loop over all existing types, if that type is implemented, check the usual checks. Is the result on the manifold? If the inverse exists, is it the inverse? Something like that can be done automatically based on the usual 3 points we pass to test_manifold.

mateuszbaran:
> What do you mean by “right information”? What if a certain feature is (for what reason ever) not yet covered by tests? It is still a feature then. So for me `manifold_feature` (which should maybe really just return true/false) being true does not mean all tests (or even only one test) is run on that function. Just that it exists.

Hmm... maybe it would make sense to indicate which features are covered by tests and which are not?

> Why would 2 not work?

My original idea was to determine things like `test_project_point` or `test_vector_transport` through `manifold_feature` but if it relies on testing for errors, an incorrect implementation would just disable a test and it would be relatively hard to spot it without looking at code coverage or generated docs. Lots of our features are covered only by `test_manifold`. How would you spot a random regression that manifests in an `ErrorException`? Right now a test would just fail. With this change, a test would be disabled and we'd have to either compare feature tables between versions or look carefully at changes in code coverage (and right now code coverage takes a long time to settle). At least I don't see any other option.

> For retractions, for loop over all existing types, if that type is implemented, check the usual checks. Is the result on the manifold? If the inverse exists, is it the inverse? Something like that can be done automatically based on the usual 3 points we pass to test_manifold.

Yes, that's reasonable.

kellertuer:
> > What do you mean by “right information”? What if a certain feature is (for what reason ever) not yet covered by tests? It is still a feature then. So for me `manifold_feature` (which should maybe really just return true/false) being true does not mean all tests (or even only one test) is run on that function. Just that it exists.
> 
> Hmm... maybe it would make sense to indicate which features are covered by tests and which are not?

That means running all tests during docs generation, too? That would make generating docs quite long – for what benefit? I am not sure.

> 
> > Why would 2 not work?
> 
> My original idea was to determine things like `test_project_point` or `test_vector_transport` through `manifold_feature` but if it relies on testing for errors, an incorrect implementation would just disable a test and it would be relatively hard to spot it without looking at code coverage or generated docs. Lots of our features are covered only by `test_manifold`. How would you spot a random regression that manifests in an `ErrorException`? Right now a test would just fail. With this change, a test would be disabled and we'd have to either compare feature tables between versions or look carefully at changes in code coverage (and right now code coverage takes a long time to settle). At least I don't see any other option.

This is indeed a problem; is there a way to check whether a certain function _signature_ exists? Currently I am checking for existence by trying to call it. This only yields true if the signature is valid _and_ the code runs correctly. I would be fine with just checking the first (in `manifold_feature`). That would avoid this trouble – and testing whether it runs fine is then a matter of a test.

> > For retractions, for loop over all existing types, if that type is implemented, check the usual checks. Is the result on the manifold? If the inverse exists, is it the inverse? Something like that can be done automatically based on the usual 3 points we pass to test_manifold.
> 
> Yes, that's reasonable.

And I hoped to be able to do this for all functions – maybe together with the signature test it should still be a valid way? One could provide keywords to mark tests that do fail, maybe, i.e. instead of `test_exp` we would have that exp is checked if the signature exists and if you set `exp_broken` to true it is expected to fail? If we manage to get this fancy enough we can just collect a set of signatures of functions that are currently broken.

mateuszbaran:
> That means running all tests during docs generation, too? That would make generating docs quite long – for what benefit? I am not sure.

The checks could be run during testing, although then `manifold_feature` would be significantly more complicated. I'm not really satisfied with that idea, I'm just looking for possible ways to couple `manifold_feature` with `test_manifold`. Maybe it's not a good idea after all.

> This is indeed a problem; is there a way to check whether a certain function _signature_ exists? Currently I am checking for existence by trying to call it. This only yields true if the signature is valid _and_ the code runs correctly. I would be fine with just checking the first (in `manifold_feature`). That would avoid this trouble – and testing whether it runs fine is then a matter of a test.

> edit: Ha! There is http://mortenpi.eu/documenter-html/stdlib/base.html#Base.method_exists ! I will revise my code after teaching today.

I don't see that function in Julia 1.4.1, maybe it's something old? This is not the official documentation. Julia has the `hasmethod` function for that, although forwards generated by our decorator framework would also count there. We could also directly inspect results of the `methods` function, then the forwards could just be discarded.

> This is indeed a problem; is there a way to check whether a certain function _signature_ exists? Currently I am checking for existence by trying to call it. This only yields true if the signature is valid _and_ the code runs correctly. I would be fine with just checking the first (in `manifold_feature`). That would avoid this trouble – and testing whether it runs fine is then a matter of a test.

Then I think direct inspection of results of `methods` would be the right approach. For example looks at the result of `methods(manifold_dimension).ms[1].sig`.

kellertuer:
I think I would still prefer to split the two things
* existence of a functon
* testing of a function

feature aims just to provide existence not testing/correctnes; testing could be automated more using this existence.

And ah, the function is now called https://docs.julialang.org/en/v1/base/base/#Base.hasmethod or one could use https://docs.julialang.org/en/v1/base/base/#Core.applicable – but I can also look closer at `methods`, but I think forwards are fine. I mean, inner for the sphere exists (via forwarding to the embedding).

mateuszbaran:
> but I think forwards are fine. I mean, inner for the sphere exists (via forwarding to the embedding).

That's an example where forwards are fine. The `:intransparent` cases that raise `ErrorException` wouldn't be.

Also when a manifold doesn't have `exp!` (like `FixedRankMatrices`), `exp` is still technically implemented (has a default that calls `exp!`) but I wouldn't say that feature exists.

The logic could be something like:
1. Does the function have a method specifically for that manifold? If yes, return `true`.
2. Is this a decorator and function is `:parent` or `:transparent`? If yes, go recursively to parent or wrapped manifold. Check whether functions the fallback relies upon are implemented.
3. Otherwise return `false`.

I may have missed something.

kellertuer:
No that actually sounds quite good, I will try to follow that path then.

kellertuer:
For the intransparent case, it is always still applicable so there I tried calling the function then.

kellertuer:
I think I nearly have your approach, just for the intransparent case I could not figure out how to check if the exact method exists or if only the fallback to an error exists – only with a try again.

Further – currently I am unable to fix the remaining bugs since calling the feature function just makes Julia crash and I can't see where that comes from and none of my tries to see where that happens, was successful.

kellertuer:
While in the long run I would like to have a function like this, that prints a nice list of features of a manifold, I think this approach is too complicated and too less “Julia-ish”, so I'll close this and restart, when I have a better idea how to check for existing features/functions.

PR #39, opened 2020-04-28
vector_transport_direction! should be :parent (it has a fallback)

mateuszbaran:


mateuszbaran:
Fixes a problem discovered in https://github.com/JuliaManifolds/Manifolds.jl/pull/177 (failure in vector transport on `Hyperbolic`).

PR #40, opened 2020-05-04
Introduce Schilds and Pole ladder

kellertuer:
This PR aims to provide two further methods for vector transport.
The initial implementation here is in separate functions. What I would like to discuss

* Should we do a type `SchildsLadder()` for `vector_transport_to` (and similarly `PoleLadder()`)?
  a disadvantage for pole might be that pole is efficient for multiple applications but you have to be able to pass in the center. So if we introduce these, I would still vote to keep the original functions in the background.

* We could also provide with both approximate methods to transport along an arbitrary curve (e.g. a `t -> geodesic(p,X,t)` if we introduce a `DiscreteTransport(s)` where `s` can be the number of equidistant points (on [0,1]) for discretisation _or even_ a `range()` to use. Then one would to from each sample point to the next a Schild's (or pole) ladder.
  What to you think?
 _Edit_: Since https://github.com/geomstats/geomstats/blob/c70efc4da90bc277b5d6befcefd887b3432d9f98/geomstats/geometry/connection.py#L341 is the same approach along a geodesic, but one can even to that along any curve (approximating the discrete points with connecting geodesics), I would say – yes we should to that. Maybe with a point `DiscreteVectorTransport(method, range)` even. Since for the small steps one could then also use `ProjectionTransport` or such, I think.


mateuszbaran:
> * Should we do a type `SchildsLadder()` for `vector_transport_to` (and similarly `PoleLadder()`)?
>   a disadvantage for pole might be that pole is efficient for multiple applications but you have to be able to pass in the center. So if we introduce these, I would still vote to keep the original functions in the background.

I think we should, in some form. We can have a separate function for a single step of the pole ladder that takes the center and can be reused elsewhere.


> * We could also provide with both approximate methods to transport along an arbitrary curve (e.g. a `t -> geodesic(p,X,t)` if we introduce a `DiscreteTransport(s)` where `s` can be the number of equidistant points (on [0,1]) for discretisation _or even_ a `range()` to use. Then one would to from each sample point to the next a Schild's (or pole) ladder.
>   What to you think?
>   _Edit_: Since https://github.com/geomstats/geomstats/blob/c70efc4da90bc277b5d6befcefd887b3432d9f98/geomstats/geometry/connection.py#L341 is the same approach along a geodesic, but one can even to that along any curve (approximating the discrete points with connecting geodesics), I would say – yes we should to that. Maybe with a point `DiscreteVectorTransport(method, range)` even. Since for the small steps one could then also use `ProjectionTransport` or such, I think.

I definitely want to have that but we need to have a good API. I'll write some thoughts about it later.

mateuszbaran:
OK, after thinking about it a bit, there is a few things that need to be done:
1. I think functions for single steps of these ladders should work on just points, that is `exp(M,p,X)` is given instead of `X`. Similarly, only `exp`s of transported vectors should be returned.
2. Transport along arbitrary curves would be nice with adaptive selection of discretization with some accuracy parameters which we are not planning to do here (I think?). I suggest just passing an array that corresponds to a point on a `PowerManifold` (with either array or nested representation of points) that discretizes the curve at knots selected by the user. Vector would be transported between successive points in that discretization using a specified scheme.
We could also provide some automatic discretization of curves inside vector transport but anyway to me it looks like the interface with explicit passing of points to transport between should be done first.
3. In your Schild's ladder step you use `shortest_geodesic` with parameter 2. Previously my thinking was that `shortest_geodesic` is not required to be defined for parameters outside [0,1] even if a unique extension exists. I prefer explicitly using `log`, multiplication by 2 and `exp`  there.

kellertuer:
Thanks for this thoughtful response. Just to clarify (before I start coding later)

> 1. I think functions for single steps of these ladders should work on just points, that is `exp(M,p,X)` is given instead of `X`. Similarly, only `exp`s of transported vectors should be returned.

So the ladders should just work on the points? We would do `schilds_ladder(M, p, s, q)` ? The vector transports would then have to do the previous exp (I wasn't sure, maybe one could replace this with a retraction, computationally definitely, analytically I am not 100% sure) and the subsequent log (inverse retraction).

> 2. Transport along arbitrary curves would be nice with adaptive selection of discretization with some accuracy parameters which we are not planning to do here (I think?). I suggest just passing an array that corresponds to a point on a `PowerManifold` (with either array or nested representation of points) that discretizes the curve at knots selected by the user. Vector would be transported between successive points in that discretization using a specified scheme.
>    We could also provide some automatic discretization of curves inside vector transport but anyway to me it looks like the interface with explicit passing of points to transport between should be done first.

I would make the discretisation a property of the transport, since we anyways have a `DiscretizedVectorTransport(ds, method)` (which mould even be done with any method actually, maybe excluding discretised transport itself), so the first simple one could just take a set of points (or from a power manifold) or a geodesic and a set of timepoints. The idea would be that the starting point is not necessary since its known for the transport to be `p` and the remaining are either stored in discretised transport or could later be incrementally determined by accuracy parameters. We would have different discretization-schemes `ds` for that.

> 3. In your Schild's ladder step you use `shortest_geodesic` with parameter 2. Previously my thinking was that `shortest_geodesic` is not required to be defined for parameters outside [0,1] even if a unique extension exists. I prefer explicitly using `log`, multiplication by 2 and `exp`  there.

Yes, for sure, no problem.

kellertuer:
* I addressed (1), the only thing I don't like that much is, that the resulting approximate transport is -log (not log) for pole so that takes an explanation everywhere.
* I fixed (3)
* I introduced the vector transports – that now not only call the ladders but have to do an initial exp and a final (-)log.
* In order for this to not escalate (well we often do that, I know), I could do the discretised in a separate (next) PR? If so, this mainly requires tests.

mateuszbaran:
> So the ladders should just work on the points? We would do `schilds_ladder(M, p, s, q)` ? The vector transports would then have to do the previous exp (I wasn't sure, maybe one could replace this with a retraction, computationally definitely, analytically I am not 100% sure) and the subsequent log (inverse retraction).

Yes, I think this should reduce the total number of `exp`s and `log`s performed during a transport along a curve. Anyway, single-step transport should be in-place and non-allocating. So temporaries should be re-used between steps when transporting along a curve.

> * I addressed (1), the only thing I don't like that much is, that the resulting approximate transport is -log (not log) for pole so that takes an explanation everywhere.

Doesn't that actually make sense? When pole-transporting a vector along a curve it seems to make sense to not flip the vector at each step. You just flip it at the end (or not) depending on how many steps you've made. I think you should start with transporting along a curve and then work out the right set of arguments for functions that perform single step. The single-step function looks pretty much an internal thing so it makes sense to make it return whatever is convenient.

Unless I misunderstood something.

> * In order for this to not escalate (well we often do that, I know), I could do the discretised in a separate (next) PR? If so, this mainly requires tests.

Transport along a discretized curve would require redoing the single-step functions you've written so I think it makes sense to make both single-step and transport along a discretized curve in one PR.

> _Edit_: for now I did not generalise this to retractions, because I am not yet sure, where to put that in the vector transport – it should then become two fields within the `SchildsLaddertransport`/`PoleLadderTransport` (retraction_type and inverse_retraction_type)?

Yes, I think it makes sense to put retraction and inverse retraction in `SchildsLaddertransport`/`PoleLadderTransport`.

> I would make the discretisation a property of the transport, since we anyways have a `DiscretizedVectorTransport(ds, method)` (which mould even be done with any method actually, maybe excluding discretised transport itself), so the first simple one could just take a set of points (or from a power manifold) or a geodesic and a set of timepoints. The idea would be that the starting point is not necessary since its known for the transport to be `p` and the remaining are either stored in discretised transport or could later be incrementally determined by accuracy parameters. We would have different discretization-schemes `ds` for that.

`DiscretizedVectorTransport(ds, method)` could go in a separate PR I think, I'd start with transporting along an explicit collection of points.

kellertuer:
> Yes, I think this should reduce the total number of `exp`s and `log`s performed during a transport along a curve. Anyway, single-step transport should be in-place and non-allocating. So temporaries should be re-used between steps when transporting along a curve.

That's a good point, I will add non-allocating ones (or at least try to), then. For single step they could maybe be used, too.
> > * I addressed (1), the only thing I don't like that much is, that the resulting approximate transport is -log (not log) for pole so that takes an explanation everywhere.
> 
> Doesn't that actually make sense? When pole-transporting a vector along a curve it seems to make sense to not flip the vector at each step. You just flip it at the end (or not) depending on how many steps you've made. I think you should start with transporting along a curve and then work out the right set of arguments for functions that perform single step. The single-step function looks pretty much an internal thing so it makes sense to make it return whatever is convenient.

Oh, maybe you read that wrong? In `pole_ladder` I don't do that, so that does not do the minus,
For iterative steps that means you have to track the number of steps to know whether its log to -log: for odd ones, so after 1 or 3 steps its minus for even its log, I think, If the image in my head is correct.
In terms of returning that's really not important, I just got a little clumsy in explaining the details. If we declare them merely intern I would move the references to either the type of the vector transport (better there transport) and the internal ones get a nice shorter doc. That would solve the problems I had for me, too

> Transport along a discretized curve would require redoing the single-step functions you've written so I think it makes sense to make both single-step and transport along a discretized curve in one PR.

Ok Will do so then, next. Since the above point with internal is clear for me; maybe tomorrow evening.

> 
> > _Edit_: for now I did not generalise this to retractions, because I am not yet sure, where to put that in the vector transport – it should then become two fields within the `SchildsLaddertransport`/`PoleLadderTransport` (retraction_type and inverse_retraction_type)?
> 
> Yes, I think it makes sense to put retraction and inverse retraction in `SchildsLaddertransport`/`PoleLadderTransport`.

Cool, I will also adapt internal functions and the types then.
 
> `DiscretizedVectorTransport(ds, method)` could go in a separate PR I think, I'd start with transporting along an explicit collection of points.

but that would still be exactly a `DiscretizedVectorTransport(ds,method)` just that `ds` is a vector of points. I would not go to a product manifold I think; similar to the mean where we have vectors of points, that should suffice here, too. But ok, adding a version along a given (discrete) path in this PR is also great. Thanks for the comments. I think I know where this should go then.

mateuszbaran:
> but that would still be exactly a `DiscretizedVectorTransport(ds,method)` just that `ds` is a vector of points. I would not go to a product manifold I think; similar to the mean where we have vectors of points, that should suffice here, too. But ok, adding a version along a given (discrete) path in this PR is also great. Thanks for the comments. I think I know where this should go then.

OK, maybe I misread something but somehow I understood that `ds` is supposed to be time points on a curve, not samples. If it's samples then it's OK. BTW, I'd like to have both power manifold representations available for vector transport (nested and array). If you prefer, you do it any way you like and then I'll make necessary changes to support both versions. That's because I'll have lots of curves in the array representation and it would be pretty inefficient having to convert them to the nested representation for vector transport.

Thanks for working on this, it'll definitely be useful to me too :slightly_smiling_face: .

kellertuer:
I think I finishd the pole and schild ladder (up to testing), but I got a little stuck when starting the type for the `DiscretizedVectorTransportMethod`.

Currently we can't do that on the power manifold, since that is not in Base (yet?). I would love to see it here, but until then we might have to do just the vector-of-points approach.
Also I have to find a way that usually this dos susequent vector transports, but for pole&schild it does instead just the inners (and an exp in front and a corresponding log in the end, for odd pole-iterations with a minus).

mateuszbaran:
Maybe we could do `DiscretizedVectorTransportMethod` in Manifolds.jl for now?

kellertuer:
I would prefer a vector based one here,
und a power manifold based one in manifolds? You might have to pass the power manifold then anyways (for nested one could get that from the size but for array I am not 100% sure?) in order to do `s[M,i]` where s is said vector.

mateuszbaran:
Having `DiscretizedVectorTransportMethod` the function `vector_transport_along` seems to me a bit redundant.

mateuszbaran:
Or maybe we should use `DiscretizedVectorTransportMethod` for `vector_transport_along`?

mateuszbaran:
I have extracted stuff related to vector transport to a separate file (just like bases are separate), the main file was getting too long. I have also very roughly sketched how I think we could support discretized transport for points given as a vector in Base and extend it to any power representation in Manifolds.jl. What do you think?

kellertuer:
Oh, you'Re right, it should more be something for the `c` parameter of `vector_transport_along`, i.e. maybe more a `DiscretizedCurve` type?

mateuszbaran:
I think `DiscretizedVectorTransport` is fine, I'd prefer to have a `DiscretizedCurve` type separately (also it could be given in the `points` field of `DiscretizedVectorTransport`).

kellertuer:
but you are also right that the `DiscretizedVectorTransport` is equivalent with `transport_along`, isn't it? I would maybe commit the `DiscretizedTransport since it is more generally along with a discretised curve.

mateuszbaran:
> but you are also right that the `DiscretizedVectorTransport` is equivalent with `transport_along`, isn't it?

Yes, that's what I think `transport_along` was supposed to be doing. 

> I would maybe commit the `DiscretizedTransport since it is more generally along with a discretised curve.

At this point I'm not even fully sure what a `DiscretizedCurve` is supposed to store inside. Many operations on discretized curves will happily work on arrays while some will not, see `AbsrtractCurve` here: https://github.com/JuliaManifolds/Manifolds.jl/pull/110 . Some operations require also parameters at which the curve was sampled, finally some would like a function that interpolates (with cached interpolation coefficients). Should discretization points be a parameter of the manifold of discretized curves? If not, how do we quickly check if two curves were discretized at the same parameters? So many questions.

kellertuer:
I have the following types of curves in my mind

* `SampledCurve` – which maybe really does not require a type since its really just a vector of points, it could be implemented that way (with vector) here
* `DiscretizedCurve` which is a function `t -> M` that maps from time to manifold as well as a set of timepoints where to sample at
* `DynamicallyDiscretizedCurve`, where you specify a characterization, i.e. how far two points should be apart and its automatically sampled that way (but otherwise is similar to the last one),

and yes they would all be `AbstractCurves` in your sense.
The simplest one (vector even without a special type) I would put here to provide a basic implementation.
More efficient ways (PowerManifold and array representation for example) could then be done in Manifolds.jl based on AbstractCurve Types – Maybe a `PowerCurve` that stores a power manifold and a point on that as an alternative to the simple vector from here?

mateuszbaran:
OK, so maybe let's keep this PR simple and just provide transport along a sampled curve? We can work on more complex curve types later.

kellertuer:
I implemented the `along` type – and I think I would prefer this over the to with an discretised curve, the function sounds strange - one says to and provides a path? That's along. I am not sure my version got as efficient as your discretised version,
but I would vote to remove the discretised curve and stay with the simple vector version. If you can enhance my implementation to be more safe and/or memory efficient, feel free to do so.

mateuszbaran:
I'll take a look at your code and see what can be improved 🙂 .

mateuszbaran:
I have made some small improvements, I think the main things to do here are:

- [ ] some test
- [ ] looking into reducing allocations
- [ ] consistent usage of retractions and inverse retractions
- [ ] checking if `AbstractPointSequence` provides sufficient extensibility

kellertuer:
Great! This is at first glance what I was thinking of. My only remark is: Is the new type for the vector of points really necessary? I had hoped that any vector, i.e. `Vector{T}` should suffice assuming that the basic case here (without PowerManifold) is basically a nested approach.

mateuszbaran:
Maybe I could make it so that a `Vector` of points (or maybe even an `AbstractVector`?) would also work, I'll think about it.

If my idea works maybe I could even make a nice and fast wrapper for flat array-represented points on `PowerManifold`.

kellertuer:
Yes, my initial thought was `AbstractVector{T}` :) 

My hope is (1) that the extension to nested is automatic; but (2) an extension to using `p[M,i]` on a Power Manifold is hopefully also easy (in Manifolds.jl, until we decide to Move Power/Product here).

kellertuer:
I just did some simple tests for pole and schild, including the ValidationManifold, though its the easy ones: exactness on Euclidean. But we also have no Hyperbolic space (where they still should work and pole should be exact at least).

Missing are the `_along` tests which also include the decision, whether just `AbstractVector{T}` suffices.

mateuszbaran:
> I just did some simple tests for pole and schild, including the ValidationManifold, though its the easy ones: exactness on Euclidean. But we also have no Hyperbolic space (where they still should work and pole should be exact at least).

Before merging this I think we should make some new tests for vector transports in `Manifolds.jl` and see if they work correctly. I wouldn't be too concerned about only having it tested in Manifolds.jl and not here.

> Missing are the `_along` tests which also include the decision, whether just `AbstractVector{T}` suffices.

I'll work on it a bit today.

kellertuer:
Well here I want to make sure that then at least the basics work, so the Euclidean case (which I covered therefore).

mateuszbaran:
Yes, sure, there should be a test for the Euclidean case here.

mateuszbaran:
I've checked and it turns out we can use `AbstractVector` in `vector_transport_along`.

src/vector_transport.jl
```diff
@@ -0,0 +1,636 @@
...
+)
+    d = exp(M, p, X)
+    m = p
+    clen = length(c)
+    for i in 1:(clen - 1)
+        # precompute mid point inplace
+        ci = c[i]
+        cip1 = c[i + 1]
+        log!(M, Y, ci, cip1)
+        exp!(M, m, ci, Y / 2)
```
mateuszbaran, line 417: Do we need to use `log!` and `exp!` here instead of retraction and inverse retraction from `method`?

mateuszbaran: Similarly in other cases below.

kellertuer: I think so since we need the mid point.

mateuszbaran: OK, in this case it makes sense, but what about other `exp`s and `log`s? for example in lines 409, 431, 458, 480?

mateuszbaran: BTW, maybe we should add `mid_point` and `mid_point!` that you have in Manopt.jl so that it can be used here?

kellertuer: I would actually be in favour of that, yes :)

For the remaining lines:
* 409, 431 – well since in between we use retractions to move around, we could do that there too,
* similarly 458 and 480
* but... 579, 586, 602, and 609 should then be adapted, too.

For all of those (even the retracts we already use) I am not sure about theory, but yes, _if_ we leave log/exp and generalise we could do for all non-mids.

mateuszbaran: OK, I'll add `mid_point` and adapt these lines.

mateuszbaran: I think it should work correctly after the last commit.

src/ManifoldsBase.jl
```diff
@@ -545,6 +545,42 @@ function manifold_function_not_implemented_message(M::Manifold, f, x...)
...
+
+"""
+    mid_point!(M::Manifold, q, p1, p2)
+
+Calculate the middle between the two point `p1` and `p2` from manifold `M`.
+By default uses [`log`](@ref), divides the vector by 2 and uses [`exp!`](@ref).
+Saves the result in `q`.
+"""
+function mid_point!(M::Manifold, q, p1, p2)
+    X = allocate_result(M, log, p1, p2)
```
kellertuer, line 567: Is this maybe overdoing the „avoid-allocation and pass a variable down“ approach? I already thought that on the pole and schild; do you think that's really necessary? If so, it can stay of course.

mateuszbaran:
I think there could be a bit more done about allocations but I'm OK with doing it in another PR. Now the only thing that remains to be done here is trying to use these vector transports in Manifolds.jl.

src/ManifoldsBase.jl
mateuszbaran: I can't be 100% sure without some benchmarks but ultimately different things will dominate the calculation for different manifolds. I'll remove the 5-argument `mid_point!` for now then.

kellertuer: If we notice a huge Time spent on allocations we can surely add that again. What do you think about the X-allocation in pole and schild? For them I thought the same, but they are also called far more often in the for loop, so I would be fine with keeping those.

mateuszbaran: Which X-allocation? There are a few ( `X*=2` is an interesting one to solve) but I think it's OK to take care of allocations later.

kellertuer: Let's keep them as they are, `schilds_ladder!` (and pole) are merely internal functions anyways (as opposed to `mid_point`).

kellertuer:
There was a initial step missing in the `_along` methods for pole and schild, namely the transport from p to c[1]. I added that and had hoped to cover most new lines, let's see how that turns out.

kellertuer:
Though I wrote `_ along` tests they don't seem to really work looking at the coverage. Despite this I think this PR is finished.

mateuszbaran:
Isn't the `_along` function only called for `DefaultManifold` which has a specialized method that only does `copyto!`?

kellertuer:
Ah! We should then remove that maybe?

mateuszbaran:
We can either remove that method or also test `_along` methods using another manifold.

kellertuer:
I'd prefer to remove that default, since we now have a good implementation of that (using transport_to, which is implemented).

kellertuer:
Ah, I missed that you introduced that default, so I indeed wrote a short Manifold without that default but all necessary functions to check the generic manifold `vector_transport_along`. and even found a small error for the `m` that needs to be allocated, otherwise `p` was changed.

kellertuer:
This PR ha just the 4 lines left that were uncovered before and I can't see what codecov does there, so I think this PR is ready to merge.

mateuszbaran:
I would prefer to have https://github.com/JuliaManifolds/Manifolds.jl/pull/187 in a mostly working condition before this is merged, just to be sure there are no obvious design problems here.

I think we could also remove the default implementation of projection transport in this PR: https://github.com/JuliaManifolds/Manifolds.jl/issues/172

kellertuer:
Good idea. Again a job for this evening.

And of course, we can wait for the new PR and do a combined-merge then.

src/vector_transport.jl
```diff
@@ -598,17 +598,8 @@ projection [`project`](@ref project(M::Manifold, p, X)) may also change vector
...
-        M,
-        vector_transport_to!,
-        Y,
-        p,
-        X,
-        q,
-        m,
-    ))
-end
+vector_transport_to!(M::Manifold, Y, p, X, q, m::ProjectionTransport)
```
kellertuer, line 601: Ah! Another good reason to carefully check for #35 :)

mateuszbaran: Yes, we should do that eventually.

kellertuer:
I checked for the circle mean. My method yields the same numbers (up to 12th digit precision in display) but seems more stable with respect to the weighted case. So I have to look for a reference, where I got my formula from...

Issue #41, opened 2020-05-04
Change reductions of generators to mapreduce


mateuszbaran:
It doesn't really matter in Julia 1.4 but somehow the performance is significantly worse on nightly.

PR #43, opened 2020-05-07
Small cleanup

### LINKED ISSUE #34: Should representation_size return nothing by default?
As in the title, I think this would allow for better support of manifolds that don't have `representation_size`.

mateuszbaran:
Resolves #34 and #41


mateuszbaran:
I think we can wait with tagging 0.9.0 until #40 is merged.

PR #44, opened 2020-05-08
Github Actions

kellertuer:
This PR introduces GitHub Actions for testing and format. Does the Compat require a pair of Keys (public to deploy, private as secret)? Then I could add compat in this PR, too.

### DIFF
--- a/.github/workflows/CompatHelper.yml
+++ b/.github/workflows/CompatHelper.yml
@@ -1,0 +1,20 @@
+name: CompatHelper
+
+on:
+  schedule:
+    - cron: '00 00 * * *'
+
+jobs:
+  CompatHelper:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: julia-actions/setup-julia@latest
+        with:
+          version: 1.4
+      - name: Pkg.add("CompatHelper")
+        run: julia -e 'using Pkg; Pkg.add("CompatHelper")'
+      - name: CompatHelper.main()
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          COMPATHELPER_PRIV: ${{ secrets.COMPATHELPER_PRIV }}
+        run: julia -e 'using CompatHelper; CompatHelper.main()'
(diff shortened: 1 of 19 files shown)


mateuszbaran:
Thanks, using Github actions is a good idea. I don't know if Compat requires keys but codecov almost certainly does (if we want it to be triggered by Github CI).

kellertuer:
Codecov just needs a secret that I already added.

kellertuer:
This should be good to go. I activated all actions, also CompatHelper with its open issues SSH Key; the formatter runs on the current state with no changes.

Two things I noticed, formatter does not like

* `using A.b` (where we actually overwrote the behaviour so import is better) becomes
```julia
using A:A
using b:b
```
which is strange
* if the last line(s) of a function are an if like
```julia
[...]
    if A
        return B
    else
        return C
    end
end
```
formatter snugs a `return upfront the `if`. For now I deactivate formatting for the two functions/macros where we do that.

The tests now are all running, so this is finished and waits for approval. Once merged, Travis & AppVeyour can be deactivated. As once can see Codecov was triggered and at time of writing Appveyor/Travis are not yet finished, so I am quite sure that's from the actions.

mateuszbaran:
Thanks, it looks quite good now. For me `using A: something` isn't weird actually, I'd keep it. `return` before `if`/`else` is weird but it doesn't break anything so we can change it back to not having that `return` when https://github.com/domluna/JuliaFormatter.jl/issues/233 is fixed.

BTW,
```
return if something
    return 3
end
```
actually would make (a bit of) sense because the `if` isn't guaranteed to return (it wouldn't if the condition is false).

kellertuer:
Og it was `import A.b` that breaks (gets `using A: A`(newline)?using b:b`), sorry, I was a little fast with my Investigation here, but opened an issue over at JuliaFormatter.

Concerning the return – I also opened an issue – it seems to work since the inner returns return something and the outer is never triggered – but it also looks ugly. For the two functions where we have that I for now disabled formatting.

mateuszbaran:
You shouldn't bump the version, 0.9 wasn't released yet. I was waiting for #40 since it's also breaking.

kellertuer:
Hm, I see, ok.

PR #47, opened 2020-08-05
fixing mid_point on Julia 1.0

mateuszbaran:


test/runtests.jl
```diff
@@ -4,7 +4,9 @@ using ManifoldsBase
 
     num_ambiguities = length(Test.detect_ambiguities(ManifoldsBase))
```
kellertuer, line 5: In Manifolds even the call here failed, so would it maybe be better to even put this line into the version test?

src/ManifoldsBase.jl
```diff
@@ -565,8 +565,19 @@ Saves the result in `q`.
 """
 function mid_point!(M::Manifold, q, p1, p2)
     X = log(M, p1, p2)
-    X /= 2
-    return exp!(M, q, p1, X)
+    return exp!(M, q, p1, X / 2)
+end
+
+@static if VERSION <= v"1.1"
```
kellertuer, line 571: Why does this not work? It looks like exp/log should work anyways and despite that we only divide by two? Maybe `fill(X / 2)` could just be `X ./ 2`?

test/runtests.jl
mateuszbaran: Are you sure? Just calculating the number of ambiguities shouldn't fail.

kellertuer: Well, it was with `Test.detect_ambiguities(Base)` which ran into a Stack Overflow, so in `Manifolds` also that call is within the `if`.

kellertuer: But it's also not that bad, so do what you feel is better; I'll already accept this :)

mateuszbaran: OK, I've put it in that `if`.

kellertuer:
Oh and could you write a short test for the new case?

mateuszbaran:
I've added a test but these lines won't be detected as covered by codecov since coverage data is generated on a different version of Julia.

kellertuer:
You're right, well, still good to have the test.

PR #48, opened 2020-09-11
Fixing embedded manifolds

mateuszbaran:
This is a bit more involved than I originally thought it would be. It should fix the issue noticed here: https://github.com/JuliaManifolds/Manifolds.jl/pull/216#discussion_r487152910 . There are also a few related fixes. Let me know if this makes sense.

### DIFF
--- a/src/EmbeddedManifold.jl
+++ b/src/EmbeddedManifold.jl
@@ -89,6 +89,16 @@ function EmbeddedManifold(
     return EmbeddedManifold{𝔽,MT,NT,ET}(M, N)
 end
 
+function allocate_result(M::AbstractEmbeddedManifold, f::typeof(embed), x...)
+    T = allocate_result_type(M, f, x)
+    return allocate(x[end], T, representation_size(decorated_manifold(M)))
+end
+
+function allocate_result(M::AbstractEmbeddedManifold, f::typeof(project), x...)
+    T = allocate_result_type(M, f, x)
+    return allocate(x[end], T, representation_size(base_manifold(M)))
+end
+
 """
     base_manifold(M::AbstractEmbeddedManifold, d::Val{N} = Val(-1))
 
@@ -112,10 +122,10 @@ base_manifold(M::EmbeddedManifold, d::Val{N} = Val(-1)) where {N} = M.manifold
     check_manifold_point(M::AbstractEmbeddedManifold, p; kwargs)
 
 check whether a point `p` is a valid point on the [`AbstractEmbeddedManifold`](@ref),
-i.e. that `embed(M, p)` is a valid point on the embedded manifold.
+i.e. that `project(M, p)` is a valid point on the embedded manifold.
 """
 function check_manifold_point(M::AbstractEmbeddedManifold, p; kwargs...)
-    q = embed(M, p)
+    q = project(M, p)
     return invoke(
         check_manifold_point,
         Tuple{typeof(get_embedding(M)),typeof(q)},
@@ -128,7 +138,7 @@ end
 """
     check_tangent_vector(M::AbstractEmbeddedManifold, p, X; check_base_point = true, kwargs...)
 
-check that `embed(M,p,X)` is a valid tangent to `embed(p,X)`, where `check_base_point`
+check that `project(M, p, X)` is a valid tangent to `project(M, p)`, where `check_base_point`
 determines whether the validity of `p` is checked, too.
 """
 function check_tangent_vector(
@@ -142,8 +152,8 @@ function check_tangent_vector(
         mpe = check_manifold_point(M, p; kwargs...)
         mpe === nothing || return mpe
     end
-    q = embed(M, p)
-    Y = embed(M, p, X)
+    q = project(M, p)
+    Y = project(M, p, X)
     return invoke(
         check_tangent_vector,
         Tuple{typeof(get_embedding(M)),typeof(q),typeof(Y)},
--- a/src/EmbeddedManifold.jl
+++ b/src/EmbeddedManifold.jl
@@ -15,6 +15,12 @@ The embedding is further specified by an [`AbstractEmbeddingType`](@ref).
 This means, that technically an embedded manifold is a decorator for the embedding, i.e.
 functions of this type get, in the semi-transparent way of the
 [`AbstractDecoratorManifold`](@ref), passed on to the embedding.
+
+!!! note
+    
+    Points on an `AbstractEmbeddedManifold` are still represented using representation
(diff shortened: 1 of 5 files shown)


src/EmbeddedManifold.jl
```diff
@@ -112,10 +122,10 @@ base_manifold(M::EmbeddedManifold, d::Val{N} = Val(-1)) where {N} = M.manifold
     check_manifold_point(M::AbstractEmbeddedManifold, p; kwargs)
 
 check whether a point `p` is a valid point on the [`AbstractEmbeddedManifold`](@ref),
-i.e. that `embed(M, p)` is a valid point on the embedded manifold.
+i.e. that `project(M, p)` is a valid point on the embedded manifold.
 """
 function check_manifold_point(M::AbstractEmbeddedManifold, p; kwargs...)
-    q = embed(M, p)
+    q = project(M, p)
```
kellertuer, line 128: maybe this should not be necessary at all, since we assume that `M.manifold` itself does not change, but just the type of embed/projct you use?

src/EmbeddedManifold.jl
```diff
@@ -89,6 +89,16 @@ function EmbeddedManifold(
     return EmbeddedManifold{𝔽,MT,NT,ET}(M, N)
 end
 
+function allocate_result(M::AbstractEmbeddedManifold, f::typeof(embed), x...)
+    T = allocate_result_type(M, f, x)
+    return allocate(x[end], T, representation_size(decorated_manifold(M)))
```
kellertuer, line 94: Why the last and not the first?

src/EmbeddedManifold.jl
kellertuer: But well, one actually can check that `embed(...)` yields a valid point on `M.embedding` for sure.

src/EmbeddedManifold.jl
```diff
@@ -112,10 +122,10 @@ base_manifold(M::EmbeddedManifold, d::Val{N} = Val(-1)) where {N} = M.manifold
...
 check whether a point `p` is a valid point on the [`AbstractEmbeddedManifold`](@ref),
-i.e. that `embed(M, p)` is a valid point on the embedded manifold.
+i.e. that `project(M, p)` is a valid point on the embedded manifold.
 """
 function check_manifold_point(M::AbstractEmbeddedManifold, p; kwargs...)
-    q = embed(M, p)
+    q = project(M, p)
     return invoke(
         check_manifold_point,
         Tuple{typeof(get_embedding(M)),typeof(q)},
```
kellertuer, line 131: ...as we do here. So it should be `embed` atop not `project`.

src/EmbeddedManifold.jl
```diff
@@ -142,8 +152,8 @@ function check_tangent_vector(
         mpe = check_manifold_point(M, p; kwargs...)
         mpe === nothing || return mpe
     end
-    q = embed(M, p)
-    Y = embed(M, p, X)
+    q = project(M, p)
+    Y = project(M, p, X)
```
kellertuer, line 156: see atop, this should stay embed.

src/EmbeddedManifold.jl
mateuszbaran: In `x` is either point and we want to return point, or point and vector and we want to return vector, so the result is more similar to the last argument.

src/EmbeddedManifold.jl
mateuszbaran: Let's assume we have R^2 embedded in R^3. `p` is supposed to be a 3-element vector. We want to convert it to the 2-element vector representation expected by the `check_manifold_point` method invoked below. So that looks like what `project` does I think. I only touched this part because otherwise tests were failing and it sort of makes sense.

kellertuer: My interpretation until now was (and that'S also where the Euclidean embeddings come from), that a point `p` on the manifold (R^2 in your example) can have different embeddings we would like to transfer it to (one of them is R^3) but it might be in _any_ representation (depending on which manifold you have).

For example one could think of M being `M=Symmetric(2)` – that can be nicely embedded in N=R3 but points p are 2x2 matrices. We only know
- check_manifold_point(E.Manifold,p) is true
- check_manifold_point(E.embedding, embed(E,p)) is true

Embedded manifold `E(M,N)` acts like `M` but tells us where we embed into if we call embed. (or reversely) where we project from) but it does not change `p`s from `M` itself. It “decorates” `M` with an embedding – but only embeds points when we call `embed`.

src/EmbeddedManifold.jl
kellertuer: Thanks for the explanation – clever!

src/EmbeddedManifold.jl
mateuszbaran: Yes, you are right. Points are by default stored in the `E.manifold` representation and can be converted using `embed` to the `E.embedding` representation. I will fix this.

kellertuer: Still great to discuss, since before I wasn't that sure either, but this way it feels correct.

mateuszbaran:
I think the PR should be OK now :slightly_smiling_face: .

test/embedded_manifold.jl
```diff
@@ -40,11 +62,11 @@ struct NotImplementedEmbeddedManifold3 <: AbstractEmbeddedManifold{ℝ,DefaultEm
         M = PlaneManifold()
         @test repr(M) == "PlaneManifold()"
         @test ManifoldsBase.default_decorator_dispatch(M) === Val{false}()
-        @test get_embedding(M) == ManifoldsBase.DefaultManifold(3)
+        @test get_embedding(M) == ManifoldsBase.DefaultManifold(1, 3)
         # Check fallbacks to check embed->check_manifoldpoint Defaults
-        @test_throws DomainError is_manifold_point(M, [1, 0], true)
-        @test_throws DomainError is_tangent_vector(M, [1, 0], [1, 0, 0], true)
-        @test_throws DomainError is_tangent_vector(M, [1, 0, 0], [0, 0], true)
+        @test is_manifold_point(M, [1, 0], true)
```
kellertuer, line 67: Why are these now working? Especially the tangent vector should not work...

mateuszbaran: That's because embedding itself can "fix" points now -- it doesn't check if the representation before embedding is correct, just copies what is needed. New methods of `allocate_result` ensure that an array of the right size is filled.

kellertuer: Ah, ok.

kellertuer (approved):

PR #49, opened 2020-09-15
More rigorous size tests before allocation

kellertuer:
I was using a “feature”, where on any manifold the allocation was performed using the size of the point instead of the representation size. This small PR makes this more rigorous and introduces a `check_size` for points and vectors, that might even simplify tests for certain manifolds (by a few lines).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "ManifoldsBase"
 uuid = "3362f125-f0bb-47a3-aa74-596ffd7ef2fb"
 authors = ["Seth Axen <seth.axen@gmail.com>", "Mateusz Baran <mateuszbaran89@gmail.com>", "Ronny Bergmann <manopt@ronnybergmann.net>", "Antoine Levitt <antoine.levitt@gmail.com>"]
-version = "0.9.3"
+version = "0.10.0"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "ManifoldsBase"
 uuid = "3362f125-f0bb-47a3-aa74-596ffd7ef2fb"
 authors = ["Seth Axen <seth.axen@gmail.com>", "Mateusz Baran <mateuszbaran89@gmail.com>", "Ronny Bergmann <manopt@ronnybergmann.net>", "Antoine Levitt <antoine.levitt@gmail.com>"]
-version = "0.9.2"
+version = "0.9.3"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "ManifoldsBase"
 uuid = "3362f125-f0bb-47a3-aa74-596ffd7ef2fb"
 authors = ["Seth Axen <seth.axen@gmail.com>", "Mateusz Baran <mateuszbaran89@gmail.com>", "Ronny Bergmann <manopt@ronnybergmann.net>", "Antoine Levitt <antoine.levitt@gmail.com>"]
-version = "0.10.0"
+version = "0.9.3"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
--- a/src/EmbeddedManifold.jl
+++ b/src/EmbeddedManifold.jl
@@ -92,22 +92,12 @@ end
 
 function allocate_result(M::AbstractEmbeddedManifold, f::typeof(embed), x...)
     T = allocate_result_type(M, f, x)
-    return allocate(x[end], T, representation_size(decorated_manifold(M)))
+    return allocate(x[1], T, representation_size(decorated_manifold(M)))
 end
 
 function allocate_result(M::AbstractEmbeddedManifold, f::typeof(project), x...)
     T = allocate_result_type(M, f, x)
-    return allocate(x[end], T, representation_size(base_manifold(M)))
-end
-
-function allocate_result(M::EmbeddedManifold, f::typeof(embed), x...)
-    T = allocate_result_type(M, f, x)
-    return allocate(x[end], T, representation_size(get_embedding(M)))
-end
-
-function allocate_result(M::EmbeddedManifold, f::typeof(project), x...)
-    T = allocate_result_type(M, f, x)
-    return allocate(x[end], T, representation_size(base_manifold(M)))
+    return allocate(x[1], T, representation_size(base_manifold(M)))
 end
 
 """
@@ -207,18 +197,6 @@ get_embedding(::EmbeddedManifold)
 function get_embedding(M::EmbeddedManifold)
     return M.embedding
 end
-
-function project(M::EmbeddedManifold, p)
-    q = allocate_result(M, project, p)
-    project!(M, q, p)
-    return q
-end
-function project(M::EmbeddedManifold, p, X)
-    Y = allocate_result(M, project, p, X)
-    project!(M, Y, p, X)
-    return Y
-end
-
 
 function show(io::IO, M::EmbeddedManifold{𝔽,MT,NT}) where {𝔽,MT<:Manifold{𝔽},NT<:Manifold}
     return print(io, "EmbeddedManifold($(M.manifold), $(M.embedding))")
--- a/src/EmbeddedManifold.jl
+++ b/src/EmbeddedManifold.jl
@@ -97,7 +97,7 @@ end
 
 function allocate_result(M::AbstractEmbeddedManifold, f::typeof(embed), x...)
     T = allocate_result_type(M, f, x)
-    return allocate(x[end], T, representation_size(get_embedding(M)))
+    return allocate(x[end], T, representation_size(decorated_manifold(M)))
 end
 
 function allocate_result(M::AbstractEmbeddedManifold, f::typeof(project), x...)
@@ -123,7 +123,52 @@ type the internally stored enhanced manifold `M.manifold` is returned.
 """
 base_manifold(M::EmbeddedManifold, d::Val{N} = Val(-1)) where {N} = M.manifold
 
-decorated_manifold(M::AbstractEmbeddedManifold) = base_manifold(M)
+
+"""
+    check_manifold_point(M::AbstractEmbeddedManifold, p; kwargs)
+
+check whether a point `p` is a valid point on the [`AbstractEmbeddedManifold`](@ref),
+i.e. that `embed(M, p)` is a valid point on the embedded manifold.
+"""
+function check_manifold_point(M::AbstractEmbeddedManifold, p; kwargs...)
+    return invoke(
+        check_manifold_point,
+        Tuple{typeof(get_embedding(M)),typeof(p)},
+        get_embedding(M),
+        p;
+        kwargs...,
+    )
+end
+
+"""
+    check_tangent_vector(M::AbstractEmbeddedManifold, p, X; check_base_point = true, kwargs...)
+
+check that `embed(M, p, X)` is a valid tangent to `embed(M, p)`, where `check_base_point`
+determines whether the validity of `p` is checked, too.
+"""
+function check_tangent_vector(
+    M::AbstractEmbeddedManifold,
+    p,
+    X;
+    check_base_point = true,
+    kwargs...,
+)
+    if check_base_point
+        mpe = check_manifold_point(M, p; kwargs...)
+        mpe === nothing || return mpe
+    end
+    return invoke(
+        check_tangent_vector,
+        Tuple{typeof(get_embedding(M)),typeof(p),typeof(X)},
+        get_embedding(M),
+        p,
+        X;
+        check_base_point = check_base_point,
+        kwargs...,
+    )
+end
+
+decorated_manifold(M::AbstractEmbeddedManifold) = M.embedding
 
 """
     get_embedding(M::AbstractEmbeddedManifold)
@@ -133,7 +178,7 @@ Return the [`Manifold`](@ref) `N` an [`AbstractEmbeddedManifold`](@ref) is embed
 get_embedding(::AbstractEmbeddedManifold)
 
 @decorator_transparent_function function get_embedding(M::AbstractEmbeddedManifold)
-    return M.embedding
+    return decorated_manifold(M)
 end
 
 function show(
@@ -163,17 +208,15 @@ function decorator_transparent_dispatch(
     ::AbstractEmbeddedManifold,
     args...,
 )
-    return Val(:transparent)
+    return Val(:intransparent)
 end
-
 function decorator_transparent_dispatch(
     ::typeof(check_tangent_vector),
     ::AbstractEmbeddedManifold,
     args...,
 )
-    return Val(:transparent)
+    return Val(:intransparent)
 end
-
 function decorator_transparent_dispatch(
     ::typeof(embed),
     ::AbstractEmbeddedManifold,
--- a/src/EmbeddedManifold.jl
+++ b/src/EmbeddedManifold.jl
@@ -17,7 +17,7 @@ functions of this type get, in the semi-transparent way of the
 [`AbstractDecoratorManifold`](@ref), passed on to the embedding.
 
 !!! note
-    
+
     Points on an `AbstractEmbeddedManifold` are still represented using representation
     of the embedded manifold. Use [`embed`](@ref) to go to the representation of the embedding
     and [`project`](@ref) to go the other way.
@@ -131,6 +131,8 @@ check whether a point `p` is a valid point on the [`AbstractEmbeddedManifold`](@
 i.e. that `embed(M, p)` is a valid point on the embedded manifold.
 """
 function check_manifold_point(M::AbstractEmbeddedManifold, p; kwargs...)
+    mse = check_size(M,p)
+    mse === nothing || return mse
     q = embed(M, p)
     return invoke(
         check_manifold_point,
@@ -159,6 +161,8 @@ function check_tangent_vector(
         mpe === nothing || return mpe
     end
     q = embed(M, p)
+    mse = check_size(M,p,X)
+    mse === nothing || return mse
     Y = embed(M, p, X)
     return invoke(
         check_tangent_vector,
--- a/src/EmbeddedManifold.jl
+++ b/src/EmbeddedManifold.jl
@@ -238,6 +238,13 @@ function decorator_transparent_dispatch(
     return Val(:parent)
 end
 function decorator_transparent_dispatch(
+    ::typeof(embed),
+    ::EmbeddedManifold,
+    args...,
+)
+    return Val(:parent)
+end
+function decorator_transparent_dispatch(
     ::typeof(embed!),
     ::AbstractEmbeddedManifold,
     args...,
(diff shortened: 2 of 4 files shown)


kellertuer:
Hehe. This one breaks your (@mateuszbaran ) understanding of how all this should work. So we should agree on a common base and implement that here.

I expect now that a point that comes into `embed` (similarly a tangent) is of representation size of the manifold.
You are also ok with sizes of the embedding, right?

mateuszbaran:
Thanks, this is a good solution for the problem you've discovered in https://github.com/JuliaManifolds/Manifolds.jl/pull/216 . And actually this works how I would expect it to work after our last discussion of embedded manifolds. I have just forgotten about this case.

src/ManifoldsBase.jl
```diff
@@ -238,6 +238,46 @@ type.
...
+function check_size(M::Manifold, p)
+    n = size(p)
+    m = representation_size(M)
+    if length(n) != length(m)
+        return DomainError(
+            n,
+            "The point $(p) can not belong to the manifold $(M), since its size $(n) does not fit the required representation size ($(m)).",
+        )
+    end
+    if any(n != m)
```
mateuszbaran, line 261:
```suggestion
    if n != m
```

src/ManifoldsBase.jl
```diff
@@ -238,6 +238,46 @@ type.
...
+    mse === nothing || return mse
+    n = size(X)
+    m = representation_size(M)
+    if length(n) != length(m)
+        return DomainError(
+            n,
+            "The tangent vector $(X) can not belong to the manifold $(M), since its size $(n) does not fit the required representation size ($(m)).",
+        )
+    end
+    if any(n != m)
```
mateuszbaran, line 276:
```suggestion
    if n != m
```

src/ManifoldsBase.jl
```diff
@@ -238,6 +238,46 @@ type.
...
+By default, `check_size` returns `nothing`, i.e. if no checks are implemented, the
+assumption is to be optimistic.
+"""
+function check_size(M::Manifold, p)
+    n = size(p)
+    m = representation_size(M)
+    if length(n) != length(m)
+        return DomainError(
+            n,
+            "The point $(p) can not belong to the manifold $(M), since its size $(n) does not fit the required representation size ($(m)).",
```
mateuszbaran, line 258: Maybe "is not equal to" would be more precise than "does not fit"?

mateuszbaran:
Wait, no, sorry, I've read that again more carefully and it's the other way in this PR. The worst thing is that both my and your PR sort of make sense and sort of not exactly.

Anyway, why does `check_manifold_point` actually need to call `embed`? Shouldn't `check_manifold_point` just have `:transparent` forwarding?

kellertuer:
Ok, let me try to explain, what I think, the idea is, we currently have implemented. We have two cases

1) the `EmbeddedManifold(M,N)` of a manifold `M` and its embedding `N`. This acts as a decorator similar to `MetricManifold`, i.e. in my mind it does not change how `p` or `X` are stored. It just provides an indicator which embedding to use. There might be a default one implemented (the `Sphere(n)` in `Euclidean(n+1)` for example), then we do not use this case necessarily, but for a second embedding we might specify it using this case.
Currently this is also, what we do, `p` and `X` do not change but you can implement different `embed!`/`project!` methods dispatching on this type

2) The `AbstractEmbeddingManifold` part. In my understanding, this is used to fix the default embedding (`MyManifold <: AbstractEmbeddedManifold`) in order to reuse code. Still `p` and `X` have a representation independent of the `AbstractEmbeddedManifold` we choose, but usually the `embed!`/`project!` should be cheap here.
For the Sphere `embed` is the identity, also for the `SymmetricMatrices` or the `CenteredMatrices`, since the representation we use _is_ actually the one in the embedding.
The case you mentioned is exactly this case, where we check validity of a point by first checking that its correct when embedded.
This can not be `:transparent`. For example for the sphere what we actually do is, check, that the vector we have is a valid `Euclidean(n+1)` vector (for example not a metric and not complex and such) and then check that it'S norm 1. Since the first part is already implemented in `Euclidean` we use the `AbstractEMbedding`s `embed` (identity here) to check the “real Euclideanness” of `p` using that implementation (instead of implementing it _again_ in `Sphere). And then only check the norm constraint afterwards.
And for this second case, the generic implementation ( https://github.com/JuliaManifolds/ManifoldsBase.jl/blob/8789cab9a590704cf9e151a08f45160be5b71630/src/EmbeddedManifold.jl#L133 ) just checks this check in the embedding.

I hope that makes clear how it currently is (also in this PR), how I thought it should be and how I used it – especially the second case – in quite some manifolds.

Looking forward to hearing your ideas.

mateuszbaran:
> 1. the `EmbeddedManifold(M,N)` of a manifold `M` and its embedding `N`. This acts as a decorator similar to `MetricManifold`, i.e. in my mind it does not change how `p` or `X` are stored. It just provides an indicator which embedding to use. There might be a default one implemented (the `Sphere(n)` in `Euclidean(n+1)` for example), then we do not use this case necessarily, but for a second embedding we might specify it using this case.
>    Currently this is also, what we do, `p` and `X` do not change but you can implement different `embed!`/`project!` methods dispatching on this type

Currently `EmbeddedManifold` decorates the embedding `N`, since that's what `decorated_manifold` returns. But `base_manifold` returns `M`. Documentation seems to get it wrong here: https://juliamanifolds.github.io/Manifolds.jl/stable/interface.html#Technical-Details . Also this documentation is different from what the code does: https://juliamanifolds.github.io/Manifolds.jl/stable/interface.html#ManifoldsBase.base_manifold-Union{Tuple{EmbeddedManifold},%20Tuple{N},%20Tuple{EmbeddedManifold,Val{N}}}%20where%20N .

Before I'll continue, which manifold do you think should be considered as decorated by `EmbeddedManifold`?

kellertuer:
Technically it might decorate `N`, philosophically it decorates `M`, I think. We discussed that a while ago I think

mateuszbaran:
There is a very lengthy discussion about it in https://github.com/JuliaManifolds/Manifolds.jl/pull/114 but it's hard to find anything there and apparently the result is still inconsistent.

Given how `AbstractEmbeddedManifold` is used, it usually deals with only one representation so `embed!` should be just a `copyto!` and `project!` won't change the representation. `EmbeddedManifold` is right now the only exception since it is used for dealing with two different representations. We agreed that the representation of the embedded manifold `M` is used by default, so why would it technically decorate the embedding `N` and not `M`?

kellertuer:
If we say that for an `AbstractEmbeddedManifold` always means, that it is implicitly embedded, we also do not need any copyto, we can just leave the embed out. I would also be fine with that; it would just restrict the cases we can do `AbstractEmbeddedManifolds`, but all cases that we have would still be covered; and I would be fine with that.

mateuszbaran:
I think I don't follow you now, why would we leave embed out? And what restrictions do you mean?

kellertuer:
Well, let's be more precise:
I would like to leave the `EmbeddedManifold`as is.
For the `AbstractEmbeddedManifold` one could maybe better interpret that as a manifold that is represented by certain points in the embedding (like the sphere). Then the `embed`line ( https://github.com/JuliaManifolds/ManifoldsBase.jl/blob/8789cab9a590704cf9e151a08f45160be5b71630/src/EmbeddedManifold.jl#L134 ) is not necessary.

For the decoration, I always think it should be `M` that is decorated – but technically since we want to pass on things _to_ the embedding (to not reimplement them) it is currently `N`.

mateuszbaran:
OK, so let's start with leaving `EmbeddedManifold` mostly as is. Since the representation of the embedded manifold is used, I would say `decorated_manifold` should return the embedded manifold. I just checked and it only breaks a single test in ManifoldsBase, one that specifically checks for that. Technically, forwarding anything to the embedding requires a call to `embed` then so I don't really see any convincing (technical or not) argument for considering `EmbeddedManifold` a decorator of the embedding.

Secondly, I don't really see the logic behind our current variant of `check_manifold_point` for `AbstractEmbeddedManifold`. For `EmbeddedManifold` it should just forward to the embedded manifold, and for other types it should have a custom implementation anyway.

For other `AbstractEmbeddedManifold`, I think we are only interested in cases where `embed` does nothing? I would like to find some consistency between `AbstractEmbeddedManifold` and `EmbeddedManifold`.

kellertuer:
You'Re right the AbstractEmbeddedManifold could just forward to the embedding; sphere currentlich passes on to this method (and checks norm afterwards). Thats what I meant with removing the embed method there. And yes the cases of AbstractManifold are all Manifolds represented by points in the embedding already, hence there embed does nothing.

kellertuer:
I am slightly confused, that Format fails, since locally running the same command works?
Despite that I changed what we discussed and added tests for the `check_size` function, which I might use in Manifolds to reduce code.

mateuszbaran:
The formatter from github action uses the 0.8 release and you probably are on 0.7.something. Anyway, there seems to be something broken about changes suggested by that 0.8 version.

I'll check your changes in a moment.

mateuszbaran:
I'm preparing a proposal for a few changes.

mateuszbaran:
One question: do we assume that for `AbstractEmbeddedManifold`s other than `EmbeddedManifold` `embed!` is always just `copyto!`? I.e. there is no need for embedding? You've started to go in that direction but our test `PlaneManifold` violates that in a very bad way (`embed!` is implemented as a `copyto!` but representation sizes differ between manifold and its embedding).

kellertuer:
No. I don'T want to exclude manifolds, where it's not just copy. Imagine the Sphere in angles (Lon/Lat – yes, horrible to compute in but) then embedding in R3 is not just copy.
I hope I only went into that direction for the AbstractEmbeddedManifold but not for the embedded one, I thought that was the conclusion here, am I wrong?

kellertuer:
> The formatter from github action uses the 0.8 release and you probably are on 0.7.something. Anyway, there seems to be something broken about changes suggested by that 0.8 version.
> 
> I'll check your changes in a moment.

ah, i see, somehow Formatter seems broken, it `0.8` always reports false on my machine and introduces a newline at several places (between `abstract` and `type` on every run. Still trying to do an mwe.

kellertuer:
Concerning Formatter, I tried narrowing down, where it fails – looks interesting, see https://github.com/domluna/JuliaFormatter.jl/issues/291

mateuszbaran:
> No. I don'T want to exclude manifolds, where it's not just copy. Imagine the Sphere in angles (Lon/Lat – yes, horrible to compute in but) then embedding in R3 is not just copy.
> I hope I only went into that direction for the AbstractEmbeddedManifold but not for the embedded one, I thought that was the conclusion here, am I wrong?

It wasn't well articulated before I think. In any case I would also prefer to not restrict embedding for `AbstractEmbeddedManifold`. The only issue is that quite a few things need to be cleaned up. I'm figuring out what exactly.

> Concerning Formatter, I tried narrowing down, where it fails – looks interesting, see [domluna/JuliaFormatter.jl#291](https://github.com/domluna/JuliaFormatter.jl/issues/291)

Thanks for looking into it!

kellertuer:
> The only issue is that quite a few things need to be cleaned up. I'm figuring out what exactly.

Looking forward to your suggestion, since with the status as it is now, I would be quite ok.

src/EmbeddedManifold.jl
```diff
@@ -97,7 +97,7 @@ end
 
 function allocate_result(M::AbstractEmbeddedManifold, f::typeof(embed), x...)
     T = allocate_result_type(M, f, x)
-    return allocate(x[end], T, representation_size(decorated_manifold(M)))
```
kellertuer, line 100: This is for sure the syntactically correct way (though they might do the same as it is currently implemented

mateuszbaran:
I've pushed some changes. `EmbeddedManifold` and `AbstractEmbeddedManifold` both use representation of the embedded manifold and decorate it. Checks are now transparent because of 1. For things like Sphere it makes no difference since both representations are the same.

mateuszbaran:
One thing that will have to be changed if this is accepted is forwarding to the embedding making use of `embed`.

kellertuer (commented):
> I've pushed some changes. `EmbeddedManifold` and `AbstractEmbeddedManifold` both use representation of the embedded manifold
... by default. That's breaking but I might be fine with that.
If a manifold does use another representation but wants to be an embedded manifold, it hast to implement `embed!` anyways.

The most breaking one might be that you changed the decorated to the `get_embedding` – this has to be adapted in a few manifolds – but sure is then better readable.

kellertuer:
Just a first small remark from testing with `CenteredMatrices`

Can we introduce the following default?
````julia
get_embedding(M::AbstractEmbeddedManifold) = get_decorated_manifold(M)
````
since that's what the abstract manifolds actually model? Then this would not break any test (and maybe be even a non-breaking PR?).

File: src/EmbeddedManifold.jl [hunk 5]
```diff
@@ -181,7 +133,7 @@ Return the [`Manifold`](@ref) `N` an [`AbstractEmbeddedManifold`](@ref) is embed
 get_embedding(::AbstractEmbeddedManifold)
 
 @decorator_transparent_function function get_embedding(M::AbstractEmbeddedManifold)
-    return decorated_manifold(M)
+    return M.embedding
```
kellertuer, line 136: Usually the abstract embedded manifolds do not store this, since the `decorated_manifold` is the one that _is_ the embedding.

kellertuer: This is also the only change that breaks things and would be much work to all current `AbstractEmbeddedManifold`s. None of them has this field currently (since its just the parent we have to get).

mateuszbaran: This is the key change of my commit, `AbstractEmbeddedManifold` now decorated the embedded manifold. I believe this makes embedded manifolds more coherent. I probably should restrict this method to `EmbeddedManifold` and then go and fix Manifolds.jl.

With this change, `AbstraceEmbeddedManifold`s are much easier to understand: they just decorate in a sense the embedded manifold, providing `embed` to go to the embedding and `project` to go back.

src/EmbeddedManifold.jl
mateuszbaran: Actually after that commit they don't do the same thing ;).

src/EmbeddedManifold.jl
kellertuer: But you then store the embedding twice?
Once `M <: N` (i.e. it acts transparently for some parts to its embedding `N`) _and_ you store the same manifold in `M.embedding`? The only thing that might cause is inconsistencies when those two are different?

Then it really decorates the embedded manifold, then `decorated_manifold` should return the manifold that is decorated (as it does for other decorators, on the syntax level) and here the semantic is that this decorated manifold is the embedding. That was the original modelling and is the one I favour.

mateuszbaran: Before this change `AbstractEmbeddedManifolds` didn't really work for cases where `embed!` wasn't a no-op so some amount of fixes is necessary anyway.

kellertuer: I don'T see why it should not work?

Especially I would prefer not to store the embedding twice (and introduce a lot of `.embedding` fields).

mateuszbaran: `EmbeddedManifold` represents points using the representation of embedded manifold, so naturally all `AbstractEmbeddedManifold`s should do the same, so to use methods of the embedding calling `embed!` is actually necessary. Unless we go back to representing points using the representation of the embedding.

kellertuer: Now you confuse me. You are writing
1) `AbstractEmbeddedManifolds` use representation of the embdedding but
2) to use methods in the embedding one still has to call `embed!`? That contradicts 1). and so does the next sentence.

I think for the `EmbeddedManifold` your sentence is wrong. The concrete case first just “couples” two manifolds in order to define how to embed one in the other. That's why you have both fields.

For the abstract case: Yes. The manifold uses a representation in the embedding. But this is given by transparency and parent and not by a field `.embedding`.

mateuszbaran: > I don'T see why it should not work?
> 
> Especially I would prefer not to store the embedding twice (and introduce a lot of `.embedding` fields).

There won't be any additional `.embedding` fields :).

mateuszbaran:
As discussed in the video call, I've made `EmbeddedManifold` not an `AbstractEmbeddedManifold`.

kellertuer:
And that indeed really fits better; there was never any transparency, so the embedding type is unnecessary for this type.

mateuszbaran:
I ran most of the tests in Manifolds.jl using this branch and there were no failures. That's nice 🙂 . I think the main thing that needs to be still done here is checking documentation but I need to take a break from this stuff.

kellertuer:
For sure. We are not in a hurry. Thanks for all your work.

I think it breaks the current (new) euclidean embedded manifolds (they now have a parametric type to much), but the rest could be fine.

mateuszbaran:
I've pushed a few fixes and the code seems fine now :slightly_smiling_face: . I will check your Euclidean embedded.

kellertuer:
I like how well it turns out now. The only thing that breaks on `Manifolds` are now the embedded manifolds for Euclidean fail, since I specified the embeddingtype, which one just has to delete now (makes that even nicer, since the embedding type is also nowhere used).

mateuszbaran:
Yes, we managed to work out a nice solution :slightly_smiling_face: .

kellertuer:
Yes I think despite a little bit of test coverage, we are good to go. Thanks for the help.

mateuszbaran:
Could you check your embedded Euclidean PR against this branch? There is something weird with dispatch. I probably marked embed and project as `:parent` incorrectly or something.

kellertuer:
> Could you check your embedded Euclidean PR against this branch? There is something weird with dispatch. I probably marked embed and project as `:parent` incorrectly or something.

I take a look, it seems the allocation gets something wrong somewhere (in size).

mateuszbaran:
> > 
> 
> I take a look, it seems the allocation gets something wrong somewhere (in size).

Yes, that's because it thinks that `embed` has `transparent` dispatch instead of `parent` for `EmbeddedManifold` for some reason.

kellertuer:
It was a little tricky, but I carefully narrowed it down. With the change, that the `EmbeddedManifold` now acts very similar to `MetricManifold`, namely it provides a decorator to `M` by specifying a certain embedding `N` (not that allows for a default embedding as it does for the metric), there was a problem for dispatching embedding`(and similarly `project`,
* its not transparent (that would mean we use the default embedding, and that is exactly what we _don't_ want to do here
* its not parent
* its not intransparent, since we exactly now what to do (we also already have the allocates.

So for now implementing them solves the problem, one could incoke the ones of `Manifold`, too.
Especially one could see that in the test coverage, the allocation for embed or project (for the Embeddedmanifold) was never called).

With that I also noticed a wrong addressing in the Euclidean embed, but fixed that, too.

kellertuer:
Maybe parent actually does work, it sees I could remove the special implementations indeed. I hope now this is finished. Let's wait for coverage.

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "ManifoldsBase"
 uuid = "3362f125-f0bb-47a3-aa74-596ffd7ef2fb"
 authors = ["Seth Axen <seth.axen@gmail.com>", "Mateusz Baran <mateuszbaran89@gmail.com>", "Ronny Bergmann <manopt@ronnybergmann.net>", "Antoine Levitt <antoine.levitt@gmail.com>"]
-version = "0.9.2"
+version = "0.9.3"
```
mateuszbaran, line 4: I think this is breaking enough to call it 0.10.0.

Other than that, I think this is good to go as soon as you are happy with how it works with https://github.com/JuliaManifolds/Manifolds.jl/pull/216

mateuszbaran (approved):

Project.toml
kellertuer: It basically only breaks said PR - but to be honest, that's the first time we actually use the embedded manifold. I checked with 216 and I am fine with that. So we can surely merge this.

I don'T have a strong preference whether to call it `0.9.3` or `0.10.0` so I leave the decision up to you.

mateuszbaran: I think we can also call it a technically breaking bugfix. I'll leave 0.9.3 then.

PR #51, opened 2020-09-22
Simplify abstract embedded

kellertuer:
I had a minute to spare and this already works with the manifolds actually.

### DIFF
--- a/src/EmbeddedManifold.jl
+++ b/src/EmbeddedManifold.jl
@@ -182,8 +182,8 @@ function embed(M::EmbeddedManifold, p)
     return q
 end
 
-embed!(::AbstractEmbeddedManifold, q, p) = copyto!(q,p)
-embed!(::AbstractEmbeddedManifold, Y, p, X) = copyto!(Y,X)
+embed!(::AbstractEmbeddedManifold, q, p) = copyto!(q, p)
+embed!(::AbstractEmbeddedManifold, Y, p, X) = copyto!(Y, X)
 
 """
     get_embedding(M::AbstractEmbeddedManifold)
--- a/src/EmbeddedManifold.jl
+++ b/src/EmbeddedManifold.jl
@@ -70,8 +70,8 @@ struct TransparentIsometricEmbedding <: AbstractIsometricEmbeddingType end
 """
     EmbeddedManifold{𝔽, MT <: Manifold, NT <: Manifold} <: AbstractDecoratorManifold{𝔽}
 
-A type to represent that a [`Manifold`](@ref) `M` of type `MT` can be embedded into a
-manifold `N` of type `NT`.
+A type to represent an explicit embedding of a [`Manifold`](@ref) `M` of type `MT` embedded
+into a manifold `N` of type `NT`.
 
 !!!note
     for a first (default) embedding, this type is not required, one can just implement
--- a/src/EmbeddedManifold.jl
+++ b/src/EmbeddedManifold.jl
@@ -182,6 +182,8 @@ function embed(M::EmbeddedManifold, p)
     return q
 end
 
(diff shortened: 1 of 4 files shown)


mateuszbaran:
I thought we were considering removing `embed!` for `AbstractEmbeddedManifold` entirely?

kellertuer:
What do you further want to move? I thought those two lines it all it does?

mateuszbaran:
Maybe let's discuss it on Slack :slightly_smiling_face: .

mateuszbaran:
Looks nice, I just think now the docstring for `embed` should be corrected to say that:
  1. It is expected to be explicitly defined for `EmbeddedManifold`.
  2. It is defined as identity for `AbstractEmbeddedManifold`s.
  3. It may or may not be defined for other manifolds.

src/EmbeddedManifold.jl
```diff
@@ -66,12 +70,14 @@ struct TransparentIsometricEmbedding <: AbstractIsometricEmbeddingType end
 """
     EmbeddedManifold{𝔽, MT <: Manifold, NT <: Manifold} <: AbstractDecoratorManifold{𝔽}
 
-A type to represent that a [`Manifold`](@ref) `M` of type `MT` is indeed an emebedded
-manifold and embedded into the manifold `N` of type `NT`.
-Points and tangent vectors are still represented in the representation of `M` and the
-manifold `M` is being decorated (contrary to [`AbstractEmbeddedManifold`](@ref)).
-To go to the representation in the embedding, the function [`embed`](@ref) should be used.
-[`project`](@ref) can be used to go the other way.
+A type to represent that a [`Manifold`](@ref) `M` of type `MT` can be embedded into a
```
mateuszbaran, line 73: Maybe you could add that this type represents an explicit embedding of `M` into `N`? The phrase "can be embedded" is a bit vague.

src/EmbeddedManifold.jl
```diff
@@ -66,12 +70,14 @@ struct TransparentIsometricEmbedding <: AbstractIsometricEmbeddingType end
...
+manifold `N` of type `NT`.
+
+!!!note
+    for a first (default) embedding, this type is not required, one can just implement
```
mateuszbaran, line 77:
```suggestion
    This type is not required if a manifold `M` is to be embedded in one specific manifold `N`. One can then just implement
```

src/EmbeddedManifold.jl
```diff
@@ -66,12 +70,14 @@ struct TransparentIsometricEmbedding <: AbstractIsometricEmbeddingType end
...
+    [`embed!`](@ref) and [`project!`](@ref), only for a second –maybe considered non-default–
```
mateuszbaran, line 78:
```suggestion
    [`embed!`](@ref) and [`project!`](@ref). Only for a second –maybe considered non-default–
```

src/EmbeddedManifold.jl
```diff
@@ -66,12 +70,14 @@ struct TransparentIsometricEmbedding <: AbstractIsometricEmbeddingType end
...
+    embedding, this type should be considered in order to dispatch on different embed
+    and project methods
```
mateuszbaran, line 80:
```suggestion
    and project methods for different embeddings `N`.
```

src/ManifoldsBase.jl
```diff
@@ -312,13 +312,17 @@ end
 """
     embed!(M::Manifold, q, p)
 
-Embed point `p` from the [`Manifold`](@ref) `M` into the ambient space and return the result in `q`.
+Embed point `p` from the [`Manifold`](@ref) `M` intointo an ambient space.
```
mateuszbaran, line 315:
```suggestion
Embed point `p` from the [`Manifold`](@ref) `M` into an ambient space.
```

mateuszbaran (approved):

PR #52, opened 2020-10-17
Introduce generic Gram Schmidt

kellertuer:
This method extracts the Gram Schmidt method already present within the projected basis approaches to an independent function working on bases or sets of vectors.


src/bases.jl
```diff
@@ -518,6 +506,94 @@ function _get_vectors(B::CachedBasis{𝔽,<:AbstractBasis,<:DiagonalizingBasisDa
...
+        nrmΞₙ = norm(M, p, Ξₙ)
+        if nrmΞₙ == 0
+            warn_linearly_dependent && @warn "Input vector $(n) has length 0."
+            @goto skip
+        end
+        Ξₙ ./= nrmΞₙ
+        for k in 1:length(Ξ)
+            if !isapprox(real(inner(M, p, Ξ[k], Ξₙ)), 0; kwargs...)
+                warn_linearly_dependent &&
+                    @warn "Input vector $(n) is not linearly independent of output basis vector $(k)."
```
kellertuer, line 582: @mateuszbaran I have a question on this error message. If the nth vector is in the span of the previous vectors, it should be zero after lines 570-572, so the 574 should be active.

So when actually would this method kick in? I haven't completely figured this out but the error message seems to vague, since we also do not compare the actual input vector but its “rework” (remainder after 571&normalized).

mateuszbaran: IIRC this check is done like that to deal in some way with numerical error. I don't know if this is the best way to do that, likely not, but I think this makes sense. 574 only checks for exactly zero while 580 checks for near-zero inner product with one of the previous basis vectors. The check in 574 is I think just to make sure that division in 578 doesn't explode.

The error message indeed doesn't explain how the linear independence is tested, this could be improved.

kellertuer: That sounds reasonable. Yes then mainly the error message should be improved. I hope my doc string covers all special cases you had in mind.

mateuszbaran: Yes, the docstring looks more or less complete. I'll make a more detailed review later.

sethaxen:
It looks like in addition to a refactor, this relaxes the constraint that the number field of the coordinates be real (might have that backwards). Is that intentional and well-tested?

kellertuer:
I don't follow you exactly here? Can you elaborate?
I just copied the code from the previous function that just worked on the ProjectedBasis (see the old line 279) and made it a seperate function, that can be called on arbitrary sets of vectors or bases (not just the set of vectors created in said prior function).
So I don't see where we impose the coordinates to be real, but if we do, we did that previously already, too.

sethaxen:
I'll to reply more later, but here's some of our earlier discussion: https://github.com/JuliaManifolds/Manifolds.jl/pull/86#issuecomment-570857661.
Note that `get_basis` for `ProjectedOrthonormalBasis{:gram_schmidt,ℝ}` is limited to `ℝ`, which requires that the coefficients be real, since the implementation of `gram_schmidt` we use calls `real(inner(...))`. For e.g. a complex manifold, one might want complex coefficients, in which case you can use a smaller set of basis vectors, and then I think the only difference is you wouldn't call `real(...)`. What is the kwarg `field` meant to represent?

kellertuer:
Ah, then the basis should be real, yes. The field just stems from the case that

1) the original method used `number_field(B)`, but if we only have vectors, I wasn't sure where to get the basis type from, so I took the number field of M. Now since the method is called from basis and vectors types, I needed that kwarg.

kellertuer:
I checked your question again Seth and youRe right, since we are only working with real coefficients, there is no need for field and the field of the basis and the manifold have to match (i.e. we can't do a real basis for a complex tangent space because we do not compute in complex coefficients).
I fixed this and removed the field-field.

mateuszbaran:
Looks OK I think. SVD-based orthonormalization could also be separated out and maybe we could have one generic function for orthonormalization with an argument that tells which method should be used.

kellertuer:
Sure, maybe in a next PR?

mateuszbaran:
Yes, sure. Then I'd just prefer to not export the `gram_schmidt` function in this one.

kellertuer:
Sure, I can still use it directly in Manifolds though, and we then only export a common `orthonormalize` function?

mateuszbaran:
Yes, of course.

kellertuer:
Can we merge this? I can check for the svd base approach in a PR maybe next week.

mateuszbaran:
Yes, I'd just suggest removing the export of `gram_schmidt`.

kellertuer:
Oh, sure, sorry, I actually just forgot to do that (and not only write that I'll do that).

mateuszbaran (approved):

Issue #54, opened 2020-11-26
AbstractPowerManifold as an interface?


kellertuer:
I am currently thinking of simplifying the asymptote export interface over in Manopt.jl. For that I would need the
abstract power manifold

https://github.com/JuliaManifolds/Manifolds.jl/blob/9e43ea07ea7c8e69106136a70efc0147f19ba6cf/src/manifolds/PowerManifold.jl#L31

here, to (somewhen) get rid of the dependency of Manifolds in Manopt (i.e. only depend on base). Could we maybe move that over? Just the abstract one and maybe all abstract functions that do not introduce dependencies here.

kellertuer:
I just noticed that for some implementations I need to actually to a `PowerManifold` of some size (currently), but as far as I see that is mainly to construct a zero tangent vector then. So that can surely be refactored to not be based on that.

mateuszbaran:
How do you imagine to use this interface without `Manifolds.jl`? We can move some parts of `AbstractPowerManifold` to `ManifoldsBase.jl` but I'm not sure what could actually be done with the parts that don't introduce new dependencies.

kellertuer:
To be precise I think I would need

* first and foremost a type to dispatch on (would make my asy_exports much nicer, to do the image exports with power and the others without)
* for some algorithms `p[M, I]` is what I need
* if we want to even transfer all passing-elementwise-on functions (like `exp`/`log`/`zero_tangent_vector`) or maybe even `vector transport`(though then also` PowerVectorTransport`has to be moved) 

... but for this last point if that introduces too much dependencies, the first two (or even just the very first) would be fine, too.

mateuszbaran:
Would moving just the nested representation be enough? I think it could be moved here without introducing new dependencies.

kellertuer:
Sure, that would split both approaches we have, but the interface would provide the straight-forward/simple one of nestedness, which would elaborate on why Power is parametrised as it is. That fits very good I think.

mateuszbaran:
I've copied more or less what could go to ManifoldsBase.jl. There are no tests yet, feel free to add some to the ManifoldsBase.jl PR.

PR #55, opened 2020-11-26
Moving power manifold to base

mateuszbaran:
Will solve https://github.com/JuliaManifolds/ManifoldsBase.jl/issues/54 . Has a counterpart in Manifolds.jl: https://github.com/JuliaManifolds/Manifolds.jl/pull/274 .


kellertuer:
Thanks, I will start this soon, great that nearly all of the methods seem to be able to move. The fist test indicates you might have accidentially moved a flat function?

mateuszbaran:
Yes, I've moved a bit too many. I've moved some back in the second commit.

kellertuer:
Thanks for this great start, currently this introduces 2 ambiguities in `get_vector`s and `get_coordinates`s `allocate_result` with respect to the `CachedBasis`. I tried to fix that, but was not yet able to.

kellertuer:
It's still some work to do, and we have to adapt the docs a little bit, since I also moved Error types and a `size_to_tuple` internal function; but then the `Interface.html` will just also get an internal functions part, once this is finished and we finish the corresponding branch over on Manifolds.

All in all I feel like this is a very good idea to introduce this here.

mateuszbaran:
Hmm.. I think the remaining coverage failures are false positives.

kellertuer:
Thanks for the help. I will check tomorrow whether we can get the codecovs to succeed by finding one line, that's all what might be missing to stay in the same range as before.

kellertuer:
I checked and you are right, the remaining ones seem to be false misses.

kellertuer:
Since this is breaking it would be the version 0.10.0? I am fine with this version, especially since locally the other branch works. Also: Should we geht a DOI here, too?

mateuszbaran:
Yes, it looks like it should be 0.10.

I don't think we need a DOI for ManifoldsBase.jl.

kellertuer:
I also think we do not necessarily need one, but on the other hand it would also just work automatically. Well we can add it also, when we find it necessary (and usually one would use the one of Manifolds.jl anyways then I think).

PR #57, opened 2020-12-28
Introduce two new vector transports

kellertuer:
Namely DifferentiatedRetractionVectorTransport and ScaledVectorTransport.

- [x] Check test coverage

### DIFF
--- a/src/vector_transport.jl
+++ b/src/vector_transport.jl
@@ -381,6 +381,13 @@ function schilds_ladder!(
     return retract!(M, sl, p, X, retraction)
 end
 
+function show(io::IO, ::ParallelTransport)
+    return print(io, "ParallelTransport()")
+end
+function show(io::IO, m::ScaledVectorTransport)
+    return print(io, "ScaledVectorTransport($(m.method))")
+end
+
 """
     vector_transport_along(M::Manifold, p, X, c)
     vector_transport_along(M::Manifold, p, X, c, method::AbstractVectorTransportMethod)
@@ -713,7 +720,7 @@ function vector_transport_to!(
     q,
     m::ScaledVectorTransport{T},
 ) where {T<:AbstractVectorTransportMethod}
-    vector_transport_to!(M, Y, X, q, m.method)
+    vector_transport_to!(M, Y, p, X, q, m.method)
     Y .*= norm(M, p, X) / norm(M, q, Y)
     return Y
 end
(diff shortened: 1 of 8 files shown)


kellertuer:
OK, I first do not understand why the three ambiguities should be ambiguities, in my mind they are not, so I can not resolve them.

src/vector_transport.jl
```diff
@@ -88,6 +138,37 @@ struct PoleLadderTransport{
...
+
+    ScaledVectorTransport(m::AbstractVectorTransportMethod)
+
+[^SatoIwai2013]:
+    > Sato, H., Iwai, T.: _A new, globally convergent Riemannian conjugate gradient method_,
+    > Optimization, 2013, Volume 64(4), pp. 1011–1031.
+    > doi: [10.1080/02331934.2013.836650](https://doi.org/10.1080/02331934.2013.836650),
+    > arXiv: [1302.0125](https://arxiv.org/abs/1302.0125).
+"""
+struct ScaledVectorTransport{T<:AbstractVectorTransportMethod} <:
```
mateuszbaran, line 165: That paper by Sato states that scaled vector transport is not a vector transport because it's not linear. Are our `AbstractVectorTransportMethod`s supposed to be linear or not?

kellertuer: uff I don't know, I initially just thought those are easy to add and now I am for ages in fixing ambiguities again. Maybe its better to leave it out then?

mateuszbaran: I can solve these ambiguities :slightly_smiling_face:

kellertuer: But you are right, we usually assume the vector transport to be linear and all that I know are, so the question is then, how to handle this one...

kellertuer: Ah it seems the ambiguities are resolved now (still I work ages on those); so the question is, do we want to have it though it is nonlinear?

mateuszbaran: It appears that there are no ambiguities now though? The one on nightly is unrelated.

kellertuer: Well, it took me 30 minutes of pondering around to find and fix 2 or 3 of them.

mateuszbaran: In case there are some applications that require a linear vector transport, we can also have a `AbstractLinearVectorTransport` type.

kellertuer: But that would require specifying most of our already-existing ones to declare them linear, right?

mateuszbaran: Yes, of course. I can do that if you want.

kellertuer: Let me just shortly finish show & tests.

kellertuer:
Test coverage is as good as before, the - 3 lines are due to an optimization that spared a few lines.

mateuszbaran (approved):
I've added a new type for linear vector transports. I think I've marked the old transports correctly? If yes, this can be merged.

kellertuer:
Yes they are marked correctly as far as I can see; so I will merge. Thanks for the help!

PR #59, opened 2021-01-17
Bases of vector spaces

mateuszbaran:
Some work on adding bases of cotangent spaces. This is supposed to solve problems with `flat!` and `sharp!` from this PR: https://github.com/JuliaManifolds/Manifolds.jl/pull/325 .

### DIFF
--- a/src/DecoratorManifold.jl
+++ b/src/DecoratorManifold.jl
@@ -82,15 +82,15 @@ end
 # Type
 #
 """
-    AbstractDecoratorManifold{𝔽} <: Manifold{𝔽}
+    AbstractDecoratorManifold{𝔽} <: AbstractManifold{𝔽}
 
 An `AbstractDecoratorManifold` indicates that to some extent a manifold subtype
-decorates another [`Manifold`](@ref) in the sense that it either
+decorates another [`AbstractManifold`](@ref) in the sense that it either
 
 * it extends the functionality of a manifold with further features
 * it defines a new manifold that internally uses functions from the decorated manifold
 
-with the main intent that several or most functions of [`Manifold`](@ref) are transparently
+with the main intent that several or most functions of [`AbstractManifold`](@ref) are transparently
 passed through to the manifold that is decorated. This way a function implemented for a
 decorator acts transparent on all other decorators, i.e. they just pass them through. If
 the decorator the function is implemented for is not among the decorators, an error is
@@ -101,7 +101,7 @@ Transparency of functions with respect to decorators can be specified using the 
 [`@decorator_transparent_fallback`](@ref), [`@decorator_transparent_function`](@ref) and
 [`@decorator_transparent_signature`](@ref).
 """
-abstract type AbstractDecoratorManifold{𝔽} <: Manifold{𝔽} end
+abstract type AbstractDecoratorManifold{𝔽} <: AbstractManifold{𝔽} end
 
 #
 # Macros
@@ -264,7 +264,7 @@ macro decorator_transparent_function(fallback_case, input_ex)
                 return error(error_msg)
             end
             function ($fname)(
-                $(argnames[1])::Manifold,
+                $(argnames[1])::AbstractManifold,
                 $(callargs[2:end]...);
                 $(kwargs_list...),
             ) where {$(where_exprs...)}
@@ -418,7 +418,7 @@ end
 For any manifold that is a subtype of [`AbstractDecoratorManifold`](@ref), this function
 indicates whether a certain manifold `M` acts as a default decorator.
 
-This yields that _all_ functions are passed through to the decorated [`Manifold`](@ref)
+This yields that _all_ functions are passed through to the decorated [`AbstractManifold`](@ref)
 if `M` is indicated as default. This overwrites all [`is_decorator_transparent`](@ref)
 values.
 
@@ -430,7 +430,7 @@ the undecorated case.
 Value returned by this function is determined by [`default_decorator_dispatch`](@ref),
 which returns a `Val`-wrapped boolean for type stability of certain functions.
 """
-is_default_decorator(M::Manifold) = _extract_val(default_decorator_dispatch(M))
+is_default_decorator(M::AbstractManifold) = _extract_val(default_decorator_dispatch(M))
 
 """
     default_decorator_dispatch(M) -> Val
@@ -439,12 +439,12 @@ Return whether by default to dispatch the the inner manifold of
 a decorator (`Val(true)`) or not (`Val(false`). For more details see
 [`is_decorator_transparent`](@ref).
 """
-default_decorator_dispatch(M::Manifold) = Val(false)
+default_decorator_dispatch(M::AbstractManifold) = Val(false)
 
 """
-    is_decorator_transparent(f, M::Manifold, args...) -> Bool
+    is_decorator_transparent(f, M::AbstractManifold, args...) -> Bool
 
-Given a [`Manifold`](@ref) `M` and a function `f(M, args...)`, indicate, whether an
+Given a [`AbstractManifold`](@ref) `M` and a function `f(M, args...)`, indicate, whether an
 [`AbstractDecoratorManifold`](@ref) acts transparently for `f`. This means, it
 just passes through down to the internally stored manifold.
 Transparency is only defined for decorator manifolds and by default all decorators are transparent.
@@ -454,22 +454,22 @@ this behaviour, see [`decorator_transparent_dispatch`](@ref).
 If a decorator manifold is not in general transparent, it might still pass down
 for the case that a decorator is the default decorator, see [`is_default_decorator`](@ref).
 """
-function is_decorator_transparent(f, M::Manifold, args...)
+function is_decorator_transparent(f, M::AbstractManifold, args...)
     return decorator_transparent_dispatch(f, M, args...) === Val(:transparent)
 end
 
 """
-    decorator_transparent_dispatch(f, M::Manifold, args...) -> Val
+    decorator_transparent_dispatch(f, M::AbstractManifold, args...) -> Val
 
-Given a [`Manifold`](@ref) `M` and a function `f(M,args...)`, indicate, whether a
+Given a [`AbstractManifold`](@ref) `M` and a function `f(M,args...)`, indicate, whether a
 function is `Val(:transparent)` or `Val(:intransparent)` for the (decorated)
-[`Manifold`](@ref) `M`. Another possibility is, that for `M` and given `args...`
+[`AbstractManifold`](@ref) `M`. Another possibility is, that for `M` and given `args...`
 the function `f` should invoke `M`s `Val(:parent)` implementation, see
 [`@decorator_transparent_function`](@ref) for details.
 """
-decorator_transparent_dispatch(f, M::Manifold, args...) = Val(:transparent)
+decorator_transparent_dispatch(f, M::AbstractManifold, args...) = Val(:transparent)
 
-function _acts_transparently(f, M::Manifold, args...)
+function _acts_transparently(f, M::AbstractManifold, args...)
     return _val_or(
         default_decorator_dispatch(M),
         decorator_transparent_dispatch(f, M, args...),
@@ -507,7 +507,7 @@ end
 
 Return the manifold decorated by the decorator `M`. Defaults to `M.manifold`.
 """
-decorated_manifold(M::Manifold) = M.manifold
+decorated_manifold(M::AbstractManifold) = M.manifold
 
 @decorator_transparent_signature distance(M::AbstractDecoratorManifold, p, q)
 
--- a/src/DecoratorManifold.jl
+++ b/src/DecoratorManifold.jl
@@ -725,5 +725,5 @@ decorated_manifold(M::Manifold) = M.manifold
     m::SchildsLadderTransport,
 )
 
-@decorator_transparent_signature zero_tangent_vector(M::AbstractDecoratorManifold, p)
-@decorator_transparent_signature zero_tangent_vector!(M::AbstractDecoratorManifold, X, p)
+@decorator_transparent_signature zero_vector(M::AbstractDecoratorManifold, p)
+@decorator_transparent_signature zero_vector!(M::AbstractDecoratorManifold, X, p)
--- a/src/DecoratorManifold.jl
+++ b/src/DecoratorManifold.jl
@@ -491,12 +491,7 @@ end
 
 @decorator_transparent_signature check_point(M::AbstractDecoratorManifold, p; kwargs...)
 
-@decorator_transparent_signature check_tangent_vector(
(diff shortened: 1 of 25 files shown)


src/DecoratorManifold.jl
```diff
@@ -268,16 +268,18 @@ macro decorator_transparent_function(fallback_case, input_ex)
...
+                return error(
+                    string(
+                        ManifoldsBase.manifold_function_not_implemented_message(
+                            $(argnames[1]),
+                            $fname,
+                            $(argnames[2:end]...),
+                        ),
+                        " Usually this is implemented for a ",
+                        $(argtypes[1]),
+                        ". Maybe you missed to implement this function for a default?",
```
kellertuer, line 280: “this function for a default” – you are referring to a default type? Otherwise one could also write „a default for this function“.

kellertuer (approved):
So this basically introduces a second parameter to the basis indicating what the basis is for? Looks ok to me, I just noticed a minor rephrasing of an error message we could do.

kellertuer (changes requested):
Sorry I did not want to approve this, since tests are still missing,

src/DecoratorManifold.jl
mateuszbaran: This text was there before. Now I'm not quite sure what was it supposed to say.

mateuszbaran:
> So this basically introduces a second parameter to the basis indicating what the basis is for? Looks ok to me, I just noticed a minor rephrasing of an error message we could do.

Yes, it's essentialy this, as well as moving `FVector` to ManifoldsBase.jl and adding a field in it for the basis. The new parameter for basis is not entirely necessary but it looked weird to have a basis and not knowing what kind of space it is a basis of.

src/DecoratorManifold.jl
kellertuer: Ah, I did not notice that; we could still rephrase it :)

kellertuer:
Since it is backwards compatible, I think it's fine to add the parameter.

mateuszbaran:
I just noticed that the vector space is indeed needed in the basis, to be able to correctly interpret coordinates in `get_vector`.

kellertuer:
That's why I carefully asked – we can surely keep in that order and yes, ambiguity fighting is not nice, for sure. Thanks for the explanations.

> Yes, that's equivalent, I can use this instead of `<:Any`.

Great. I think that looks a little nicer

mateuszbaran:
Non-implemented error messages are quite unhelpful when debugging wrong `invoke`s. I've removed one, maybe we should remove all of them?

kellertuer:
We discussed that anyways following Lyndons notes, I would say it's reasonable to remove them, yes.

mateuszbaran:
I think this is ready now. The one new coverage failure looks like a false positive.

mateuszbaran:
While improving documentation I noticed that `vector_space_dimension` is referenced in ManifoldsBase.jl but defined in Manifolds.jl. I think we could change its signature to make it take separately manifold and vector space type and move it to ManifoldsBase.jl as well.

mateuszbaran:
CI seems to fail for no reason :disappointed: .

src/ManifoldsBase.jl
```diff
@@ -830,29 +830,30 @@ Converts a size given by `Tuple{N, M, ...}` into a tuple `(N, M, ...)`.
 Base.@pure size_to_tuple(::Type{S}) where {S<:Tuple} = tuple(S.parameters...)
 
 """
-    zero_tangent_vector!(M::Manifold, X, p)
+    zero_vector!(M::Manifold, X, p)
 
 Save to `X` a vector such that retracting `X` to the [`Manifold`](@ref) `M` at `p`
```
kellertuer, line 835:
```suggestion
Save to `X` the tangent vector from the tangent space ``T_p\mathcal M`` at `p`,  that represents the zero vecto, i.e. such that retracting `X` to the [`Manifold`](@ref) `M` at `p`
```

src/ManifoldsBase.jl
```diff
@@ -830,29 +830,30 @@ Converts a size given by `Tuple{N, M, ...}` into a tuple `(N, M, ...)`.
...
 produces `p`.
 """
-zero_tangent_vector!(M::Manifold, X, p) = log!(M, X, p, p)
+zero_vector!(M::Manifold, X, p) = log!(M, X, p, p)
 
 """
-    zero_tangent_vector(M::Manifold, p)
+    zero_vector(M::Manifold, p)
 
 Return the tangent vector from the tangent space at `p` on the [`Manifold`](@ref) `M`, that
```
kellertuer, line 843:
```suggestion
Return the tangent vector from the tangent space ``T_p\mathcal M`` at  `p` on the [`Manifold`](@ref) `M`, that
```

kellertuer (approved):
This is fine with me, moving vector space stuff here and even removing one `ErrorException`to be a `MethodError` now, nice.

mateuszbaran:
Thanks! I'll still wait a bit with merging this, until https://github.com/JuliaManifolds/Manifolds.jl/pull/325#pullrequestreview-615177351 gets more polished.

mateuszbaran:
Since this is breaking anyway, I used this PR to clean up `TVector/CoTVector` a bit here.

mateuszbaran:
I also think about generalizing `is_tangent_vector` to any vector spaces, similarly to `zero_vector`. What do you think about renaming it to `is_vector`?

kellertuer:
Can we maybe (to be nonbreaking) keep `is_tangent_vector` and map it to the more general variant of `is_vector` (I think with a tangent space argument inside?)

I don't like the term `AbstractMVector` what's an MVector? compared to MPoint it is a ManifoldVector? That does not make sense.
I would be fine with calling MPoint as `ManifoldPoint` for completeness, or `AbstractManifoldPoint` to be precise and consistent (MPoint was a name I introduced when I was only vaguely aware of naming schemes).
But this here should be an `AbstractTangentVector` I think, though carrying a type and especially the `Abstract` prefix seem really reasonable

And since we a re at it – if we are precise `Manifold` would also be `AbstractManifold`? But then – this is a lot of other changes for a PR that was concerned with vectors and bases ;)

mateuszbaran:
> Can we maybe (to be nonbreaking) keep `is_tangent_vector` and map it to the more general variant of `is_vector` (I think with a tangent space argument inside?)

Yes, we can, though for `zero_vector` we decided earlier to just drop `zero_tangent_vector`, and this is a very similar situation.

> I don't like the term `AbstractMVector` what's an MVector? compared to MPoint it is a ManifoldVector? That does not make sense.
> I would be fine with calling MPoint as `ManifoldPoint` for completeness, or `AbstractManifoldPoint` to be precise and consistent (MPoint was a name I introduced when I was only vaguely aware of naming schemes).
> But this here should be an `AbstractTangentVector` I think, though carrying a type and especially the `Abstract` prefix seem really reasonable

> _edit_: Ah I missed that the abstract case is meant for both cotangent and tangents, then maybe `AbstractFibreVector` or `AbstractManifoldFibreVector` - does that sound reasonable?

I don't really like `AbstractMVector` either (naming is hard :sweat_smile:). It's supposed to be a vector from a fiber of a vector bundle, for example a tangent vector or a cotangent vector -- so that we won't have to repeat everything for tangent vectors, cotangent vectors and other more complicated constructs. 

I can surely change `MPoint` to `AbstractManifoldPoint`. Also `AbstractFibreVector` sounds nice.

> And since we a re at it – if we are precise `Manifold` would also be `AbstractManifold`? But then – this is a lot of other changes for a PR that was concerned with vectors and bases ;)

Sure, why not? If there are many breaking changes, I think it's better to bundle them together instead of making multiple breaking releases one after another :slightly_smiling_face: .

kellertuer:
Ah sure, without the manifold, so just `AbstractFibreVector`is fine with me. And sure, if we break something, let's also break the manifold then ;)

mateuszbaran:
So, what do we finally do about `is_tangent_vector`?

kellertuer:
My small problem, alsready with `zero_vector` is, that the function name does not fully reflect, what the function is doing; while for the `zero_vector` I can see the consistency with all of Julia (and the first two arguments help), I feel the same for `is_vector` – the name itself does not reflect the main thing to check – that it's a tangent (or cotangent check then, too?). With that argument we could also rename `is_manifold_point` to `is_point`, since the `M` is in both cases anyway necessary to unterstand what the function is doing – for the `is_vector` even the `p` is necessary, to see where the tangent space is at, ok.

Maybe that's my concern, that the function name itself does not provide enough detail to understand what the function is doing, but you have to look for the whole signature.
A plut point is, that the function name is shorter. But also that even more yields, that we should then also switch to `is_point(M,p)`, too. The implicit knowledge vor the package would then be that vector means (co)tangent vector and point means point on the manifold.

mateuszbaran:
Yes, `is_vector` is supposed to work on cotangent vectors as well.

Another option I can see here is renaming `zero_tangent_vector` -> `zero_fibre_vector` and `is_tangent_vector` -> `is_fibre_vector`.

Maybe @sethaxen has some opinion on naming as well?

mateuszbaran:
OK. I think we don't need both `is_manifold_point` and `is_point` so I think we should pick one of them. `is_vector` is needed anyway and one remaining issue here is whether `is_vector` should call `is_tangent_vector` or the other way round. I think it will be extremely rare to overload `is_vector` (or `zero_vector`) for manifolds, for vector spaces other than the tangent space.

kellertuer:
> OK. I think we don't need both `is_manifold_point` and `is_point` so I think we should pick one of them. `is_vector` is needed anyway and one remaining issue here is whether `is_vector` should call `is_tangent_vector` or the other way round. I think it will be extremely rare to overload `is_vector` (or `zero_vector`) for manifolds, for vector spaces other than the tangent space.

We overload/define `zero(_tangent)_vector!` quite often, I think; will that get obsolete, so will there be a good default?

So, I think if we drop backwards compatibility, then we should get rid of `is_tangent_vector` and `is_manifold_point` and just use `is_point` and `is_vector`, though both names are not completely self-descriptive (you have to have in mind point->manifold, vector->tangent space or any other fibre).

mateuszbaran:
> We overload/define `zero(_tangent)_vector!` quite often, I think; will that get obsolete, so will there be a good default?

Yes, we have some overloads for tangent vectors. What I mean is that for any _other_ fibre there would usually be decent defaults. For tangent vectors it won't be obsolete.

> So, I think if we drop backwards compatibility, then we should get rid of `is_tangent_vector` and `is_manifold_point` and just use `is_point` and `is_vector`, though both names are not completely self-descriptive (you have to have in mind point->manifold, vector->tangent space or any other fibre).

This patch already breaks quite a bit so I wouldn't worry about backwards compatibility. I'll get rid of `is_tangent_vector` and `is_manifold_point` then.

kellertuer:
Ok, let's just make sure, that @sethaxen at some point agrees, too; but I imagine he is also quite busy currently.

sethaxen:
Sorry, there's a lot of discussion here. Can someone summarize the current proposal/disagreement?

kellertuer:
I try to do summarise in order of renaming, since that was a trail of thoughts all this followed. The first is function names

1. `zero_tangent_vector` is renamed to `zero_vector` for consistency with Julia
2. `is_tangent_vector` is renamed to `is_vector` for consistency with 1.

Here my criticism is, that now the function name bears a little less information about what the function is doing, but if you have the full signature `is_vector(M, p, X)` it's fine. Even more these functions can now be used for any fibres – for the first maybe with an optional parameter, for the second depending on the type of `X`. So especially for co tangent vectors, these also should work.

3. `is_manifold_point` is renamed to `is_point` since that is then consistent with 2. – for both one now needs the implicit knowledge either that points are on manifolds and vectors are in tangent spaces or the full signature `is_point(M,p)`. For this I am ok, since I think I Walt to move to using `p ∈ M` anyways

Then we came across types

4. `MPoint` is renamed to `AbstractManifoldPoint`, since that is more informative and the `Abstract` is more consistent with the rest of our naming scheme
5. `TVector` and `CoTVector` get an abstract super type `AbstractFibreVector`, and while I writing this summary, we should actually call them `AbstractTangentVector` and `AbstractCoTangentVector`, respectively.

Finally since it is breaking everything anyways already here

6. `Manifold` is renamed to `AbstractManifold` to be more consistent.

4.-6. were actually names I introduced, I think, when I started Manopt.jl and when I was not completely aware of the Julia naming scheme, so I think, since this is breaking anyways and renames a lot, this is a good time to introduce this consistency.

That‘s the proposal and I think currently there is no disagreement really – I think there is a small tradeoff, that `is_vector` bears a little too less information (without the rest of the signature), but it is good that it then can be used on arbitrary fibers. 

But for sure, we would like to know your opinion, too.

mateuszbaran:
That's a good summary :slightly_smiling_face: . I'd just add that I want to do some generalization here for arbitrary fibers because PR https://github.com/JuliaManifolds/Manifolds.jl/pull/325 finally makes it possible to handle them well.

kellertuer:
I just noticed, since we are renaming everything to shorten names, what about changing `check_base_point` to `check_point` (and document that a little better in `check_tangent_vector`. What do you think?

mateuszbaran:
Sure, that would be consistent with the other changes.

mateuszbaran:
I've formatted code locally using JuliaFormatter 0.13.10. Here JuliaFormatter 0.14 was used and it crashed on parsing our files :confused: .

kellertuer:
don‘t we have check_base mentioned here somewhere? I think at least is_vector (or check vector) should mention it in its doctoring. I will investigate the Julia<formatter issue tomorrow (can reproduce it even on master locally here).

mateuszbaran:
I've submitted an issue: https://github.com/domluna/JuliaFormatter.jl/issues/410 .

mateuszbaran:
What do you mean by `check_base`?

kellertuer:
Cool, thanks for submitting that.

Ah yes, I mean `check_base_point`which we might want to call `check_point`? That should be mentioned maybe in both doc strings even.

mateuszbaran:
OK, I see, that `check_base_point`. I think `check_base_point` is still fine? A little verbosity there doesn't look bad and the name is still correct.

mateuszbaran:
I'm also thinking that we probably should just implement something like `_check_vector` for each manifold, and do something like
```julia
function check_vector(M::Manifold, p, X; check_base_point=true, kwargs...)
    if check_base_point
        mpe = check_point(M, p; kwargs...)
        mpe === nothing || return mpe
    end
    return _check_vector(M, p, X; kwargs...)
end
```
so that `check_vector` wouldn't have to duplicate the logic for checking the base point for each manifold.

kellertuer:
Maybe the easier solution is to only have `check_base(_point)` for `is_vecor` (i.e. the high level interface) and not for `check_*`(i.e. the lower one)?

kellertuer:
Cool, one can see on Manifolds.jl how much redundancy this removes; great!

PR #60, opened 2021-02-06
Adds an explicit zero_tangent_vector! for PowerManifold.

kellertuer:
Instead of first falling back to log, this first called zero_tangent_vector element wise.


mateuszbaran (approved):
Looks fine. Could you also bump the patch version to tag a new release?

PR #61, opened 2021-03-04
Fixes the case of nested manifolds to do a single getindex if we access an element

kellertuer:
This PR fixes the following example (of manifolds)

```julia
using Manifolds
a = zeros(3)
N = PowerManifold(Circle(), NestedPowerRepresentation(), 3)
a[N,1]
```
which previously returned a view to a `0`-dimensional array (i.e. `a[N,1][]` could access the number again).

While I think this fix is nice, there is a point I am not sure about:

If `represenation_size(M)` is not `()` the single access to the one element can (and should?) better still be a `view` I think? Or is it for the nested case, that the inner “refrence” is returned anyways? Then the fix here should be enough.

### DIFF
--- a/src/PowerManifold.jl
+++ b/src/PowerManifold.jl
@@ -234,10 +234,7 @@ The tolerance for the last test can be set using the `kwargs...`.
 """
 function check_manifold_point(M::AbstractPowerManifold, p; kwargs...)
     rep_size = representation_size(M.manifold)
-    e = [
-        (i, check_manifold_point(M.manifold, _read(M, rep_size, p, i); kwargs...)) for
-        i in get_iterator(M)
-    ]
+    e = [(i, check_manifold_point(M.manifold, p[M, i]; kwargs...)) for i in get_iterator(M)]
     errors = filter((x) -> !(x[2] === nothing), e)
     cerr = [ComponentManifoldError(er...) for er in errors]
     (length(errors) > 1) && return CompositeManifoldError(cerr)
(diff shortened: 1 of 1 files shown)


kellertuer:
This might still be a little tricky, since allocation is different, if one sees the `Circle()^5` as a nested power manifold.

mateuszbaran:
Why did you replace `_read` with indexing? I prefer to keep that `_read` there. The `getindex` change should be sufficient but that no-`collect` path should I think be used for both nested and array power manifolds?

kellertuer:
Because `_read` does not work for nested `Circle`s, that's why we do no call it for them but do getindex directly without a read. Maybe you see a better way, but this nearly works (up to allocation for nested circle).
Leaving it as `reads leads to the same errors I had before changing the get index, since you then still call reads and the reads do not work with power of circle...

To be honest I am not sure whether `PowerManifold(Circle(),3)` is actually nested (the elements are floats so its basically nested) or array (which it is, too).

And I even think the code is more readable now with all the `p[M,i]` calls, but that's maybe only a personal preference. I really like this variant here due to its readability.

I am also not yet sure how to fix the allocation, (for example for log on power of circle), I haven't had time to fix that.

Feel free to propose a better way, you have more experience on the power manifolds implementation than I have.

kellertuer:
Addendum: I would also be fine with the resolution, that this is not a nested one we are speaking about and this is just an Array one – if it is fixed. I am sure you'll find it :)

This is also the last problem I have with the current Manopt PR, all other stuff now works.

mateuszbaran:
`_read` will be necessary for performance when I change how indexing works in `HybridArrays.jl`. It provides a convenient place for overriding default indexing for specific array types. I'll redo this PR.

kellertuer:
Ok, then I have no idea how this is supposed to work and look forward to hopefully learn this from your changes (and whether changes to the array style in Manifold.jl are necessary).

For reference, even on master the current example https://github.com/JuliaManifolds/Manopt.jl/blob/master/examples/CyclicProximalPoint/S1_Signal_TV12_CPPA.jl fails if I remember correctly (it definitely does on the new PR over there).

PR #62, opened 2021-03-05
Fix power indexing for the case of empty representation_size and inte…

mateuszbaran:
This PR should replace #61 .


mateuszbaran:
There is something weird going on... I need to investigate it a bit more.

kellertuer:
Thanks for taking a look :) This looks simpler than my PR.

kellertuer (approved):

mateuszbaran:
It looks like this change breaks `Manifolds.jl`. Do we tag 0.11.0 or consider it a bugfix and immediately release a fix for `Manifolds.jl`?

kellertuer:
I would say it's a bugfix, since there was no way the `representation_size=()` cases worked with power manifolds.

Let's wait for my proper fix on `Project.toml` (I hope I found something) over on Manifolds, such that your tests and docs generation works in the sister-branch – then we should release both short after each other.

PR #63, opened 2021-03-08
Adds integrations tests workflow

### LINKED ISSUE #58: Add integration test with Manifolds
We should have an integration test with Manifolds, that is, an action that runs Manifolds's test suite using the latest changes to ManifoldsBase. This will ensure that potentially breaking changes are identified before being merged. This workflow is really nice: https://github.com/JuliaDiff/ChainRulesTestUtils.jl/pull/105

kellertuer:
Currently this integration test covers Manifolds.jl
It extends tests in time by quite a bit, but on the other hand, this package is not PR'ed that often and if so we want to be sure it does not break Manifolds.jl.

This resolves #58.

### DIFF
--- a/.github/workflows/IntegrationTest.yml
+++ b/.github/workflows/IntegrationTest.yml
@@ -1,0 +1,47 @@
+name: IntegrationTest
+on:
+  push:
+    branches: [master]
+    tags: [v*]
+  pull_request:
+
+jobs:
+  test:
+    name: ${{ matrix.package.repo }}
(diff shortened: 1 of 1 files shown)


kellertuer:
Python seems to be missing...hm.

mateuszbaran:
That's a really weird issue... maybe let's just merge it and hope it goes away soon? I have no idea what could cause that.

mateuszbaran:
Yes, having a reduced set of tests would make sense but we can leave it for later.

PR #64, opened 2021-03-23
Make distance dispatch to parent for embedded manifold

sethaxen:
Fixes part of https://github.com/JuliaManifolds/Manifolds.jl/issues/338.

### DIFF
--- a/src/EmbeddedManifold.jl
+++ b/src/EmbeddedManifold.jl
@@ -49,7 +49,7 @@ abstract type AbstractIsometricEmbeddingType <: AbstractEmbeddingType end
 """
     DefaultIsometricEmbeddingType <: AbstractIsometricEmbeddingType
 
-An isometric embedding type that acts as a default, i.e. it has no specifig properties
+An isometric embedding type that acts as a default, i.e. it has no specific properties
 beyond its isometric property.
 """
 struct DefaultIsometricEmbeddingType <: AbstractIsometricEmbeddingType end
--- a/src/EmbeddedManifold.jl
+++ b/src/EmbeddedManifold.jl
@@ -437,7 +437,7 @@ function decorator_transparent_dispatch(
     return Val(:transparent)
 end
 function decorator_transparent_dispatch(::typeof(norm), ::AbstractEmbeddedManifold, args...)
-    return Val(:intransparent)
+    return Val(:parent)
 end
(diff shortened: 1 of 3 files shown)


src/EmbeddedManifold.jl
```diff
@@ -252,6 +252,20 @@ function decorator_transparent_dispatch(
...
+function decorator_transparent_dispatch(
+    ::typeof(distance),
+    ::AbstractEmbeddedManifold,
+    args...,
+)
+    return Val(:parent)
+end
+function decorator_transparent_dispatch(
+    ::typeof(distance),
+    ::AbstractEmbeddedManifold{𝔽,<:AbstractIsometricEmbeddingType},
```
mateuszbaran, line 264: I think this should be restricted to `TransparentIsometricEmbedding`.
```suggestion
    ::AbstractEmbeddedManifold{𝔽,TransparentIsometricEmbedding},
```

sethaxen: That's interesting. `exp` and `log` both have `TransparentIsometricEmbedding` here, while `norm` has `AbstractIsometricEmbeddingType`. All three depend on the metric though. Why the different overloads for these three?

mateuszbaran: Actually, I have no idea why `norm` is `transparent` for `AbstractIsometricEmbedding`. As far as I understand it, `norm` should be `parent` for non-transparent embedding and `transparent` for transparent embedding. Maybe @kellertuer knows why it's different?

mateuszbaran: Ah, right, I was wrong. It's correct. `AbstractIsometricEmbedding` means that `inner` and `norm` can be transparently forwarded while `exp` and `log` can only be directly forwarded for `TransparentIsometricEmbedding`. That's correct. That's basically what these embedding types mean.

mateuszbaran: Still, `norm` should probably be `:parent` for `AbstractEmbeddedManifold`?

mateuszbaran: @sethaxen Why did you change it back to `<:AbstractIsometricEmbeddingType`?

sethaxen: Whoops, accident.

mateuszbaran (approved):

kellertuer (approved):

PR #65, opened 2021-04-18
introduce copyto! for points and tangent vectors.

kellertuer:
As discussed in JuliaManifolds/Manopt.jl#78 it might be beneficial to have a `copyto!` already on the manifold level.
This PR introduces that. There are a few points to check/todo.

* [ ] it it reasonable to introduce transparency? For the validation manifold, the copyto! just solves this, but for group and metric? Sure it would also just be one line unpacking the manifold, I think.
* [ ] Technically this is breaking, since the interface for the copyto! for the ValidationManifold changed – but did we use than anywhere?
* [ ] There seems to be an ambiguity in some code that I did not tough (for 1.7 this is anticipated but I also get that on 1.6 concerning the `getindex`. Should this be resolved here? The CI is running on 1.5 still anyways.
* [ ] Test coverage

### DIFF
--- a/src/ManifoldsBase.jl
+++ b/src/ManifoldsBase.jl
@@ -298,6 +298,25 @@ function check_size(M::Manifold, p, X)
     end
 end
 
+@doc raw"""
+    copyto!(M::Manifold, q, p)
+
+Copy the value(s) from `p` to `q`, where both are points on the [`Manifold`](@ref) `M`.
+This function defaults to calling `copyto!(q, p)`, but it might be useful to overwrite the
+function at the level, where also information from `M` can be accessed.
+"""
+copyto!(::Manifold, q, p) = copyto!(q, p)
+
+@doc raw"""
+    copyto!(M::Manifold, Y, X)
+
+Copy the value(s) from `X` to `Y`, where both are tangent vector from the tangent space at
+`p` on the [`Manifold`](@ref) `M`.
+This function defaults to calling `copyto!(Y, X)`, but it might be useful to overwrite the
+function at the level, where also information from `p` and `M` can be accessed.
+"""
+copyto!(::Manifold, Y, p, X) = copyto!(Y, X)
+
 """
     distance(M::Manifold, p, q)
 
--- a/src/PowerManifold.jl
+++ b/src/PowerManifold.jl
@@ -292,6 +292,38 @@ function check_tangent_vector(
     return nothing
 end
 
+@doc raw"""
+    copyto!(M::AbstractPowerManifold{𝔽,<:Manifold{𝔽},NestedPowerRepresentation}, q, p)
+
+copy the values elementwise, i.e. call `copyto!(M.manifold, b, a)` for all elements `a`and
+`b` of `p` and `q`, respectively.
+"""
+function copyto!(M::NestedPowerRepresentation, q, p)
+    rep_size = representation_size(M.manifold)
+    for i in get_iterator(M)
+        copyto!(M.manifold, _write(M, rep_size, q, i), _read(M, rep_size, p, i))
+    end
+    return q
+end
+
+@doc raw"""
+    copyto!(M::AbstractPowerManifold{𝔽,<:Manifold{𝔽},NestedPowerRepresentation}, Y, p, X)
+
+copy the values elementwise, i.e. call `copyto!(M.manifold, B, a, A)` for all elements
+`A`, `a` and `B` of `X`, `p`, and `Y`, respectively.
+"""
+function copyto!(M::NestedPowerRepresentation, Y, p, X)
+    rep_size = representation_size(M.manifold)
+    for i in get_iterator(M)
+        copyto!(
+            M.manifold,
+            _write(M, rep_size, Y, i),
+            _read(M, rep_size, p, i),
+            _read(M, rep_size, X, i),
+        )
+    end
+    return Y
+end
 
 @doc raw"""
     distance(M::AbstractPowerManifold, p, q)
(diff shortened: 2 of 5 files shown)


mateuszbaran:
> * it it reasonable to introduce transparency? For the validation manifold, the copyto! just solves this, but for group and metric? Sure it would also just be one line unpacking the manifold, I think.

Yes, I think transparency definitely makes sense for `copyto!`.

> * Technically this is breaking, since the interface for the copyto! for the ValidationManifold changed – but did we use than anywhere?

Actually I've never used `ValidationManifold`. I think it's OK to do this small change in a non-breaking release.

> * There seems to be an ambiguity in some code that I did not tough (for 1.7 this is anticipated but I also get that on 1.6 concerning the `getindex`. Should this be resolved here? The CI is running on 1.5 still anyways.

I remember there was an ambiguity with some `getindex` on a sparse matrix type. It's extremely unlikely to occur in practice IMO so we can leave that.

src/ManifoldsBase.jl
```diff
@@ -298,6 +298,25 @@ function check_size(M::Manifold, p, X)
...
+    copyto!(M::Manifold, q, p)
+
+Copy the value(s) from `p` to `q`, where both are points on the [`Manifold`](@ref) `M`.
+This function defaults to calling `copyto!(q, p)`, but it might be useful to overwrite the
+function at the level, where also information from `M` can be accessed.
+"""
+copyto!(::Manifold, q, p) = copyto!(q, p)
+
+@doc raw"""
+    copyto!(M::Manifold, Y, X)
```
mateuszbaran, line 311:
```suggestion
    copyto!(M::Manifold, Y, p, X)
```

src/ManifoldsBase.jl
```diff
@@ -298,6 +298,25 @@ function check_size(M::Manifold, p, X)
...
+
+Copy the value(s) from `X` to `Y`, where both are tangent vector from the tangent space at
```
mateuszbaran, line 313:
```suggestion
Copy the value(s) from `X` to `Y`, where both are tangent vectors from the tangent space at
```

src/PowerManifold.jl
```diff
@@ -292,6 +292,38 @@ function check_tangent_vector(
     return nothing
 end
 
+@doc raw"""
+    copyto!(M::AbstractPowerManifold{𝔽,<:Manifold{𝔽},NestedPowerRepresentation}, q, p)
+
+copy the values elementwise, i.e. call `copyto!(M.manifold, b, a)` for all elements `a`and
```
mateuszbaran, line 298:
```suggestion
Copy the values elementwise, i.e. call `copyto!(M.manifold, b, a)` for all elements `a` and
```

src/PowerManifold.jl
```diff
@@ -292,6 +292,38 @@ function check_tangent_vector(
...
+    for i in get_iterator(M)
+        copyto!(M.manifold, _write(M, rep_size, q, i), _read(M, rep_size, p, i))
+    end
+    return q
+end
+
+@doc raw"""
+    copyto!(M::AbstractPowerManifold{𝔽,<:Manifold{𝔽},NestedPowerRepresentation}, Y, p, X)
+
+copy the values elementwise, i.e. call `copyto!(M.manifold, B, a, A)` for all elements
```
mateuszbaran, line 312:
```suggestion
Copy the values elementwise, i.e. call `copyto!(M.manifold, B, a, A)` for all elements
```

src/ValidationManifold.jl
```diff
@@ -107,16 +107,16 @@ function convert(::Type{ValidationTVector{V}}, X::V) where {V<:AbstractArray{<:N
     return ValidationTVector{V}(X)
 end
 
-function copyto!(p::ValidationMPoint, q::ValidationMPoint)
-    copyto!(p.value, q.value)
+function copyto!(M::ValidationManifold, p::ValidationMPoint, q::ValidationMPoint)
+    copyto!(M.manifold, p.value, q.value)
```
mateuszbaran, line 111:
```suggestion
    copyto!(M.manifold, p.value, q.value)
```
Shouldn't this check that `q` is a correct point? Also maybe we should be consistent in argument naming with other methods and swap the names of `p` and `q`?

src/ValidationManifold.jl
```diff
@@ -107,16 +107,16 @@ function convert(::Type{ValidationTVector{V}}, X::V) where {V<:AbstractArray{<:N
...
 end
-function copyto!(p::ValidationCoTVector, q::ValidationCoTVector)
-    copyto!(p.value, q.value)
+function copyto!(M::ValidationManifold, p::ValidationCoTVector, q::ValidationCoTVector)
+    copyto!(M.manifold, p.value, q.value)
     return p
 end
-function copyto!(Y::ValidationTVector, X::ValidationTVector)
-    copyto!(Y.value, X.value)
+function copyto!(M::ValidationManifold, Y::ValidationTVector, X::ValidationTVector)
```
mateuszbaran, line 118:
```suggestion
function copyto!(M::ValidationManifold, Y::ValidationTVector, p::ValidationMPoint, X::ValidationTVector)
```

src/ValidationManifold.jl
kellertuer: Oh, naming is inconsistent here, yes.

I think it should actually check both (input p and output q) since if you are looking for errors one might be that someone did a mistake in implementing `copyto!`?

mateuszbaran: Checking the output doesn't necessarily make sense when it was just allocated. We could just check for the correct shape but we don't have any methods for that yet.
And this method also checks that someone assigns a point here, and not for example a tangent vector to a point.

kellertuer: but `copyto!` copies the values so just before `return q` is should be a valid point on the manifold.

mateuszbaran: Right, I didn't think about that. We could check that as well :slightly_smiling_face: .

mateuszbaran (approved):
I think this looks OK and could be merge when all lines are covered by tests.

kellertuer:
I will check test coverage by tonight; what do you think about the version? Minor or major increase?

mateuszbaran:
I actually think patch version increase is OK :slightly_smiling_face: (as you already did in this PR).

kellertuer:
This is a little tricky since Codecov seems broken and I can‘t see directly which lines might be missing.

mateuszbaran:
Yes, codecov indeed seems to be broken. I'm fine with merging it as-is.

kellertuer:
Cool, thanks; I will surely address the missing lines in a future PR then.

PR #68, opened 2021-05-13
Ensure norm succeeds on numerical errors in inner(...)

TimQuelch:
Due to numerical errors the inner product can sometimes return an infinitesimal negative number (e.g. -1e-26). This negative number causes the sqrt to fail with a domain error

This fix ensures that the result of the inner product of X and X is always greater than or equal to zero (which it should be by definition)


kellertuer:
Following the general idea, that we have nice default implementations, that work (flawlessly) but might not be the most efficient one (for a special manifold, where a specific one is for some reason more efficient), I think this is a very good idea.

mateuszbaran (approved):

kellertuer:
Could you just bump the version to `0.10.7`? Then we would be good to go (i.e. merge).

kellertuer (approved):

TimQuelch:
I bumped the version. I also force-pushed to reword the commit message to be more correct

PR #71, opened 2021-05-19
Improve docs of `FVector`

mateuszbaran:
Addresses this: https://github.com/JuliaManifolds/Manifolds.jl/pull/325#issuecomment-844476404


kellertuer:
Should we maybe say, how that actually can be done? I am asking because I don‘t see that directly.

mateuszbaran:
Ah, sure, I'll add that.

mateuszbaran:
I've added an example, exported `TFVector` and `CoTFVector` and added a nice `show` method for these two types.

kellertuer (approved):

PR #75, opened 2021-06-08
Introduce different decorator types

kellertuer:
This PR is WIP to introduce a nicer (shorter) way of introducing decorators.
The `AbstractDecoratorType` is a supertype of the embedding type we already have, on the abstract level of the decorator manifold and is meant to work similarly.

This should currently already work, is backward compatible for everything “below” the AbstractDecoratorManifold, while this type itself got a second parameter (so in general that should also still work, I hope).

For example the new `DefaultMetricDecorator` can be used for completely new features, i.e. when a decorator introduces functions but does not affect old ones. 

* [x] introduce the new type and clean up the two existing decorators to have less dispatch rules defined.
* [x] check for easy entrance types to also provide, i.e. something like `DefaultMetricDecorator` which acts transparent unless the metric is involved, and ideas like that.
* [x] update documentation
* [x] if we have a good idea how to restyle the decorator approach, i.e. internal optimization, that would also be good

### DIFF
--- a/src/DecoratorManifold.jl
+++ b/src/DecoratorManifold.jl
@@ -788,7 +788,7 @@ DEFAULT_PARENT_FUNCTIONS = [
     vector_transport_to,
 ]
 
-for f in PARENT_FUNCTIONS
+for f in DEFAULT_PARENT_FUNCTIONS
     eval(
         quote
             function decorator_transparent_dispatch(
--- a/src/DecoratorManifold.jl
+++ b/src/DecoratorManifold.jl
@@ -774,7 +774,7 @@ Base.@propagate_inbounds function Base.getindex(
     return getindex(p, decorated_manifold(M), I...)
 end
 
-PARENT_FUNCTIONS = [
+DEFAULT_PARENT_FUNCTIONS = [
     distance,
     exp,
     inner,
--- a/src/DecoratorManifold.jl
+++ b/src/DecoratorManifold.jl
@@ -93,8 +93,6 @@ Note that for a function `f` and it's mutating variant `f!`
 """
 abstract type AbstractDecoratorType end
 
-
-
 """
     DefaultDecoratorType <: AbstractDecoratorType
 
@@ -128,7 +126,7 @@ Transparency of functions with respect to decorators can be specified using the 
 There are currently three modes given a new `AbstractDecoratorManifold` `M`
 * `:intransparent` – this function has to be implmented for the new manifold `M`
 * `:transparent` – this function is transparent, in the sense that the function is invoked
-  on the decorated `M.manifold`
+  on the decorated `M.manifold`. This is the default, when introducing a function or signature.
 * `:parent` specifies that (unless implemented) for this function, the classical inheritance
   is issued, i.e. the function is invoked on `M`s supertype.
 """
@@ -775,10 +773,6 @@ Base.@propagate_inbounds function Base.getindex(
 ) where {𝔽}
     return getindex(p, decorated_manifold(M), I...)
 end
-
-#
-# Dispatch rules
-#
 
 PARENT_FUNCTIONS = [
     distance,
--- a/src/DecoratorManifold.jl
+++ b/src/DecoratorManifold.jl
@@ -79,10 +79,34 @@ function _split_function(ex::Expr)
 end
 
 #
+# Transparency types
+#
+"""
+    AbstractDecoratorType
+
+Decorator types can be used to specify a basic transparency for an [`AbstractDecoratorManifold`](@ref).
+This can be seen as an initial (rough) transparency pattern to start a type with.
+
+Note that for a function `f` and it's mutating variant `f!`
+* The function `f` is set to `:parent` to first invoke allocation and call of `f!`
+* The mutating function `f!` is set to `transparent`
+"""
+abstract type AbstractDecoratorType end
+
+
+
+"""
+    DefaultDecoratorType <: AbstractDecoratorType
+
+A default decorator type, where all new functions are transparent by default.
+"""
+struct DefaultDecoratorType <: AbstractDecoratorType end
+
+#
 # Type
 #
 """
-    AbstractDecoratorManifold{𝔽} <: AbstractManifold{𝔽}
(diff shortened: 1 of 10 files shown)


kellertuer:
We till have to check whether we can introduce some nice type here.
What I did for now

1. There is now a(n abstract) transparency type on the decorator level. The old EmbeddingManifold transparency inherits from this.
2. There is a default transparency type which is just
  * `:parent` for all non-mutating functions (so they go to the parent, allocate and call the mutating one) – explicitly set
  * `:transparent` for all mutating functions (by default definition)
3. s´this already sparsified what we have to define on the `EmbeddedManifold` side.
4. I would like to not define `MetricTransparencyType` nor `GroupTransparencyType`, since both are not for types defined here but over on manifolds
5. I optimised codelength
6. unified how along/direction/to vector transports are handled – just have to adapt tests, too.

Where could we improve documentation? I do not think we can do that much here, but it has to be done over on Manifolds.jl, then.

kellertuer:
This should be ok now; note that it is breaking, since the `AbstractDecoratorManifold` has now a new parameter.

kellertuer:
Ah, one just has to add the type to the MetricManifold super type (maybe introduce a `MetricTransparencyType` and import the abstract type accordingly. For Group Manifold the Embedded Manifold just does the job, but we should refine that then.

src/DecoratorManifold.jl
```diff
@@ -739,8 +768,36 @@ decorated_manifold(M::AbstractManifold) = M.manifold
...
     p::AbstractArray,
-    M::AbstractDecoratorManifold,
+    M::AbstractDecoratorManifold{𝔽,<:AbstractDecoratorType},
     I::Union{Integer,Colon,AbstractVector}...,
-)
+) where {𝔽}
     return getindex(p, decorated_manifold(M), I...)
 end
+
+PARENT_FUNCTIONS = [
```
mateuszbaran, line 777:
```suggestion
DEFAULT_PARENT_FUNCTIONS = [
```
I think a more specific name would be better.

src/EmbeddedManifold.jl
```diff
@@ -228,400 +228,183 @@ This is used by the [`AbstractDecoratorManifold`](@ref) within
...
+    get_coordinates!,
+    get_vector,
+    get_vector!,
+    inverse_retract!,
+    mid_point!,
+    project,
+    retract!,
+    vector_transport_along,
+    vector_transport_direction,
+    vector_transport_to,
```
mateuszbaran, line 264: Why, for some functions, does this list have mutating variants but not for other?

mateuszbaran: BTW, making this dispatch statements in a loop is a great idea :slightly_smiling_face: .

kellertuer: I id not change the behaviour here, that is just what we had before.
Note that embed! is done below, as are the vector transports...for the others I am not sure, why we did it that way before.

Thanks – reduced the code by nearly 100 loc.

mateuszbaran:
I still don't quite understand why certain functions are in certain lists but at least it looks much cleaner now :slightly_smiling_face: . It would be great it this could be used to clean up metric and group manifolds a little.

> Ah, one just has to add the type to the MetricManifold super type (maybe introduce a `MetricTransparencyType` and import the abstract type accordingly. For Group Manifold the Embedded Manifold just does the job, but we should refine that then.

Metric transparency has some weird interactions with connected manifolds. Connected manifolds should override the entire metric structure though it still might make sense to have a norm on tangent spaces (but not the inner product). So the connected manifold should invalidate `inner` but not necessarily `norm` -- though I don't know how useful it could be.

src/EmbeddedManifold.jl
mateuszbaran: I'd say that, as long as the tests pass, making this change is a good idea and we can figure out better function lists later.

kellertuer:
> It would be great it this could be used to clean up metric and group manifolds a little.

The MetricManifold needs a little update anyways, and the group manifold will get its own transparency type I think and then I will rework, that anyways.
While it is breaking here, it should be nonbreaking for them, though.

> Metric transparency has some weird interactions with connected manifolds. Connected manifolds should override the entire metric structure though it still might make sense to have a norm on tangent spaces (but not the inner product). So the connected manifold should invalidate `inner` but not necessarily `norm` -- though I don't know how useful it could be.

Do you mean `ConnectionManifold`? Sure, since both metric and connection are a similar way to provide the inner products on manifolds, there is some interaction. But I would recommend to discuss that elsewhere. In short: We can have a ConnectionManifold that uses a metric to introduce the christoffel symbols and such and vice versa I think.


This PR should be finished now (despite bumping to `0.12.0`.

mateuszbaran:
> The MetricManifold needs a little update anyways, and the group manifold will get its own transparency type I think and then I will rework, that anyways.
> While it is breaking here, it should be nonbreaking for them, though.

Nice, though it still would make sense to switch its transparency declarations to loops just like here.

> Do you mean `ConnectionManifold`? Sure, since both metric and connection are a similar way to provide the inner products on manifolds, there is some interaction. But I would recommend to discuss that elsewhere. In short: We can have a ConnectionManifold that uses a metric to introduce the christoffel symbols and such and vice versa I think.

The thing is that connections don't provide inner products while overriding lots of other structure like exp, log or parallel transport.

kellertuer:
> Nice, though it still would make sense to switch its transparency declarations to loops just like here.

Sure, I will do that for them in the rework; but that is also nonbreaking

> 
> > Do you mean `ConnectionManifold`? Sure, since both metric and connection are a similar way to provide the inner products on manifolds, there is some interaction. But I would recommend to discuss that elsewhere. In short: We can have a ConnectionManifold that uses a metric to introduce the christoffel symbols and such and vice versa I think.
> 
> The thing is that connections don't provide inner products while overriding lots of other structure like exp, log or parallel transport.

Ah, right. Well we have to be careful there, but the transparency type will surely help. And we can have the LeviCevitaConnection type where it (again) is non overriding but tranparent to its metric.

mateuszbaran (approved):

PR #77, opened 2021-06-27
Introduce a `copy(M,p)` and a `copy(M, p, X)` command 

kellertuer:
...to easily copy points and tangent vectors (recursively) into newly allocated memory.


kellertuer:
Maybe the patch test was too early since the comment already contradicts this? However, tests cover the new code, we still only have two lines missing due to an macro-function codecov error, maybe the same as https://github.com/JuliaLang/julia/issues/36283

mateuszbaran (approved):
Thanks for adding this :+1: .

PR #79, opened 2021-07-24
Overridable defaults for retract, inverse_retract, and vector transport

Nikdwal:
This adds a dispatch layer setting several default operations, e.g. calling `retract` without a retraction method uses whichever retraction method is returned by `default_retraction_method(M)` instead of using `ExponentialRetraction()` for every manifold. Similarly for inverse retractions and vector transport.
Something similar could be done for bases, but there is no de facto standard basis to my knowledge. Creating a default basis would require additional methods for `get_basis` etc where the basis type is unspecified.


src/retractions.jl
```diff
@@ -143,13 +159,13 @@ Compute the inverse retraction, a cheaper, approximate version of the
 Result is saved to `X`.
 
 Inverse retraction method can be specified by the last argument, defaulting to
-[`LogarithmicInverseRetraction`](@ref). See the documentation of respective manifolds for
+`default_inverse_retraction_method(M)`. See the documentation of respective manifolds for
```
kellertuer, line 162:
```suggestion
[`default_inverse_retraction_method`](@ref)`(M)`. See the documentation of respective manifolds for
```
this way the reference to the function is also linked. Can you adapt that for all occurences?

src/vector_transport.jl
```diff
@@ -252,15 +252,27 @@ struct SchildsLadderTransport{
...
 @doc raw"""
     pole_ladder(
         M,
         p,
         d,
         q,
         c = mid_point(M, p, q);
-        retraction=ExponentialRetraction(),
-        inverse_retraction=LogarithmicInverseRetraction()
+        retraction=default_retraction_method(M),
```
kellertuer, line 274: Merely a comment – while I was not 100% sure for the default with retract (one could jus also dispatch the verity without a retraction method to a new/different default for a manifold) what I really like is how this turned out! Here the new function yields a really nice feature!

kellertuer (commented):
This looks even nicer with the keyword argument effect than I thought. I just have one remark on the docs.

kellertuer:
Concerning the basis, I do not see much of an effect to have just `get_basis` without a type, since they merely increase in features (any basis to orthogonal basis to orthonormal basis etc), so the default would indeed just always be the one with the least features, i.e. `DefaultBasis`.

Nikdwal:
So would you suggest something like `get_default_basis(M)` witch gives a `DefaultBasis`? If so, which methods would it be used in besides `get_basis`?

kellertuer:
No, I would see not much positive features in that. My main reasoning is, that basis types are more like linear increasing in features (an orthonormal basis is also an orthogonal basis just with more features), unlike for retractions or vector transports, where we have more like a unsorted bunch of types. But let‘s wait for more opinions.

Thanks for this contribution it looks like an easy but quite valuable addition.

mateuszbaran:
Thanks, this is a very nice new feature.

Regarding default basis, here it's sort of done in reverse: while for example `ExponentialRetraction` is uniquely defined for every manifold (based on its connection), things like `DefaultBasis` or `DefaultOrthonormalBasis` are intentionally vague. Its meaning is separately defined for each manifold, and the type only encodes some basic properties (like orthonormality). Anyway, maybe we should even turn orthogonality and orthonormality into traits so that there is just `DefaultBasis` with traits that depend on the manifold.

mateuszbaran (approved):

kellertuer (approved):
I think Traits are a good idea, though until now we do not have them in the dependencies here. So this is definetly something for another PR :)

Thanks for solving this so fast. Please also bump the version so we can register a new release directly when merging.

PR #80, opened 2021-07-27
Fix and stabilise gram_schmidt.

kellertuer:
Yesterday I ran into a bug with a student, where entering with an ONB into Gram Schmidt we did not get a full basis back (either an error or a incomplete set).
The problem was, that the old variant was too unstable. The new one checks, whether the obtained vector is linear dependent with any of the previous ones.
I also added that the input (V) is no longer modified, since `gram_schmidt` is without a dash. However - now one of the tests fails where the comment says it checks that projection is idempotent and I do not understand why that test should test that.

### DIFF
--- a/src/bases.jl
+++ b/src/bases.jl
@@ -707,6 +707,24 @@ When a set of vectors is orthonormalized a set of vectors is returned.
 When an [`AbstractBasis`](@ref) is orthonormalized, a [`CachedBasis`](@ref) is returned.
 """
 function gram_schmidt(
+    M::AbstractManifold{𝔽},
+    p,
+    B::AbstractBasis{𝔽};
+    warn_linearly_dependent = false,
+    return_incomplete_set = false,
+    kwargs...,
+) where {𝔽}
+    V = gram_schmidt(
+        M,
+        p,
+        get_vectors(M, p, B);
+        warn_linearly_dependent = warn_linearly_dependent,
+        return_incomplete_set = return_incomplete_set,
+        kwargs...,
+    )
+    return CachedBasis(GramSchmidtOrthonormalBasis(𝔽), V)
+end
+function gram_schmidt(
     M::AbstractManifold,
     p,
     V::AbstractVector;
--- a/src/bases.jl
+++ b/src/bases.jl
@@ -744,20 +744,13 @@ function gram_schmidt(
         end
         nrmΞₙ = norm(M, p, Ξₙ)
         if nrmΞₙ == 0
-            warn_linearly_dependent && @warn "Input vector $(n) has length 0."
+            warn_linearly_dependent &&
+                @warn "Input vector $(n) lies in the span of the previous ones."
             linear_independent = false
         end
         Ξₙ ./= nrmΞₙ
-        for k in 1:length(Ξ)
-            if isapprox(abs(real(inner(M, p, Ξ[k], Ξₙ))), 1; kwargs...)
-                warn_linearly_dependent &&
-                    @warn "Input vector $(n) is not linearly independent of output basis vector $(k)."
-                linear_independent = false
-            end
-            (!linear_independent) && break
-        end
         (!linear_independent || length(Ξ) == dim) && break
-        linear_independent && push!(Ξ, Ξₙ)
+        push!(Ξ, Ξₙ)
     end
     if return_incomplete_set || length(Ξ) == dim
         return Ξ
--- a/src/bases.jl
+++ b/src/bases.jl
@@ -707,24 +707,6 @@ When a set of vectors is orthonormalized a set of vectors is returned.
 When an [`AbstractBasis`](@ref) is orthonormalized, a [`CachedBasis`](@ref) is returned.
 """
 function gram_schmidt(
-    M::AbstractManifold{𝔽},
-    p,
-    B::AbstractBasis{𝔽};
-    warn_linearly_dependent = false,
-    return_incomplete_set = false,
-    kwargs...,
-) where {𝔽}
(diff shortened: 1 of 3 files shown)


src/bases.jl
```diff
@@ -736,29 +736,30 @@ function gram_schmidt(
...
         nrmΞₙ = norm(M, p, Ξₙ)
         if nrmΞₙ == 0
             warn_linearly_dependent && @warn "Input vector $(n) has length 0."
-            @goto skip
+            linear_independent = false
         end
         Ξₙ ./= nrmΞₙ
         for k in 1:length(Ξ)
-            if !isapprox(real(inner(M, p, Ξ[k], Ξₙ)), 0; kwargs...)
+            if isapprox(abs(real(inner(M, p, Ξ[k], Ξₙ))), 1; kwargs...)
```
kellertuer, line 752: I even think that this test is not necessary. If it would be linearly dependent of the others, we would have gotten a zero vector already.

mateuszbaran: I don't understand, why did you change this line? Could you give an example this fixes? This test is primarily aimed at detecting numerical errors. Do you think they are harmless? Naively it seems valuable to catch vectors that aren't exactly 0 (see that the test above is `== 0`, not approximately 0) but still too close to zero to be useful in practice.

kellertuer: So we often lost a vector to this test, which was perfectly fine to an ONB (I can only run into this in the code of my student currently and do not have a separate example), but if the norm of \Xi_n is not zero and one of these tests fires (either mine our yours, I think mine is more stable) – it should really just be numerical errors and I do not see why that should lead to the conclusion that the vector should not be part of the final set.

Also I think that the copy I am now doing (of `V`) since Gram Schmidt is non mutating is causing the three new errors?

mateuszbaran: One of the tests seems to explicitly check this condition. Could you maybe keep this check as an option that can be turned on through a keyword argument?

I totally agree that there should be a copy. The other two tests seem to fail because they assume that no copy is made.

kellertuer: Thanks for checking.

I am ok with only turning this on by keyword, we do that for the return and the warning, too; but I find it strange to exclude vectors from the resulting basis just due to rounding errors?

mateuszbaran: I don't exactly remember the motivation, but since it's already there, we can keep it just in case.

kellertuer: Which one should we keep then?

The new one checks for linear dependence as the warning suggests

The old one – checked whether \Xi_n was not orthogonal to any of the vectors from the basis (but warned about linear dependence)...

mateuszbaran: I'm not sure, does either one make the test pass?

kellertuer: well the previous one did for sure, but then at least the warning was wrong and I am against throwing a vector out for numerical instability. We can also have both, but I do not see much of a use in the (old) orthogonality test if before that one subtracts all parts (or in other words projects on the orthogonal complement) – for sure its orthogonal then (upto numerical precision).

mateuszbaran: OK, so let's just remove this test for now.

kellertuer: Ok. Then I only have to understand what the failing tests were assumed to be testing and adapt or delete them

kellertuer: Hm, so one of the examples in the tests now produces unit vectors where all but two are orthogonal, but these two — that is strange, are not.

kellertuer:
I found the culprit: The test, whether the norm of the vector is zero (after projecting onto the orthogonal complement of the already existing span) was too hard (with `==0`). I introduced a tolerance for that.

Then, some tests wanted to have a continuation after sich a linearly dependent vector was found, so I added a keyword argument for that (`skip_linearly_dependent`). Not skipping them means we error.

Also I added a warning if there are too many vectors and some are ignored in the end (they should then be linearly dependent, but well...)

kellertuer:
addendum: The error from above happened then as follows:
Since the (nearly zero) vector was not exactly zero (after all projection fun), it was normalised to 1 (emphasising the rounding errors tremendously), which made the new vector not orthogonal to the rest. This was what the old tests detected and then again said: Oh then it must be linearly dependent if it is not orthogonal here (though it was that only due to rounding errors and emphasising those).

kellertuer:
Yes code coverage is reduced, since we have 4 lines less now, but it is still only 2 lines that are not covered, as before.

So from my side, this PR is done and fixes, what was wrong.

src/bases.jl
```diff
@@ -696,10 +687,13 @@ The method always returns a basis, i.e. linearly dependent vectors are removed.
 
 * `warn_linearly_dependent` (`false`) – warn if the basis vectors are not linearly
   independent
+* `skip_linearly_dependent` (`false`) – whether to just skip (`true`) linear a vector that
```
mateuszbaran, line 690:
```suggestion
* `skip_linearly_dependent` (`false`) – whether to just skip (`true`) a vector that
```

src/bases.jl
```diff
@@ -728,45 +724,49 @@ function gram_schmidt(
     M::AbstractManifold,
     p,
     V::AbstractVector;
+    atol = 5 * 1e-16,
```
mateuszbaran, line 727: Could we replace this with something like `eps(number_eltype(V[1]))`? So that it also works for `Float32` for example.

kellertuer: Sure. I just missed to do it more generic in the first place. The most recent commit solves this.

mateuszbaran (approved):

PR #81, opened 2021-07-29
Make copy and copyto! transparent.

kellertuer:
This is just a small PR (to see code coverage) – I am introducing copy and copyto! to be transparent functions by default.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "ManifoldsBase"
 uuid = "3362f125-f0bb-47a3-aa74-596ffd7ef2fb"
 authors = ["Seth Axen <seth.axen@gmail.com>", "Mateusz Baran <mateuszbaran89@gmail.com>", "Ronny Bergmann <manopt@ronnybergmann.net>", "Antoine Levitt <antoine.levitt@gmail.com>"]
-version = "0.12.3"
+version = "0.12.4"
 
 [deps]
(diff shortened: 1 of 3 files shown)


mateuszbaran:
Nice. Maybe let's also introduce transparency for vector `copy` and `copyto!`?

kellertuer:
I am not sure what codecov is reporting here, since when I click through their pages to see which lines we missed, it shows again just 2 lines missed not 6.

mateuszbaran:
Coverage has problems with figuring out macro-generated code so it may be a false positive. Are you sure there is actually a call somewhere to `copy` or `copyto!` on a decorator manifold?

kellertuer:
Well the definitions are called, I only used our signature macro to introduce those definitions as transparent.

mateuszbaran:
Still, it'd be nice to have a test that fails without these definitions if it's not too difficult.

kellertuer:
Sure they should not be that difficult, I will write some later.

kellertuer:
Hm, I checked and we do not explicit transparency checks for exp/log/... either, we just check that transparency works as intended (In the decorator tests).

mateuszbaran (approved):
Hm, OK, that coverage thing is quite weird here.

kellertuer:
I still have to check for embedded manifolds, It should not be transparent for all of them (fixed rank fails currently).

PR #85, opened 2021-08-24
Make `number_of_coordinates` transparent

mateuszbaran:
More properly handles a part of https://github.com/JuliaManifolds/Manifolds.jl/pull/416 .


kellertuer (approved):
Since I was shortly confused why this is not just the manifold dimension, maybe it would be good to add a sentence in the docs that it is _usually_ the manifold dimension if the manifold and the basis have the same number field?

mateuszbaran:
OK, I've added a sentence to docs. I've restricted it to real manifolds and bases because it's not necessarily true in other cases.

PR #89, opened 2021-10-30
Introduce fallbacks for AbstractManifoldPoints and Vectors

kellertuer:
While working on a second type for the Grassmann manifold, I endet up noticing that I would reimplement quite some stuff we did for Hyperbolic already, especially including the broadcast.
Also for a default type (for the Stiefel representation) on would reimplement all exp/log/...
so I thought one could also do this pass-through to the -value field by default such that a new point/tangent type has only to reimplement the stuff that is new/different.

This is currently done after all other stuff (bases / defaults) is done, i.e. on the lowest levels

I have not yet finished
* [x] retractions / vector transports are missing
* [x] tests are missing

but this would reduce quite some code in manifolds and omit quite some code I saw I would have to write next in Grassmann. What do you think?


mateuszbaran:
I think I'd prefer a more granular approach with macros, like here for example:
```julia
macro manifold_point_forwards(T::Symbol, field::Symbol)
    return esc(quote
        allocate(p::$T) = $T(allocate(p.$field))
        allocate(p::$T, ::Type{P}) where {P} = $T(allocate(p.$field, P))
        function allocate(p::$T, ::Type{P}, dims::Tuple) where {P}
            return $T(allocate(p.$field, P, dims))
        end

        number_eltype(p::T) = typeof(one(eltype(p.$field)))
    end)
end
```
and so on for other methods.

Advantages:
1. Not all manifold points have to follow our defaults.
2. It's still easy to make a new type follow these defaults (basically writing `@eval @manifold_point_forwards MyType value`)
3. We can name the field differently in different types.
4. We can be more granular, for example while forwarding methods to the wrapped value is sometimes useful, it seems to me like a dangerous default behaviour, which in the case could be just written in a second macro and we could have different behaviours for different types.

kellertuer:
Sure, that also sounds good. Just that I do not feel expiereinced enough to program that and usually macros are also quite complex. Can you mean help? And should we do that here or first over at manifolds?

mateuszbaran:
Sure, I can write these macros. We can put some of them here, at least the part that would be useful to Validation manifold.

mateuszbaran:
This should, more or less, work now. Macro `@default_manifold_fallbacks` should probably be in Manifolds.jl for now, I don't see much use for it here.

kellertuer:
Thanks for the fast help; I think the main question for the fallbacks macro is: Is it useful as a tool that should stay here?

It is definite useful as a tool, so maybe it can stay here, since it would feel too heavy to load all of manifolds to just use it?

mateuszbaran:
OK, it can also stay here. Let me know it you need some more help with this PR.

kellertuer:
Great. I will check that the test coverage is given here while also using the macros already over at manifolds for the new type. Thanks for the help.

src/point-vector-fallbacks.jl
```diff
@@ -0,0 +1,194 @@
...
+
+List of forwarded functions:
+* [`allocate`](@ref),
+* [`copy`](@ref),
+* [`copyto!`](@ref),
+* [`number_eltype`](@ref) (only for values, not the type itself),
+* `similar`,
+* `==`.
+"""
+macro manifold_thing_forwards(T, Twhere, field::Symbol)
```
kellertuer, line 16: I will probably drop the `_thing` here. `manifold_fallbacks` sounds fair enough, “thing” is too unspecific, if we need a third word I would opt for `_element`. And are the docs right that `T` is a symbol? I would read the code as it being a type?

mateuszbaran: Yes, naming could be improved. One thing is that I don't want this and `default_manifold_fallbacks` to be confused, so maybe `manifold_element_forwards` would be better.

`T` isn't necessarily a symbol; it was in the initial version of this macro but it turned out to be not expressive enough and I had to allow for more complicated `T`.

kellertuer: I think `manifold_element_forwards` is distinguishable enough.

Ok I'll adapt the docs as well here, then.

kellertuer: I somehow did not find tie last month, and now coming back I noticed that the `manifold_element_forwards function is documented with two arguments but actually has 3. The middle one is not documented.

Sure I could try to reverse engineer from its usage … but maybe I would be off with that – so what is it intended to be used for?

kellertuer: Ah, maybe the supertype for `T`? Would it be useful to introduce a 2-parameter macro for the `_` case?

kellertuer:
This looks good so far and I am now trying to use it with Hyperbolic first and then for the actual idea on ProjectionPoints in Grassmann – we just also have to increase test coverage here on the newly introduced functions that are now automatically generated.

kellertuer:
I sketched the tests, but it seems the automatically introduced `log!` does not yet work.

mateuszbaran:
I'm just looking at it and it doesn't even reach `log` for me. I'll push a patch soon.

kellertuer:
In my error it looks like It reaches `log` then allocates a `DefaultPoint` (but not a `DefaultVector`) and then dispatches to the wrong `log!` of course (since the macro one is with a vector ;) ) - and the wrong log! for sure can not do broadcast-thingies on the DefaultPoint. Here's the two interesting lines from the error stack (not the wrong type of `Y`).

```
    [6] log!(#unused#::DefaultManifold{Tuple{3}, ℝ}, Y::DefaultPoint{Vector{DefaultTVector}}, p::DefaultPoint{Vector{Float64}}, q::DefaultPoint{Vector{Float64}})
      @ ManifoldsBase ~/Repositories/Julia/ManifoldsBase.jl/src/DefaultManifold.jl:94
    [7] log(M::DefaultManifold{Tuple{3}, ℝ}, p::DefaultPoint{Vector{Float64}}, q::DefaultPoint{Vector{Float64}})
      @ ManifoldsBase ~/Repositories/Julia/ManifoldsBase.jl/src/exp_log_geo.jl:73
```

kellertuer:
As mentioned above we could also add retractions and inverse retractions and the vector transport to that macro, if that does not destroy the rest of the world (maybe it does ;) ).

mateuszbaran:
Let's fix the current code first... I've just noticed that I have forgotten to pull new changes which caused the problem of not reaching `log`. Now I'm fighting with the `log` problem itself.

mateuszbaran:
I've fixed a bunch of cases. Currently the one that fails is `zero_vector` but I'm not sure what the right solution is in this case.

kellertuer:
I would say similar to the `log!`– allocate the corresponding `$TV` and define `zero_vector!`.

mateuszbaran:
Maybe it would make more sense to unwrap, call `zero_vector` on the unwrapped point and wrap the result?

kellertuer:
Oh, that sounds really reasonable as well!

mateuszbaran:
I've fixed that vector problem and some more. Now the problem is with `embed!` and that `MatrixVectorTransport` thing which I forgot what it actually does. Do you have some ideas about these?

kellertuer:
The MatrixVectorTransport just models a curve – creating it correctly for the DefaultPoint solved that point.
I also fixed `embed!` and introduced the vector transports, retraction and inverse retraction (and resolved their ambiguities with exp/log). Now (hopefully) just the vector transport ambiguities might be missing.

kellertuer:
Maybe we can resolve the vector transport ambiguities similar to how you resolved the vector/coordinate functions?

mateuszbaran:
I think so, I'll try doing that today.

kellertuer:
I had a moment and extended test coverage a little bit, but
* `retract!`
* `inverse_retract!`
* `vector_transport_to!`
* `vector_transport_direction!`
* a few lines in the broadcast

are not covered yet. For the first 4 I ran into Ambiguities again, I left the two for retract and inverse retract in (as broken tests) but I have no good idea yet how to resolve them.

mateuszbaran:
`retract!` and `inverse_retract!` ambiguities would be solved by the two-level scheme since `DefaultPoint` unwrapping would be at the top level and actual implementation at the lower level. Currently there is no other way than manually resolving such ambiguities.

kellertuer:
I see, maybe than it is not worth having these in the macro?

mateuszbaran:
Yes, I've removed them.

kellertuer:
Oh noes we increased the not-covered lines by 50% (2-3) – we could check for those maybe, but in general I think this PR should work for now :)

mateuszbaran:
The new non-covered line is, I think, a false positive, there is nothing we can do about it.

mateuszbaran:
I'll review it once again sometime soon but this indeed seems to be complete.

kellertuer:
When I find time (during exam/corrections time and preparing for the new semester starting January 10) I will apply these to the Projector-Grassmann-branch I started with before this detour ;)

src/ManifoldsBase.jl
```diff
@@ -569,6 +570,7 @@ include("vector_transport.jl")
 include("DecoratorManifold.jl")
 include("bases.jl")
 include("vector_spaces.jl")
+include("point-vector-fallbacks.jl")
```
mateuszbaran, line 573: Maybe we should call it `point_vector_fallbacks.jl` for consistency?

mateuszbaran: Other than this I think this PR is ready for merging.

mateuszbaran (approved):

PR #90, opened 2021-12-10
adds one missing allocation function

kellertuer:
This PR adds a small allocation forward when the dimension is also specified.

### DIFF
--- a/src/point_vector_fallbacks.jl
+++ b/src/point_vector_fallbacks.jl
@@ -28,7 +28,7 @@ macro manifold_element_forwards(T, Twhere, field::Symbol)
                 return $T(allocate(p.$field, P))
             end
             function allocate(p::$T, ::Type{P}, dims::Tuple) where {P,$Twhere}
-                return $T(allocate(p.value, P, dims))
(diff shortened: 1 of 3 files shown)


src/point_vector_fallbacks.jl
```diff
@@ -27,6 +27,9 @@ macro manifold_element_forwards(T, Twhere, field::Symbol)
             function ManifoldsBase.allocate(p::$T, ::Type{P}) where {P,$Twhere}
                 return $T(allocate(p.$field, P))
             end
+            function allocate(p::$T, ::Type{P}, dims::Tuple) where {P,$Twhere}
+                return $T(allocate(p.value, P, dims))
```
mateuszbaran, line 31:
```suggestion
                return $T(allocate(p.$field, P, dims))
```
It looks fine with this small change.

kellertuer: Thanks for noticing, this is indeed a copy-paste-error. And I will check for a test to cover these as well, for sure.

mateuszbaran (approved):

PR #91, opened 2021-12-15
Refactor ManifoldsBase to a two-level dispatch design

kellertuer:
This is a start to tackle #88.

I sketched the idea in the readme but there is a few points open to discuss

* [x] I would like to move the documentation here, since the interface page is quite long and subpages might help the overview and also the explanation of design principles.
* [x] Maybe we could also zenodo this one?
* [x] We should discuss what to do with the current decorator scheme. I feel it is a little too complex. Yes one can change the dispatch by redefining the dispatch functions, but maybe macros that implement a set of functions by calling the ones on another manifold would to the trick as well? I will sketch this in the comings days/weeks
I think this might be my project for the last week of this year, if time permits. Let‘s also discuss a few details here, for sure.

What needs to be done
* [x] actually implement the lower layer - and maybe also move more Retraction Types and vector transports to base to provide their lower layer fallbacks on the interface level already. This includes
* [x] retractions and inverse retractions
* [x] vector transports
* [x] bases related functions
* [x] For user experience most vector transport types and retraction types should be moved to base to introduce them on layer 3 as well, for example the `CaleyRetraction` or other vector tranports
* [x] We can introduce a `EmbeddedRetraction` that performs a certain retraction in the embedding and projects the result (this generalises the projectionretration, which would be the same with exp).
* [x] check documentation / update documentation.
* [x] get test coverage back
* [x] should we use a separate logo for ManifoldsBase? see [this comment](https://github.com/JuliaManifolds/ManifoldsBase.jl/pull/91#issuecomment-1007954416) for the logo proposal.

PR #92, opened 2022-02-03
Refactor ManifoldsBase to a two-level dispatch design (v2)

mateuszbaran:
I've accidentally closed the previous PR: #91.


kellertuer:
We might have to check coverage again (we missed to introduce transparency for infectivity radius), I also found a possible stack overflow error in the default for this function and fixed that – and finished two more manifolds in Manifolds.jl by this.

kellertuer:
Update – SPDs work fine (after fixing a transparency for metrics, to let point/vector checks just pass through).

mateuszbaran:
I can take a look at tangent bundle but probably not this weekend.

kellertuer:
I currently can not get `Oblique` back up running, though we might have missed to (ever) set `power_dimension`, now the write/read tricks won't work – and I do not see why (either why that should have ever worked or why it fails now). I only manage to keep constantly 3 test failing, nearly independent of what I try for Oblique in the AbstractPowerManifold.

mateuszbaran:
I've fixed Oblique, turns some replacing and non-replacing methods were mixed up a bit.

kellertuer:
I introduced check_size for ProductManifolds, but then (only 10 errors left in there) I was a little unsure, we should introduce vector_transport_to (for both representation) and parallel_transport_to/to!/direction/direction!/

but I was unsure whether to do that on level 1 (I would have expected that) or level 2 (the _ ones, we currently only very rarely do that), since currently for retract it is done on that level. If possible I would prefer Level 1.

kellertuer:
I just added parallel_transport_to/direction to Power Manifold and it seems to miss check_size (for points and vectors) as well – I currently have 2 errors on the Manopt branch for that, will add these soon then, too.

mateuszbaran:
> I was a little unsure, we should introduce vector_transport_to (for both representation) and parallel_transport_to/to!/direction/direction!/

I'm also a bit unsure but probably level 1 should be fine.

src/ManifoldsBase.jl
```diff
@@ -356,16 +356,11 @@ Distance ``d`` such that
 is injective for all tangent vectors shorter than ``d`` (i.e. has an inverse) for point `p`
 if provided or all manifold points otherwise.
 """
-function injectivity_radius(M::AbstractManifold)
-    return error(manifold_function_not_implemented_message(M, injectivity_radius))
-end
+injectivity_radius(M::AbstractManifold)
```
mateuszbaran, line 359: Shouldn't we split `injectivity_radius` on the second level similartly to `retract`? That is have `injectivity_radius_exponential`, `injectivity_radius_qr` etc.? This would save us some ambiguities.

kellertuer: Yes, I thought that as well when I saw that - it would fit our deign better.

kellertuer:
> > I was a little unsure, we should introduce vector_transport_to (for both representation) and parallel_transport_to/to!/direction/direction!/
> 
> I'm also a bit unsure but probably level 1 should be fine.

Great then we should move retract up as well. My Manopt tests are also back up and running and product is fine I think. We come close to just having Lie groups left.

mateuszbaran:
Yes, I managed to (mostly) fix ProductManifold without any magic :slightly_smiling_face: .

kellertuer:
I just tried to fix one of the remaining product manifold things but get_vector gets ambiguous, since we have the cached one as well on Product manifolds. So maybe we have to move them to layer 3 to dispatch per basis type?

kellertuer:
I think I have an idea to dispatch internally on how to pass the basis on, so we do not have to dispatch per basis on the layer 1. I will try that the next days.

kellertuer:
The initial problem was that `get_vector` was missing for product manifolds (we did `get_coordinates` already - so the non-mutating one) and currently those both dispatch in bases to fetch the special case of the CachedBasis in the `map` to the components. My idea is to write an internal `basis_parts` function that does the actual dispatch (i.e. return a `Ref(B)` bu default and the cached parts for the cached basis). that way we also only have the `AbstractBasis` function for product manifolds (and a dispatch internally)

I was thinking about doing something similar to the construction of the result as productRepr or Array... since currently also both are implemented twice in several places to construct the right return value.

mateuszbaran:
Yes, perhaps an internal dispatch would be nice. For now I've just duplicated `get_vector`, I'll push the change in a moment.

kellertuer:
Hmm, interesting I have not yet checked your commit on Manifolds, but my approach to that had a lot of ambiguities, but if so I will just do what I described above.

mateuszbaran:
I've started a bit on the group trait. The first thing I'm not sure how to solve is transparency of non-group related operations through `GroupManifold` struct. Do we have better way than defining `get_embedding(G::GroupManifold) = get_embedding(G.manifold)` and so on for each function?

kellertuer:
I would call that `base_manifold`, since it has nothing to do with an embedding, and use an `IsGroup()` trait that provides the fallback?

mateuszbaran:
Hm, I wrote about `get_embedding` because that's the method that was the failure in one of the tests. And it does make sense if a manifold has both embedded and group traits. Let's maybe define that for `IsGroup` trait, `HasBase` seems a bit too unspecific.

kellertuer:
I think that was, seeing from now, not a good decision back then. The reason was a mix between the decorator not being completely designed yet, complicated to set uo new things and the isometric embedding idea working well for the group as well concerning the passing on of functions. Now that all this is easier I would prefer a name that fits more the case we have.

maybe `IsGroup` is good but the `get_base_manifold`might be misleading, so we could take `get_riemannian_manifold` (i.e. the manifold “behind the group”)?

mateuszbaran:
I'm not particularly attached to `base_manifold` so feel free to change it :slightly_smiling_face: . `IsGroup` seems fine, except I don't quite like that there is no way to easily forward all metric-related functions. Maybe it's not particularly important though because I think group structure is the only thing that passes everything so it's not clear whether working out a more general solution is worth it. Anyway, feel free to experiment with it, I'm going to slowly work on groups when I have a free moment.

mateuszbaran:
There is one nontrivial issues with `GroupManifold` currently. This:
```julia
function retract(
    M::AbstractManifold,
    p,
    X,
    t::Real,
    m::AbstractRetractionMethod = default_retraction_method(M),
)
    return _retract(M, p, t * X, m)
end
```
goes from level 1 to level 2 while we currently only define forwards for decorators on the outermost level. So I don't know whether this shouldn't go down to level 2, we should replicate this functionality in group manifold forwards or maybe something else?

kellertuer:
I would have thought we stay on level 1 and this case is done _before_ the decorator kicks in? How is it done for embedded?

mateuszbaran:
I don't think it's actually done for embedded but staying on level 1 makes sense.

kellertuer:
Ok, yes I think this should stay on level one and just finish the default case for the t.

kellertuer:
Interestingly I have zero errors on groups? I just have a few on `PowerManifold` that I do not understand and a few on MetricManifold that I also do not yet completely understand. Have you fixed all Groups already?

kellertuer:
I tried working on the MetricManifold, but its interplay to the COnnectionManifold is not so clear to me, i.e. how they would interact. 

Current Problem – When loading OrdinaryDiffEq the Metricmanfold tests whether the Ode can be solved. In the new system, the dispatch over to the right retract in COnnectionManifold is los and I do not see how to reintroduce this. Can you take a look?

kellertuer:
I just wanted to check the first group and noticed, we might need more traits.

* `HasLeftInvariantMetric`
* `HasRightInvariantMetric`
* `HasBiInvariantMetric`

or something. What do you think? Otherwise I am a little stuck on how those decorator traits should be modelled now.

mateuszbaran:
> I tried working on the MetricManifold, but its interplay to the COnnectionManifold is not so clear to me, i.e. how they would interact.

I think they shouldn't really interact, that is either one overwrites the wrapped metric/connection structure.

> Current Problem – When loading OrdinaryDiffEq the Metricmanfold tests whether the Ode can be solved. In the new system, the dispatch over to the right retract in COnnectionManifold is los and I do not see how to reintroduce this. Can you take a look?

OK, I'll check that.



> I just wanted to check the first group and noticed, we might need more traits.
> 
> * `HasLeftInvariantMetric`
> * `HasRightInvariantMetric`
> * `HasBiInvariantMetric`
> 
> or something. What do you think? Otherwise I am a little stuck on how those decorator traits should be modelled now.

Yes, these sound like useful traits.

kellertuer:
For today I at least got the errors in Circle down from 44 to 1 failed 3 error, because is_pint/vector were missing in group forwards :)

kellertuer:
Ah, and the final errors seem to be mainly the following problem:

`inverse_retract` is currently transparent for groups and (for CircleGroup) passed onto Circle. But the `GroupLogarithmicInverseRetraction` it should not be. Similarly for the GroupExp of course.

mateuszbaran:
> > Current Problem – When loading OrdinaryDiffEq the Metricmanfold tests whether the Ode can be solved. In the new system, the dispatch over to the right retract in COnnectionManifold is los and I do not see how to reintroduce this. Can you take a look?
> 
> OK, I'll check that.

Could you point me to the problematic test? There seem to be many complicated issues with invariant metrics and I don't quite remember how it was supposed to work.

kellertuer:
> Could you point me to the problematic test? There seem to be many complicated issues with invariant metrics and I don't quite remember how it was supposed to work.

Just the first error in running `test/metric.jl`(line 228)

kellertuer:
So Circlegroup was my goal for today. The Traits are not yet added, but I know what I want to do there) - without those tests all of CircleGroup works.

* I had to do inverse_retract the same way you already did retract
* I had to introduce is_point and is_vector for group manifolds to pass through as well as representation size.

Will try the Metric (3 Traits) invariants tomorrow then.

mateuszbaran:
One major problem is that `parent_trait` isn't imported into Manifolds.jl.

mateuszbaran:
OK, I've added it. Now the remaining errors look quite ordinary.

kellertuer:
Cool. Yeah. One of them is that infectivity radius is not yet a trait functions and has only one level which causes ambiguities.

kellertuer:
Hm, I'm confused and should maybe leave it be for today. I introduced `is_point` _just_ for the Group trait and now all other ìs_points` run into stack overflows, I have no clue why.

kellertuer:
I did the traits (with `IsLeftInvairantMetric` and right and bi – with a common abstract trait super type) the one thing that does change is:
Instead of extending/decorating a metric (`LeftInvariantMetric(M)`) we now specify that a certain GroupManifold (manifold + metric) is left invariant or right or bi in variant (which I think is nicer). I just have to check how much the functions now change in the dispatch. Most will just be `TraitList`s wich the new Traits.

The nice thing is, specifying a `LeftInvariantMetric` for a manifold makes it a GroupManifold as well (automatically thanks to parent_trait). This indeed means that in the new scheme we can just dispatch per Traitlist and do not need the `default_invariance_dispatch´ functions anymore, since for example `is_invariant_metric` is now something obtained from the trait that we use for dispatch anyways (and not the other way around).

I just have to go through the code adapting this now.

mateuszbaran:
That sounds nice, so far the new trait system seems to work quite well.

kellertuer:
I have to check a few small functions still like the old `direction` and such which heavily dependent on the decorated metric thingy, but I think, semantically a metric can only be invariant with respect to a group, so we might change signatures there.

In general I agree, it makes all these things more unified, which I like.

I also changed the `base_manifold` to something we already have, namely the `decorated_manifold` so also that is more unified now.

mateuszbaran:
What's the issue with `direction`? I don't really see anything that would have to change there.

The metric can only be invariant with respect to a group, yes. So far the group is implicitly assumed.

kellertuer:
With the new system I think we make `LeftinvariantMetric(::Metric)` obsolete by shifting this information to a trait (also: explicitly coupling the metric and the group to something invariant).
this makes the LeftInvariantMetric type obsolete, but also requires direction to know the manifold.

If we would keep the `LectInfariantMetric` type we would encode the information on invariance double (which can only lead to the disadvantage of having inconsistencies).

mateuszbaran:
Hm, I'm not quite sure. How would you change the metric to a left-invariant metric on a group manifold without the `LeftInvariantMetric` type? Metric change is done by wrapping in a `MetricManifold` decorator which needs to know the type of new metric. Only then the information is passed onto traits.

kellertuer:
Can you clarify from what to what I would change? I do follow you there?
My fear in keeping that type is: What do you do with a group manifold that `IsLeftInvariantmetric` but the metric is marked as a `RightInvariantMetric(metric)` ?

mateuszbaran:
Let's maybe have an example: `MetricManifoldWithRightInvariantMetric ( GroupManifold (SomeWrappeManifold)))`
In this example `GroupManifold` may have an `IsLeftInvariantmetric` trait but it shouldn't change how metric-related functions are calculated. It's just an indication that the metric of `SomeWrappeManifold` is left-invariant wrt. `GroupManifold`. If we consider `MetricManifoldWithRightInvariantMetric` then it needs to introduce a trait that actually overrides computation of metric-related functions using group structure of `GroupManifold` and metric of `SomeWrappeManifold` at group identity.

kellertuer:
But this would be attached to the `MetricManifoldWithRightInvariantMetric` and not be a decorator of `MyMetric` you use therein. Currently it is attached to the metric, which I think is too short (since it implicitly assumes a group and I feel that is the wrong way around). If we would assume something implicitly we should do that with a (default) metric.

mateuszbaran:
Right, I see now -- you are thinking about what `LeftInvariantMetric` should store inside. I don't have any preferences for that, so pick whatever feels right.

kellertuer:
I think it is obsolete if we have the trait to store the information about invariance (of a metric w.r.t. a manifold).

mateuszbaran:
Well, if `LeftInvariantMetric` is obsolete then what do you specify as a metric for `MetricManifoldWithRightInvariantMetric`?

mateuszbaran:
Or, from the other side, where would you place the functionality of this method: https://github.com/JuliaManifolds/Manifolds.jl/blob/master/src/groups/metric.jl#L188-L195 ?

kellertuer:
that manifold would have the `IsLeftInvariantMetric` trait (maybe `Has` fits better than `Is`.

mateuszbaran:
OK, so there would be no trait for indicating the the metric of the wrapped manifold is invariant wrt group structure (without changing the computation of metric-related functions)?

mateuszbaran:
And `GroupManifold` could change metric without a `MetricManifold` wrapper?

mateuszbaran:
So from what I gather the idea is that `GroupManifold` determines whether metric-related functions are just passed to the wrapped manifold, calculated using left-invariant formulas or using right-invariant formulas. Is that right?

kellertuer:
Yes, compared to before where the metric implicitly assumed (by its type) that it is left invariant in the manifold/group it is in and decided based on that what to do.

The only disadvantage / remodelling here is that in a few places “just the metric” as a parameter is not enough.

mateuszbaran:
OK, I'll take a look at implementation details later but it sounds reasonable.

kellertuer:
thanks. The redesign is not yet completely done, I'll have to continue the next days, but if you see what I mean and how it should be done, feel free to continue.

kellertuer:
I am not yet 100% done with the remodelling but it feels more concise. One thing that was not covered before: Which `direction` to return for a `biinvariant` metric. I have to think about that. For the rest – most code got simpler, since one just passes the manifold into `direction` instead of first getting its metric and passing that.

mateuszbaran:
> One thing that was not covered before: Which `direction` to return for a `biinvariant` metric. I have to think about that.

Bi-invariant trait could imply both left-invariant and right-invariant (as parent traits). If there is no implementation specific to bi-invariant metric then we can fall back to either left- or right-invariant implementation, probably in that order because in Manifolds.jl we generally prefer left-invariance.

kellertuer:
Sounds good, I just have to check whether we had multiple parents as a feature ;) Sure it then should prioritise left.

mateuszbaran:
Yes, we do have multiple parents :slightly_smiling_face: . `parent_trait` can return a `TraitList` with multiple elements. Our trait system currently can do basically everything except matching on something more complex than a single trait, and I don't consider it a limitation that can't be removed. We just haven't needed anything more complex so far.

kellertuer:
Ah, sure, I will adapt that, that sounds great!

kellertuer:
Hm, somehow I still miss something on direction – it feels like previously you could call it with anything and it returned a direction (metric? operation? manifold? somehow all of them, but definetly also with the GroupRetraction), so I am not yet sure how to model this in the decorator scheme (see errors the test/group/metric.jl ) and I am a little stuck how to solve this.

mateuszbaran:
OK, I'll look at it today or tomorrow.

mateuszbaran:
I've fixed `direction` but I don't yet see how to proceed with fixing further errors so I've left them :slightly_smiling_face: . I guess `MetricManifold` should be completely removed from src/groups/metric.jl?

kellertuer:
Ah, and it was even me missing the point with the abstract group, which I think we should remove (its a trait now).

Then the metric manifold (explicit metric) can be a group manifold still and easy (implicit/trait) - I will check and hope we do not have to remove that. Thanks for the help.

kellertuer:
I just noticed that we still have to remove `AbstractGroupManifold`(its `operator` is now in `IsGroupManifold`. That is not too difficult just quite some changes to the code.

mateuszbaran:
Yes, it needs to be removed.

kellertuer:
I am nearly done working through, just with `exp_lie` and `log_lie` I was thinking of documenting (and passing through) `GroupExponential()` and `GroupLogarithm()` in order to also easier resolve the identity cases?

mateuszbaran:
I've managed to fix the power manifold, and I've added some tests in ManifoldsBase to prevent this kind of issue happening in the future (`get_coordinates` was broken for multi-index power manifolds).

kellertuer:
I fixed most bugs after the rework - there is just one identity case (appears in 3 tests) where I am stuck. I either get that the call is ambiguous or that the function definition is double - for the case 
(groups-general lines 222 and 230).

But then only the specific groups are missing.

mateuszbaran:
Are you still stuck with something? I've done a small fix to product manifold and merged master into that branch in Manifolds.jl.

kellertuer:
No the general group tests do work. I noticed that some metric tests are also failing but I will first work through the other groups next.

mateuszbaran:
Yes, introducing layers for `local_metric` sounds like a good idea.

kellertuer:
I finished infectivity_radius … it makes the macro we used there obsolete, it might just be that one has to define 4 radii (M, M&p, M&p&m, M&m (m=retraction, p=point). See Grassmann. The invoke thingy still works I think, but the new code is simpler.

kellertuer:
I tried working on local_metric, but I somehow can not get that to work. Luckily this is the last error on MetricManifold it seems. Somehow the dispatch or the definition of the local metrics in the tests - or even both – do not yet fit the new way we do things (i.e. the local metric defined in the tests are never called).

mateuszbaran:
I will try to fix it tomorrow :slightly_smiling_face: .

kellertuer:
Perfect, my guess is, it has to do with default metric and its passthrough properties.
I finished the two symplectic manifolds so it seems there is just groups left to check now.

mateuszbaran:
I don't understand this:
```julia
function exp(::TraitList{<:HasBiinvariantMetric}, M::MetricManifold, p, X)
    return retract(base_group(M), p, X, GroupExponentialRetraction(direction(M)))
end
function exp!(::TraitList{<:HasBiinvariantMetric}, M::MetricManifold, q, p, X)
    return retract!(base_group(M), q, p, X, GroupExponentialRetraction(direction(M)))
end
```
If a metric is biinvariant, I think we should just forward `exp` to the underlying manifold. `base_group` is currently only defined for group actions. There seem to be no methods for left-/right- but not bi-invariant metrics.

kellertuer:
base group was quite a strange thing (I am trying to figure out what the old code meant there as well) in the old code and we had this forward as well. I think using exp/log for the group exp/log on group manifolds is reasonable?

mateuszbaran:
The old code looks a bit weird. 

> I think using exp/log for the group exp/log on group manifolds is reasonable?

I think both should give the same answer but IMO the code for the wrapped manifold has a higher chance of being more optimized.

mateuszbaran:
I've managed to fix group-metric interaction tests.

mateuszbaran:
One small problem:
```julia
function active_traits(f, M::MetricManifold, args...)
    return merge_traits(
        IsMetricManifold(),
        is_default_metric(M.manifold, M.metric) ? IsDefaultMetric(M.metric) : EmptyTrait(),
        active_traits(f, M.manifold, args...),
        IsExplicitDecorator(),
    )
end
```
When `f` is a metric-dependent function like `christoffel_symbols_second`, `IsExplicitDecorator` is definitely not a correct trait. How would you solve this issue?

kellertuer:
Ah we could do a specific definition for those?
I was just activating that now (again) since that way all Group stuff just “passes through” metric (or Validation for example as well).

Thanks for fixing the metric, I am now on Validation&Group to work together nicely (some challenges with Identity still).

mateuszbaran:
> Ah we could do a specific definition for those?

Either that or not returning `IsExplicitDecorator` for such functions.

kellertuer:
I feel both are reasonable and do not have a preference.

mateuszbaran:
OK, then I'll conditionally enable `IsExplicitDecorator`.

kellertuer:
I slowly feel a little bit like turning in circles. Spent an hour on 3 errors in Validation&Group only for now to get back to Circle group (which ran before just fine) having now 42 errors.
Maybe I either have not yet understood something, or I am not able to actually reduce errors (Hyperbolic & SPDs are also erroring again for example and they were fine a while back).
For worst of cases our system is not yet as stable as we thought.

mateuszbaran:
I will take a look. I've managed to fix some tests for `MetricManifold`.

kellertuer:
I hope that (up to the manifolds that now break again) it is “just” the explicit group manifolds left

mateuszbaran:
`Hyperbolic` worked for me, and the fix for SPDs was very easy. The issue was the `IsExplicitDecorator` on `MetricManifold`. It probably still activates for too many functions because it is still there by default, so feel free to add new functions to `is_metric_function`.

kellertuer:
Yes I am also unsure why Hyperbolic failed in between. Thanks for fixing SPDs.

mateuszbaran:
One small tip: a good way to see how dispatch is going is by using Cthulhu's `@descend` command with optimizations turned off (that's how I discovered the issue with SPD).

kellertuer:
I will take a look at that one :)

mateuszbaran:
Using Debugger also works so it's also an option. VSCode debugger has issues with debugging code defined in REPL, which is not an issue for Cthulhu (though Cthulhu won't tell you much for type-unstable code).

mateuszbaran:
I've nearly fixed all tests for MetricManifold. I've left a couple of issues I wasn't sure about.

mateuszbaran:
Hm, I didn't actually remember that -- anyway, we could have a `Torus` group for that. What `CircleGroup` is supposed to work on, though, are one-element arrays.

kellertuer:
Torusgroup is another good idea. 

the implicit power manifold is the map in 

https://github.com/JuliaManifolds/Manifolds.jl/blob/419285ce6749e0087fffbd22a1c4fbed66beb35d/src/groups/circle_group.jl#L21-L25

mateuszbaran:
I think it is just a confusingly implemented support for one-element vectors.

kellertuer:
> I think it is just a confusingly implemented support for one-element vectors.

I think so, too, I just did not recognise it on Saturday and was annoyed that I could not get the tests to pass. Now I saw what it was and how to fix it (and make it more generic again). TorusGroup that might work on Matrices and Vectors seems to be a good idea (maybe with a PowerGroup behind even).

mateuszbaran:
Cool, I will try to find some time this week to help with these issues.

kellertuer:
Found (and fixed) the error with the approx inverse retraction.

kellertuer:
translation and product work fine now – the main thing missing (for SE/SL/SO) is to figure out precedence between group and embedding, this does not yet completely work as before.

mateuszbaran:
I'll take a look at `Circle` then.

mateuszbaran:
The fix for `CircleGroup` was easy but not obvious at all.

kellertuer:
Wow! I was thinking it is some problem with the types not being able to allocate - but great that the fix is even possible that generally.

mateuszbaran:
`invoke` in `IsExplicitDecorator` had incorrect types, and I'm half-convinced the same change needs to be done for `EmptyTrait`.

kellertuer:
Cool, I was already wondering about that because some did not hit - thanks for fixing those (you are far more expierienced on that side for sure).

kellertuer:
Checked the remaining groups:
`has_approx_biinvariant_metric` is still broken – the rework broke one case in SO(3) where the identity is not turned into an array anymore before going into the ODE problem (when running metric). And the SE(n) is missing (I will do that next). So there is at least not that much left in groups.

kellertuer:
Uff. I actually hit a modelling problem for SpecialEuclidean.

It is modelled as a `SemidirectProductGroup` which is internally storing a `ProductManifold` – ... now for `is_point`/`is_vector` the group says: Not my job! and passes them through – but the points themselves are not ProductRepr/ArrayPartitions but stored in Homogeneous coordinates, so that fails.

I think what we are in fact doing is something like an embedding that changes representation (into a rotation + a translation) where it would be fine, but is not yet included in the model we are doing.

Any idea how to best resolve this? I feel the current types are inconsistent, saying it is a productManifold when it actually is not (unless we do something crazy and specify subsets of an array one can state for product elements).
Now I am unsure where to “hook into” our decorators and resolve this without reimplementing too much.

kellertuer:
Ah! I had an idea – it is another representation for the product manifold (not only the group) so I implemented the checks a “level further down” on the manifold without the group decorator.

kellertuer:
While the remaining small things in Manifolds are not yet resolved (and are tricky) – it seems this PR here is nearly done, so I already did code coverage (even a line less than before, just 2 false nonhits) – so that we only should check the docs again, but this should be fine now.

mateuszbaran:
> Ah! I had an idea – it is another representation for the product manifold (not only the group) so I implemented the checks a “level further down” on the manifold without the group decorator.

I don't think it's correct to move that check to product manifold. Why can't SpecialEuclidean intercept check_point/check_vector calls, check for the special matrix representation, and only if what it gets doesn't look like it call respective check functions on product manifold?

mateuszbaran:
> While the remaining small things in Manifolds are not yet resolved (and are tricky) – it seems this PR here is nearly done, so I already did code coverage (even a line less than before, just 2 false nonhits) – so that we only should check the docs again, but this should be fine now.

Cool! I can take a look at some issue tomorrow.

mateuszbaran:
> > Ah! I had an idea – it is another representation for the product manifold (not only the group) so I implemented the checks a “level further down” on the manifold without the group decorator.
> 
> I don't think it's correct to move that check to product manifold. Why can't SpecialEuclidean intercept check_point/check_vector calls, check for the special matrix representation, and only if what it gets doesn't look like it call respective check functions on product manifold?

It seems like it should be possible to intercept it at some level with traits injected.

kellertuer:
> > > Ah! I had an idea – it is another representation for the product manifold (not only the group) so I implemented the checks a “level further down” on the manifold without the group decorator.
> > 
> > 
> > I don't think it's correct to move that check to product manifold. Why can't SpecialEuclidean intercept check_point/check_vector calls, check for the special matrix representation, and only if what it gets doesn't look like it call respective check functions on product manifold?
> 
> It seems like it should be possible to intercept it at some level with traits injected.

I think the current way I the correct one since

1) `is_point` already passes down from SpecialEuclidean (the group) to the underlying manifold and
2) there is also a `ProductRepr` on this manifold-level, which is also where we usually check the sizes

Also there was already a const for this type so it was not even much to do, just change the type of `M`upfront from the group const to the manifold const – and it works like a charm.

mateuszbaran:
Hm, yes, that's definitely a much easier solution.

mateuszbaran:
I've fixed some issues with statistics tests but I fear that they can't all be resolved without introducing the three-layered scheme for `mean`, `mean!`, `median` and `median!`.

kellertuer:
I would have thought that they are high level enough that it would not be necessary, but if it is necessary, then lets do it.

mateuszbaran:
There is quite a bit of mixing special implementations for manifolds and different methods, so it looks to me like it's needed to have full functionality.

Additionally, I've traced down one of the problems with metric manifold and a few more tests are passing now.

kellertuer:
You're right there is quite some special cases – well, I think its not too bad to decorate/trait them as well.

mateuszbaran:
Regarding the last metric.jl failure: should `project`/`project!` have the trait `IsExplicitDecorator` for a non-default metric manifold decorator? Manifolds.jl master branch currently doesn't forward these to the wrapped manifold. I don't really have a strong opinion here.

kellertuer:
I think it is a reasonable default to put the ExplicitDecorator last and let `MetricManifold` pass by default to the inner, yes.

mateuszbaran:
OK, so now all metric.jl tests pass :tada: .

kellertuer:
Nice, also the test already run through now (instead of stopping early), just something on Circle broke again for example, log_lie has to have a second layer still (resolving identity on layer 1, dispatching to manifolds only on the second) - the one I do not yet understand is the `has_approx_invariant_metric` error is one of the few ones that I do not yet understand.

mateuszbaran:
The problem there is that `HasBiinvariantMetric` doesn't correctly forward for things that are not `MetricManifold`. I'm not really sure how to fix it -- probably `HasBiinvariantMetric` for functions like `exp` should forward to the wrapped manifold for any wrapper, not just `MetricManifold`.

mateuszbaran:
Anyway, I've uploaded all fixes and I'll make a break now.

kellertuer:
I will have to check the old version as well to understand more where it dispatched to - sure.

And breaks – I am not even coding today, yesterday was quite long (though in the end I just did code coverage and docs), so today I will not commit something; breaks are also important, especially in such a technical, tricky redesign.

kellertuer:
General Linear and Special Linear are fixed. The problem was not `has_approx_invariant_metrc` but more like a PEBKAC (Problem Exists Between Keyboard and Computer) – both groups are left-invariant but not right-invariant and someone (tm; me!) transferred that code to them being invariant – so the right-invariant tests (surprisingly!) failed.

mateuszbaran:
Great, it looks like we are finally approaching the end of this refactoring :+1: .

kellertuer:
log_lie is fixed and also retract for this one is hit again,

So we have just left, that the ODERetraction hits a little too often for groups (might check that still now) and layers for mean/median.

Oh and the docs of course ;) I split a few files for groups, so the groups docs should be rearranged a little.

kellertuer:
I think after a careful read of the docs, this one should be ready to merge – because on the other (Manifolds.jl) branch we then still have to check for code coverage once 0.13.0 here is out.

kellertuer:
I started revising the docs and reading them again carefully.
This should hopefully be the last step of this PR.
Current status

- [x] Home
- [x] Example
- [x] Design principles
- [ ] The Manifold type
- [ ] Functions on Manifolds
- [ ] Manifolds
- [ ] Decorating / Extorting a Manifold
- [ ] Bases for tangent spaces

kellertuer:
I felt that vector spaces were quite hidden within the basis page – so I rephrased the manifold-type page (which contained the point type already) to also include the vector space types – also nicer for the following part about the forward-macros on the same page.

mateuszbaran:
Good idea. I should have some time during the weekend to check the documentation here.

kellertuer (approved):
I am done with reading and fixing (in between breaking) a few things. To me this looks good to go

docs/src/example.md
```diff
@@ -0,0 +1,256 @@
...
+
+## [Conclusion](@id manifold-tutorial-outlook)
+
+You can now just continue implementing further functions from `ManifoldsBase.jl`.
+but with just [`exp!`](@ref exp!(M::AbstractManifold, q, p, X)) you for example already have
+
+* [`geodesic`](@ref geodesic(M::AbstractManifold, p, X)) the (not necessarily shortest) geodesic emanating from `p` in direction `X`.
+* the [`ExponentialRetraction`](@ref), that the [`retract`](@ref retract(M::AbstractManifold, p, X)) function uses by default.
+
+For the [`shortest_geodesic`](@ref shortest_geodesic(M::AbstractManifold, p, q)) the implementation of a logarithm [`log`](@ref ManifoldsBase.log(M::AbstractManifold, p, q)), again better a [`log!`](@ref log!(M::AbstractManifold, X, p, q)) is necessary.
```
mateuszbaran, line 246:
```suggestion
For the [`shortest_geodesic`](@ref shortest_geodesic(M::AbstractManifold, p, q)) the implementation of a logarithm [`log`](@ref ManifoldsBase.log(M::AbstractManifold, p, q)), or just [`log!`](@ref log!(M::AbstractManifold, X, p, q)) is sufficient.
```

src/maintypes.jl
```diff
@@ -1,32 +1,35 @@
 """
-    AbstractManifold{F}
+    AbstractManifold{𝔽 <: AbstractNummbers}
```
mateuszbaran, line 2:
```suggestion
    AbstractManifold{𝔽}
```
This part is not in the type definition.

src/vector_transport.jl
```diff
@@ -51,45 +51,24 @@ compute ``Y = \operatorname{retr}_p^{-1}q``.
...
-       AbstractLinearVectorTransportMethod end
-function DifferentiatedRetractionVectorTransport(::R) where {R<:AbstractRetractionMethod}
-    return DifferentiatedRetractionVectorTransport{R}()
+       AbstractLinearVectorTransportMethod
+    retraction::R
+    function DifferentiatedRetractionVectorTransport(
+        r::R,
+    ) where {R<:AbstractRetractionMethod}
+        return new{R}(r)
+    end
```
mateuszbaran, line 62:
```suggestion
```
I think we don't need this constructor.

kellertuer: Why? Would this work automatically?

mateuszbaran: Yes, it looks exactly like the default one.

docs/src/example.md
```diff
@@ -0,0 +1,257 @@
+# [How to implement your own manifold](@id manifold-tutorial)
+
+```@meta
+CurrentModule = ManifoldsBase
+```
+
+## Introduction
+
+This tutorial explains, how to implement a manifold using the `ManifoldsBase.jl` interface.
```
mateuszbaran, line 9:
```suggestion
This tutorial explains how to implement a manifold using the `ManifoldsBase.jl` interface.
```

docs/src/manifolds.md
```diff
@@ -0,0 +1,63 @@
+# Manifolds
+
+While the interface `ManifoldsBase.jl` does not cover concrete manifolds, it provides a few
+helpers to build or create manifolds based on existing manifolds
+
+## [(Abstract) power manifold](@id sec-power-manifold)
+
+A power manifold is constructed like higher dimensional vector spaces are formed from the real line, just that for every point ``p = (p_1,\ldots,p_n) ∈ \mathcal M^n`` on the power manifold ``\mathcal M^n`` the entries of ``p`` are points ``p_1,\ldots,p_n ∈ \mathcal M`` on some manifold ``\mathcal M``. Note that ``n`` can also be replaced by multiple values, such that ``p`` is not a vector but a matrix or tensor of points.
```
mateuszbaran, line 8:
```suggestion
A power manifold is constructed like higher dimensional vector spaces are formed from the real line, just that for every point ``p = (p_1,\ldots,p_n) ∈ \mathcal M^n`` on the power manifold ``\mathcal M^n`` the entries of ``p`` are points ``p_1,\ldots,p_n ∈ \mathcal M`` on some manifold ``\mathcal M``. Note that ``n`` can also be replaced by multiple values, such that ``p`` is not a vector but a matrix or a multi-index array of points.
```

docs/src/manifolds.md
```diff
@@ -0,0 +1,63 @@
...
+This is the implicit case, since the manifold type itself does not carry any information about the embedding, just the trait and the function definition do.
+
+### [Explicit case: the `EmbeddedManifold`](@id subsec-explicit-embedded)
+
+The [`EmbeddedManifold`](@ref) itself is an [`AbstractDecoratorManifold`](@ref) so it is a case of the implicit embedding itself, but internally stores both the original manifold and the embedding.
+They are also parameters of the type.
+This way, additional embeddings can be modelled. That is, if the manifold is implemented using the implicit embedding approach from before but can also be implemented using a _different_ embedding, then this method should be chosen, since you can dispatch functions that you want to implement in this embedding then on the type which explicitly has the manifold and its embedding as parameters.
+
+Hence this case should be used for any further embedding after the first or if the default implementation works without an embedding and the alternative needs one.
```
mateuszbaran, line 57: Maybe we should add that this type only supports defining one embedding for each pair of manifolds.

mateuszbaran (commented):
A few more comments :).

docs/src/index.md
```diff
@@ -0,0 +1,33 @@
...
+You add it as a dependency if you plan to work on manifolds (generically) or if you plan to
+define own manifolds in a package.
+For a package that (only) depends on `ManifoldsBase.jl`, see [Manopt.jl](https://manoptjl.org/stable/),
+which implements optimization algorithms on manifolds using this interface.
+These optimisation algorithms can hence be used with any manifold implemented based on `ManifoldsBase.jl`.
+
+For a library of manifolds implemented using this interface [Manifolds.jl](https://juliamanifolds.github.io/Manifolds.jl/stable/).
+
+Your package is using `ManifoldsBase`?
+We would like to add that here as well. Either [write an issue](https://github.com/JuliaManifolds/ManifoldsBase.jl/issues/new)
```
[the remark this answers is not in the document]
kellertuer: Exactly.

docs/src/projections.md
```diff
@@ -0,0 +1,51 @@
...
+
+```math
+  \operatorname*{arg\,min}_{q\in \mathcal M}\ d_{\mathcal N}(i(q),p)
+```
+
+And this resulting ``q`` we call the projection of ``p`` onto the manifold ``\mathcal M``.
+
+The second projection we can look at is for a given a point ``p∈\mathcal M`` and a vector in ``X\in T_{i(p)}\mathcal N`` in the embedding,
+where we can similarly look for the closest tangent vector ``Y∈ T_p\mathcal M``, which we have to embed itself before itself.
+Embedding a tangent vector is the same as using the push forward ``\mathrm{d}i_p`` of the embedding (at ``p``).
```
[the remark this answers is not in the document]
kellertuer: What does IIUC mean? For rotations the embedding might again be a little strange due to the Lie algebra or the representation without the `S`?

docs/src/manifolds.md
kellertuer: Indeed it does, but I even rarely see different embeddings for one manifold into different other ones (yes there are some).
In principle one could distinguish different embeddings of one manifold M into one Manifold N by changing the type of the point p?

Overall I feel this might be too much detail?

docs/src/index.md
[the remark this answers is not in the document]
mateuszbaran: JuliaHub automatically generates list of dependents: https://juliahub.com/ui/Packages/ManifoldsBase/0nPMu/0.12.13?page=2 :slightly_smiling_face:

docs/src/manifolds.md
mateuszbaran: Yes, even `EmbeddedManifold` does not have much use currently (as far as I'm aware currently it's only for embedding SE in GL). So I don't see much reason to work on extending it to different embeddings, as it's not something needed. I'm only a bit worried that the current wording may suggest that having such different embeddings is possible.

kellertuer: THen propose a new wording :)

docs/src/projections.md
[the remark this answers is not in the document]
mateuszbaran: IIUC = if I understand correctly

Yes, it's a little strange but I think it's worth mentioning that `embed` isn't guaranteed to be that pushforward (though it usually is).

docs/src/index.md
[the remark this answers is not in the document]
mateuszbaran: Sure, why not :+1:

docs/src/projections.md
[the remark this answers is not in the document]
mateuszbaran: I'd add just "usually", I'm not sure if it's just up to change of representation.

docs/src/manifolds.md
```diff
@@ -0,0 +1,63 @@
...
+Adding a method to the [`active_traits`](@ref ManifoldsBase.active_traits) function for a manifold that returns an [`AbstractTrait`](@ref)
+[`IsEmbeddedManifold`](@ref), makes that manifold an embedded manifold. You just have to also define [`get_embedding`](@ref) so that appropriate functions are passed on to that embedding.
+This is the implicit case, since the manifold type itself does not carry any information about the embedding, just the trait and the function definition do.
+
+### [Explicit case: the `EmbeddedManifold`](@id subsec-explicit-embedded)
+
+The [`EmbeddedManifold`](@ref) itself is an [`AbstractDecoratorManifold`](@ref) so it is a case of the implicit embedding itself, but internally stores both the original manifold and the embedding.
+They are also parameters of the type.
+This way, additional embeddings can be modelled. That is, if the manifold is implemented using the implicit embedding approach from before but can also be implemented using a _different_ embedding, then this method should be chosen, since you can dispatch functions that you want to implement in this embedding then on the type which explicitly has the manifold and its embedding as parameters.
```
mateuszbaran, line 55:
```suggestion
This way, an additional embedding of one manifold in another can be modelled. That is, if the manifold is implemented using the implicit embedding approach from before but can also be implemented using a _different_ embedding, then this method should be chosen, since you can dispatch functions that you want to implement in this embedding then on the type which explicitly has the manifold and its embedding as parameters.
```
I think this wording is more appropriate without going into too much detail.

docs/src/projections.md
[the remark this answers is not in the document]
kellertuer: ok

mateuszbaran:
OK, I think this is ready to be merged now :tada: .

kellertuer:
Ok, let‘s go!

PR #93, opened 2022-02-03
Allocations with no points default to `Array`s of `Float64`

mateuszbaran:
Needed for second part of https://github.com/JuliaManifolds/Manifolds.jl/pull/450 , that is random points without array as an "example".


kellertuer:
I do not See that directly, but why can this not happen (the ambiguity)?

kellertuer:
Besides my question above, this looks like really good defaults to introduce, so with me getting the detail why this can't happen – I would be fine merging this :)

mateuszbaran:
> I do not See that directly, but why can this not happen (the ambiguity)?

You'd have to call `allocate_result(M, get_coordinates)` for example, and every time you get coordinates you have some point that you can put as an argument there. It could only happen if there was a valid way to call `get_vector` or `get_coordinates` without any points, coordinates or vectors.

kellertuer:
Ah, thanks for the explanations. You are right, this is a case that should not happen if using allocate correctly. So then, this looks fine to me.

kellertuer (approved):

PR #94, opened 2022-02-04
Introduce defaults for atol and rtol for is_point and is_vector

kellertuer:
This PR introduces a default `atol` (=0) and rtol (`sqrt(eps)`) for the high-level functions `is_point` and `is_vector`.

The lower level `check_` functions do not have this detail, in case someone wants to be sure to not have it.
Formerly I fear this is changing a default and hence breaking change – also integer vectors do not cope well with my idea of the eps-default so if there is a better idea, feel free to mention it here.

The PR yields that
1) high-level usage is easier to use and tolerant to some extend
2) low level use can still be as exact as before.

### DIFF
--- a/src/ManifoldsBase.jl
+++ b/src/ManifoldsBase.jl
@@ -417,8 +417,8 @@ function is_point(
     M::AbstractManifold,
     p,
     throw_error = false;
-    atol::Real = 0,
-    rtol = atol > 0 ? 0 : sqrt(eps(number_eltype(p))),
+    atol::Real = zero(number_eltype(p)),
+    rtol = atol > zero(float(number_eltype(p))) ? zero(float(number_eltype(p))) : sqrt(eps(float(number_eltype(p)))),
     kwargs...,
 )
     mpe = check_point(M, p; atol = atol, rtol = rtol, kwargs...)
@@ -446,8 +446,8 @@ function is_vector(
     X,
     throw_error = false;
     check_base_point = true,
-    atol = 0.0,
-    rtol = atol > 0 ? 0 : sqrt(eps(number_eltype(X))),
+    atol = zero(number_eltype(X)),
+    rtol = atol > zero(float(number_eltype(X))) ? zero(float(number_eltype(X))) : sqrt(eps(float(number_eltype(X)))),
     kwargs...,
 )
     if check_base_point
--- a/src/ManifoldsBase.jl
+++ b/src/ManifoldsBase.jl
@@ -413,8 +413,15 @@ is `true`, the function either returns `true` or throws an error. By default the
 calls [`check_point(M, p; kwargs...)`](@ref) and checks whether the returned value
 is `nothing` or an error.
 """
-function is_point(M::AbstractManifold, p, throw_error = false; kwargs...)
-    mpe = check_point(M, p; kwargs...)
+function is_point(
+    M::AbstractManifold,
+    p,
+    throw_error = false;
+    atol::Real = 0,
+    rtol = atol > 0 ? 0 : sqrt(eps(number_eltype(p))),
+    kwargs...,
+)
+    mpe = check_point(M, p; atol = atol, rtol = rtol, kwargs...)
     mpe === nothing && return true
     return throw_error ? throw(mpe) : false
 end
@@ -439,10 +446,12 @@ function is_vector(
     X,
     throw_error = false;
     check_base_point = true,
+    atol = 0.0,
+    rtol = atol > 0 ? 0 : sqrt(eps(number_eltype(X))),
     kwargs...,
 )
     if check_base_point
-        mpe = check_point(M, p; kwargs...)
+        mpe = check_point(M, p; atol = atol, rtol = rtol, kwargs...)
         if mpe !== nothing
             if throw_error
                 throw(mpe)
@@ -451,7 +460,7 @@ function is_vector(
             end
         end
     end
-    mtve = check_vector(M, p, X; kwargs...)
+    mtve = check_vector(M, p, X; atol = atol, rtol = rtol, kwargs...)
     mtve === nothing && return true
     return throw_error ? throw(mtve) : false
 end
--- a/test/domain_errors.jl
+++ b/test/domain_errors.jl
@@ -3,13 +3,13 @@ using Test
 
 struct ErrorTestManifold <: AbstractManifold{ℝ} end
 
-function ManifoldsBase.check_point(::ErrorTestManifold, x)
+function ManifoldsBase.check_point(::ErrorTestManifold, x; kwargs...)
     if any(u -> u < 0, x)
         return DomainError(x, "<0")
     end
     return nothing
 end
-function ManifoldsBase.check_vector(M::ErrorTestManifold, x, v)
+function ManifoldsBase.check_vector(M::ErrorTestManifold, x, v; kwargs...)
     mpe = check_point(M, x)
     mpe === nothing || return mpe
     if any(u -> u < 0, v)
@@ -22,13 +22,13 @@ end
     M = ErrorTestManifold()
     @test isa(check_point(M, [-1, 1]), DomainError)
     @test check_point(M, [1, 1]) === nothing
-    @test !is_point(M, [-1, 1])
-    @test is_point(M, [1, 1])
-    @test_throws DomainError is_point(M, [-1, 1], true)
+    @test !is_point(M, [-1.0, 1.0])
+    @test is_point(M, [1.0, 1.0])
+    @test_throws DomainError is_point(M, [-1.0, 1.0], true)
 
     @test isa(check_vector(M, [1, 1], [-1, 1]), DomainError)
     @test check_vector(M, [1, 1], [1, 1]) === nothing
-    @test !is_vector(M, [1, 1], [-1, 1])
-    @test is_vector(M, [1, 1], [1, 1])
-    @test_throws DomainError is_vector(M, [1, 1], [-1, 1], true)
+    @test !is_vector(M, [1.0, 1.0], [-1.0, 1.0])
+    @test is_vector(M, [1.0, 1.0], [1.0, 1.0])
+    @test_throws DomainError is_vector(M, [1.0, 1.0], [-1.0, 1.0], true)
 end
--- a/test/embedded_manifold.jl
+++ b/test/embedded_manifold.jl
@@ -114,12 +114,12 @@ struct NotImplementedEmbeddedManifold3 <: AbstractEmbeddedManifold{ℝ,DefaultEm
         @test ManifoldsBase.default_embedding_dispatch(M) === Val(false)
         @test get_embedding(M) == ManifoldsBase.DefaultManifold(1, 3)
         # Check fallbacks to check embed->check_manifoldpoint Defaults
-        @test_throws DomainError is_point(M, [1, 0, 0], true)
-        @test_throws DomainError is_point(M, [1 0], true)
-        @test is_point(M, [1 0 0], true)
-        @test_throws DomainError is_vector(M, [1 0 0], [1], true)
-        @test_throws DomainError is_vector(M, [1 0 0], [0 0 0 0], true)
-        @test is_vector(M, [1 0 0], [1 0 1], true)
+        @test_throws DomainError is_point(M, [1.0, 0.0, 0.0], true)
+        @test_throws DomainError is_point(M, [1.0 0.0], true)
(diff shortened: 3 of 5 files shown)


kellertuer:
I am not sure why JuliaFormatter fails. Locally I have the same version and that runs fine. Hm.

src/ManifoldsBase.jl
```diff
@@ -413,8 +413,15 @@ is `true`, the function either returns `true` or throws an error. By default the
...
 is `nothing` or an error.
 """
-function is_point(M::AbstractManifold, p, throw_error = false; kwargs...)
-    mpe = check_point(M, p; kwargs...)
+function is_point(
+    M::AbstractManifold,
+    p,
+    throw_error = false;
+    atol::Real = 0,
+    rtol = atol > 0 ? 0 : sqrt(eps(number_eltype(p))),
```
mateuszbaran, line 421: I'm not sure if that's the right default. What about something like
```julia
rtol = atol > 0 ? zero(number_eltype(p)) : default_rtol(M, p)
```
where `default_rtol(M, p)` would be by default more on the order of `sqrt(manifold_dimension(M))*eps(number_eltype(p))`? Or `prod(representation_size(M))` instead of `manifold_dimension(M)` if `representation_size(M)` is not nothing?

kellertuer: I also was not sure, sure we can scale by the sqrt of the manifold dimension, on the other hand its the relative tolerance, so I thought the left hand side (being a quotient anyways) would cancel the manifold dimension out.

mateuszbaran: My mental image for the proposed choice is that when points are represented using N numbers, then an epsilon change in every single one of them should, by default, also give a "close point". When each "coordinate" differs by eps(), then the Euclidean distance is sqrt(N)*eps().

kellertuer: Hm, i am unsere here,m for me a relative error would mean that the sqrt(n) is “factored out” before that check anyways, but I will try to find out

kellertuer: We could adapt the `rtoldefault` that Julia itself uses as wel? For me the norm implementation indicates that the Julia default does not do a sort(N) in the rtol: https://github.com/JuliaLang/julia/blob/7ecbfa9c926bdf380f824766b9d165ec9b133eaa/base/floatfuncs.jl#L313-L318

mateuszbaran: Hm, you may be right but `eps(number_eltype(p))` seems too small and `sqrt(eps(number_eltype(p)))` seems too large. Very likely there is no reasonable manifold-independent default value.

kellertuer: Remember that it is relative, so the manifold dimension (like vector dimensions in the classical case) should be taken care of.

kellertuer: So do we keep the current state? Then I would just document it and this PR should be fine.

mateuszbaran: I have one other minor point about type stability but leaving sqrt(eps) here seems to be the best option.

mateuszbaran, line 421:
```suggestion
    atol::Real = zero(number_eltype(p)),
    rtol = atol > zero(float(number_eltype(p))) ? zero(float(number_eltype(p))) : sqrt(eps(float(number_eltype(p)))),
```

src/ManifoldsBase.jl
```diff
@@ -439,10 +446,12 @@ function is_vector(
     X,
     throw_error = false;
     check_base_point = true,
+    atol = 0.0,
+    rtol = atol > 0 ? 0 : sqrt(eps(number_eltype(X))),
```
mateuszbaran, line 450:
```suggestion
    atol = zero(number_eltype(X)),
    rtol = atol > zero(float(number_eltype(X))) ? zero(float(number_eltype(X))) : sqrt(eps(float(number_eltype(X)))),
```

test/domain_errors.jl
```diff
@@ -22,13 +22,13 @@ end
...
+    @test_throws DomainError is_point(M, [-1.0, 1.0], true)
 
     @test isa(check_vector(M, [1, 1], [-1, 1]), DomainError)
     @test check_vector(M, [1, 1], [1, 1]) === nothing
-    @test !is_vector(M, [1, 1], [-1, 1])
-    @test is_vector(M, [1, 1], [1, 1])
-    @test_throws DomainError is_vector(M, [1, 1], [-1, 1], true)
+    @test !is_vector(M, [1.0, 1.0], [-1.0, 1.0])
+    @test is_vector(M, [1.0, 1.0], [1.0, 1.0])
+    @test_throws DomainError is_vector(M, [1.0, 1.0], [-1.0, 1.0], true)
```
mateuszbaran, line 33: With my change to `is_vector` and `is_point` this should work with integers.

mateuszbaran (commented):
I'm not sure what problem does this change actually solve? I've noticed that these defaults are exactly the same as in Julia's `isapprox` so all it seems to do is overriding potential manifold-specific tolerances introduce in `check_vector` and `check_point`?

src/ManifoldsBase.jl
mateuszbaran: I've just made a comment below about it :slightly_smiling_face: .

kellertuer:
> I'm not sure what problem does this change actually solve? I've noticed that these defaults are exactly the same as in Julia's `isapprox` so all it seems to do is overriding potential manifold-specific tolerances introduce in `check_vector` and `check_point`?

Well for now they usually endet up being `atol=rtol=0` which is often too strong. This introduces the same default as in Julia for our (`M`-prefixed) `isapprox`es for points and vectors.

test/domain_errors.jl
kellertuer: ah, I was lazy and assumed, that these are just not good types anymore then. I will add an integer test then to ensure that this is further doable.

mateuszbaran:
Which code causes `rtol` to be 0? I've checked for `Euclidean` and `Sphere` and `rtol`s are greater than 0:
```julia
julia> using Manifolds

julia> M = Euclidean(2)
Euclidean(2; field = ℝ)

julia> isapprox(M, [1, 1], [1, 1+1e-9])
true

julia> M = Sphere(1)
Sphere(1, ℝ)

julia> isapprox(M, [1, 1], [1, 1+1e-9])
true
```

kellertuer:
Interesting, I am not sure, but it was the reason `Grassmann` failed in the slack example.

mateuszbaran:
In this case I'd assume that just Grassmann should have less strict default tolerance.

kellertuer:
Since I just had to set atop to `2e-14` I assume it has an `rtol` equal to zero, but I am not yet sure where this comes from. I also do not know why your examples _do_ have a relative tolerance, since I do not see where our isapproxies do get that from.

mateuszbaran:
Grassman example has a positive rtol but it doesn't matter that much when the two compared matrices are close to zero, for example:
```julia
julia> a = p'*vv
4×4 Matrix{Float64}:
 -2.45927e-16   2.48871e-17   9.14629e-17  -5.65537e-17
  1.89489e-16  -3.88603e-17  -2.96737e-16  -1.15517e-17
  2.34291e-16  -2.13559e-17  -9.71807e-17   4.89858e-17
  1.14101e-16  -1.48652e-17  -9.15055e-17   1.38072e-17

julia> b = vv'*p
4×4 Matrix{Float64}:
 -2.45927e-16   1.89489e-16   2.34291e-16   1.14101e-16
  2.48871e-17  -3.88603e-17  -2.13559e-17  -1.48652e-17
  9.14629e-17  -2.96737e-16  -9.71807e-17  -9.15055e-17
 -5.65537e-17  -1.15517e-17   4.89858e-17   1.38072e-17
```
See here: https://github.com/JuliaManifolds/Manifolds.jl/blob/f3e1ff5fe567d1df9dac327de22ae913dc8a20f1/src/manifolds/Grassmann.jl#L104 .

kellertuer:
you mean as a default for `check_point`? Hm, I do not feel well with that, it breaks with the global default then? Let's keep it as is I would say.

mateuszbaran:
Right, we probably shouldn't touch the default. One thing we can do is to write in documentation of `is_vector` or `check_vector` that for vectors close to the zero vector it may be a good idea to manually specify atol.

kellertuer:
That sounds reasonable.

kellertuer:
Hm, seems I am not clever enough for productRepr and such. I adapted `vector_transport_direction` to `parallel_transport_to` (and direction and both also in mutating). the non-mutating works, the mutating works if I call in the debugger from within the function the PT on a single component, but the map ends in the parallel_transport_to! method not to be found. NO clue why (since calling it actually separately in the debuger works). 

My Power Manifold update might have broken essential. Hmpf. Maybe this still gets an infinite story.

mateuszbaran:
I'll check that tomorrow or on Friday.

kellertuer:
I at least fixed essential again – and improved error messages when checking sizes on PowerManifolds.

JeffreySarnoff:
This topic is one with which I have experience.

"never use floating-point tests for equality" does not go far enough, often. Comparing floats that obtain as vector norms, which may appear when determining the angle between two vectors, is unsafe as it is usually done [AngleBetweenVectors.jl](https://github.com/JeffreySarnoff/AngleBetweenVectors.jl).  Your earlier approach to the determination of `point` and `line` seems to satsifice rather than resolve or, better, to extend usefulness.  Whenever geometrical entities are woven into a software design, the users tend to lean heavily on their own expectations of how these entities are to behave and interwork.  Almost never is the user that is expert in some other domain going to consider tolerances beyond `are there?` `is the default good for me .. hope so.`
   
If you are going to use floaty geometricification, which is entirely rational from a performance perspective, maybe your work would often work well for others by providing an inner accuracy that exceeds the outer accuracy to be specified.  If it is possible to off-load abs and rel accuracy examination from most [not all] casual uses of your work *I have not examined manifolds, so dunno*, you should find describing how to best work with these pkgs much simpler.

kellertuer:
What is floaty geometrification? I somehow seem to miss what you want to tell us here.

JeffreySarnoff:
> What is floaty geometrification?

I revised the comment.

JeffreySarnoff:
what do those two very similar matricies each mean, what do they each signify or encode and how do they differ in a non-numerical sense?

```
julia> a = p'*vv
julia> b = vv'*p
```

mateuszbaran:
I've replicated a bit different but similar in style example.
We have a matrix `p` such that `p'p` is approximately a unit matrix:
```julia
julia> p = [-0.04705910362254517 -0.12156937692933076; 0.24137167076604518 -0.34985331139364395; 0.6096415977032226 0.7704681918979749; 0.7535663737639254 -0.5188470992242143]
4×2 Matrix{Float64}:
 -0.0470591  -0.121569
  0.241372   -0.349853
  0.609642    0.770468
  0.753566   -0.518847

julia> p'p
2×2 Matrix{Float64}:
  1.0          -4.63875e-18
 -4.63875e-18   1.0

```
Now we'd like to have a random matrix `X` such that `p'*X = -X'*p` (let's take the real numbers case here), so we take a random matrix and modify it to satisfy the equation:
```julia
julia> X = randn(4, 2)
4×2 Matrix{Float64}:
 -1.46691    -0.799284
 -0.301611   -0.109984
  0.23441     0.544457
  0.0145982  -1.39093

julia> X = X - p * p' * X
4×2 Matrix{Float64}:
 -1.4043    -0.677247
 -0.178008   0.506923
 -0.209134  -0.00938728
  0.138511  -0.197069
```
But numerically it doesn't exactly satisfy it:
```julia
julia> p'X+X'p
2×2 Matrix{Float64}:
 -3.11344e-17  -1.74839e-17
 -1.74839e-17   3.36018e-18

julia> isapprox(p'X, -X'p)
false
```

JeffreySarnoff:
You should not be treating matrices the same way that scalars  test for equality [approximate or determinate].

kellertuer:
Thanks for al l your input. I still have not yet understood what the recommendation here now would be.

kellertuer:
Hm, somehow I still miss something on direction – it feels like previously you could call it with _anything_ and it returned a direction (metric? operation? manifold? somehow all of them), so I am not yet sure how to model this in the decorator scheme (see errors the `test/group/metric.jl` ) and I am a little stuck how to solve this.

kellertuer:
Maybe we keep the current state for now - at some point it might be nice to check for a few specific manifold and update our high.level functions and their tolerance.s

PR #95, opened 2022-03-30
Fix zero argument result allocation for nested power manifold

mateuszbaran:
A small fix needed for https://github.com/JuliaManifolds/Manifolds.jl/pull/467 .

### DIFF
--- a/src/PowerManifold.jl
+++ b/src/PowerManifold.jl
@@ -209,8 +209,12 @@ function allocate_result(M::PowerManifoldNested, f, x...)
         ]
     end
 end
-function allocate_result(::PowerManifoldNestedReplacing, f, x...)
-    return copy(x[1])
+function allocate_result(M::PowerManifoldNestedReplacing, f, x...)
+    if length(x) == 0
+        return [allocate_result(M.manifold, f) for _ in get_iterator(M)]
+    else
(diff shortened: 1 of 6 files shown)


src/PowerManifold.jl
```diff
@@ -200,7 +200,7 @@ function Base.:^(
 end
 
 function allocate_result(M::PowerManifoldNested, f, x...)
-    if representation_size(M.manifold) === ()
```
kellertuer, line 203: can you explain what case this does change (in pratice/examples)?

src/PowerManifold.jl
```diff
@@ -209,8 +209,12 @@ function allocate_result(M::PowerManifoldNested, f, x...)
         ]
     end
 end
-function allocate_result(::PowerManifoldNestedReplacing, f, x...)
-    return copy(x[1])
```
kellertuer, line 213: Same question as above – just curious :)

mateuszbaran: What was failing before is `allocate_result(PowerManifold(...), rand)`, because there was no `x[1]`.

kellertuer: Ah! sure – that makes sense.

kellertuer (approved):

PR #96, opened 2022-04-06
Resolve some basis-related ambiguities

mateuszbaran:
This resolves some ambiguities detected in Manifolds.jl tests.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "ManifoldsBase"
 uuid = "3362f125-f0bb-47a3-aa74-596ffd7ef2fb"
 authors = ["Seth Axen <seth.axen@gmail.com>", "Mateusz Baran <mateuszbaran89@gmail.com>", "Ronny Bergmann <manopt@ronnybergmann.net>", "Antoine Levitt <antoine.levitt@gmail.com>"]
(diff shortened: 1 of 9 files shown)


mateuszbaran:
There is also a small issue that some methods of `vector_transport_along` require `c::AbstractVector` while other do not. We should probably decide one way or the other.

kellertuer:
I think I would opt for putting the AbstractVector, what do you think?

mateuszbaran:
OK, I'll add `AbstractVector` then.

I have no idea why we lost coverage though.

kellertuer:
Uh now we even lost 10 lines, hm. We might have to check that, then; I can try to find time, but only on Friday.

mateuszbaran:
I'll take a look at it, at least things related to vector transport could be real.

mateuszbaran:
Hm, we don't have some generic `DifferentiatedRetractionVectorTransport` methods but no concrete implementation. I'm not sure why it didn't show up earlier in coverage.

mateuszbaran:
Coverage if full now, I think this can be merged.

kellertuer:
Looking forward to seeing the effect on Manifolds :)

kellertuer (approved):

PR #97, opened 2022-04-14
More strict typing of some methods

mateuszbaran:
I've put some bounds on types in a few methods. We may also need a few more methods for basis-related function forwarding but I'll think a bit more how to introduce them.


kellertuer:
Looks good.

mateuszbaran:
I think this is ready now.

PR #99, opened 2022-04-25
Introduce a default for embed/embed! and retraction by projection

kellertuer:
update the default for vector transport by projection, which now can also include a embed (since we have a default).

### DIFF
--- a/src/retractions.jl
+++ b/src/retractions.jl
@@ -709,10 +709,7 @@ computes the mutating variant of the [`ProjectionRetraction`](@ref).
 By default this [`porject!`](@ref)s `X` onto the manifold (in place of `q`)
 after [`embed!`](@ref)ing `X`.
 """
-retract_project!(M::AbstractManifold, q, p, X) = project!(M, q, embed(M, p, X))
-# Note that we have to use embed (not embed!) since we do not have memory to store this embedded value in
-
-
+retract_project!(M::AbstractManifold, q, p, X)
 
 function retract_project! end
 
--- a/src/retractions.jl
+++ b/src/retractions.jl
@@ -706,8 +706,6 @@ function retract_pade! end
     retract_project!(M::AbstractManifold, q, p, X)
 
 computes the mutating variant of the [`ProjectionRetraction`](@ref).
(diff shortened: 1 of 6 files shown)


src/retractions.jl
```diff
@@ -706,8 +706,13 @@ function retract_pade! end
     retract_project!(M::AbstractManifold, q, p, X)
 
 computes the mutating variant of the [`ProjectionRetraction`](@ref).
+By default this [`porject!`](@ref)s `X` onto the manifold (in place of `q`)
```
mateuszbaran, line 709:
```suggestion
By default this [`project`](@ref)s `X` onto the manifold (in place of `q`)
```

src/retractions.jl
```diff
@@ -706,8 +706,13 @@ function retract_pade! end
     retract_project!(M::AbstractManifold, q, p, X)
 
 computes the mutating variant of the [`ProjectionRetraction`](@ref).
+By default this [`project`](@ref)s `X` onto the manifold (in place of `q`)
+after [`embed!`](@ref)ing `X`.
 """
-retract_project!(M::AbstractManifold, q, p, X)
+retract_project!(M::AbstractManifold, q, p, X) = project!(M, q, embed(M, p, X))
```
mateuszbaran, line 712:
```suggestion
retract_project!(M::AbstractManifold, q, p, X) = project!(M, q, p + embed(M, p, X))
```
Shouldn't it be more like that?

kellertuer: Do we have a + in the embedding? Not in general.

mateuszbaran: Well, `retract_project!(M::AbstractManifold, q, p, X) = project!(M, q, embed(M, p, X))` seems incorrect -- when `X` is a zero vector, retraction should just return `p`, and your version won't do that. So I think we can either accept that this only works when we have `+` in the embedding or not have this default definition at all?

kellertuer: Then, maybe, no default at all is what I would prefer, I think

mateuszbaran (approved):

PR #101, opened 2022-04-28
Improve error message for embedded manifold is_point/is_vector

kellertuer:
(Yet another small update) to have nicer error messages for the case of embedded manifolds when the check for validity in the embedding (already) fails.

I am not sure whether we might have missed a `check_size` in the embedding here (that is one thing that is not done in the newly used `check_` compare to the old `is_`), and we still have to check for test coverage.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "ManifoldsBase"
 uuid = "3362f125-f0bb-47a3-aa74-596ffd7ef2fb"
 authors = ["Seth Axen <seth.axen@gmail.com>", "Mateusz Baran <mateuszbaran89@gmail.com>", "Ronny Bergmann <manopt@ronnybergmann.net>", "Antoine Levitt <antoine.levitt@gmail.com>"]
-version = "0.13.5"
+version = "0.13.6"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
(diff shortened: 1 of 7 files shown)


kellertuer:
Besides test coverage, I am surprised my code fails on Julia 1.0, I still have to check why, because `mpe` can not be nothing when entering the `ManifoldDomainError` (the if around) but it somehow still does end up being so.

kellertuer:
I see what happened, I had a typo in the previous commit which I even pushed (separately) so it triggered the tests and that failed.

So then only test coverage is left to check.

kellertuer:
I hopefully just finished coverage (and added the same style to check size).

Currently the inner error is printed as
```
[0, 0, 0] is not a tangent vector to [1, 1, 0] on PosQuadrantManifold() because it is not a valid tangent vector in its embedding: DomainError(0, "X[1] ≤ 0")
```
Can this maybe improved / is there a nicer print for errors in strings?

kellertuer:
> Can this maybe improved / is there a nicer print for errors in strings?

I stole an idea from our other error prints (changing the last print in the `show error` to another `show error` and the error now prints as 

```
ManifoldDomainError: [0, 0, 0] is not a point on PosQuadrantManifold() because it is not a valid point in its embedding.
DomainError with 0:
p[1] ≤ 0
```

mateuszbaran (approved):
Looks nice :+1: .

PR #104, opened 2022-05-03
Kellertuer/fix check nested embeddings

### LINKED ISSUE #102: check_vector or is_vector – problem with iterated embeddings
In pur last PR we changed the line

https://github.com/JuliaManifolds/ManifoldsBase.jl/blob/5fa0a5c8b46694cdf8186c3668d40ae91f2ed809/src/decorator_trait.jl#L386

from calling ìs_vector` to `check_vector`. I just noticed a problem with that. Namely if we have iterated embeddings (General Grassmann is embedded into Generalised Stiefel is embedded into Euclidean) this only checks one level and does not iterate this (since the decorator is only done for `is_vector`. What do you think should we do about this? If possible I would like to switch back.

kellertuer:
This PR fixes #102 .


src/decorator_trait.jl
```diff
@@ -330,20 +330,24 @@ function is_point(
...
+    try
+        pt = is_point(get_embedding(M, p), embed(M, p), te; kwargs...)
+        !pt && return false # no error thrown (deactivated) but returned false -> return false
+    catch e
+        if e isa DomainError || e isa AbstractManifoldDomainError
+            e = ManifoldDomainError(
+                "$p is not a point on $M because it is not a valid point in its embedding.",
+                e,
+            )
+            te && throw(e)
```
mateuszbaran, line 342:
```suggestion
            ew = ManifoldDomainError(
                "$p is not a point on $M because it is not a valid point in its embedding.",
                e,
            )
            te && throw(ew)
```
Maybe let's call the new variable a different name to avoid confusion.

src/decorator_trait.jl
```diff
@@ -330,20 +330,24 @@ function is_point(
     kwargs...,
 )
     # to be safe check_size first
+    try
+        pt = is_point(get_embedding(M, p), embed(M, p), te; kwargs...)
+        !pt && return false # no error thrown (deactivated) but returned false -> return false
+    catch e
+        if e isa DomainError || e isa AbstractManifoldDomainError
```
mateuszbaran, line 337: If we get a different error we probably want to rethrow it? If I see correctly the current code makes them ignored.

src/decorator_trait.jl
kellertuer: sure, I started with the throw outside the if, then it was easier to keep the variable name.

src/decorator_trait.jl
kellertuer: oh yes, the original code had the throw outside (hence the e that just was reencapsulated), but you're right we should re throw them.

kellertuer:
I think we can also omit the decoration of `check_size` again now since we are back to calling the (already decorated= `is_` functions again?

kellertuer:
One thing I would like to change is the default of `embed` if that is ok, in the old variant the default often was to just return `p` or `X` (for points, vectors, respectively) now several tests fail because with a wrong size (only checked after embed) the allocation/copyto fails. So I would prefer to have an embed that maybe does not fall back to `embed!`? What do you think?

mateuszbaran:
> I think we can also omit the decoration of `check_size` again now since we are back to calling the (already decorated= `is_` functions again?

I'm not sure actually. Decorators can introduce additional valid sizes (i.e. affine matrices for SE(n)), so if someone wrapped SE(n) in something we would probably want to go through undecorating?

> So I would prefer to have an embed that maybe does not fall back to `embed!`? What do you think?

I don't have a preference here. We have to extend the non-mutating `embed` for some manifolds either way.

kellertuer:
I think we can stick with the allocation, currently in Manifolds there is only about 10 tests failing due to that and I will fix these next then.

So here, just code coverage is missing.

mateuszbaran:
Well, we have to either special-case allocation of `embed` (by adding methods to `allocate_result`) or write both mutating and non-mutating `embed` for embeddings that change the size.

kellertuer:
Oh I think it is more a decision what to do as a default (for identity embeddings).

Up to the embed default we had (often) as a default `embed(M,p) = p` - does not check size, really just says: it is really the identity
Now we usually have to allocate (right size of embedding) and copy over p, which I think is nicer, but fails if `p` is too large.

 I just checked and the old default was assumed to work for about 3 or 4 manifolds, where I now manually introduced that. So we can leave this here as is I think. We could also check that p is of right size within embed, if you like, but I am now fine with the current state I think.

kellertuer:
The remaining 6 lines we basically never hit
a) the return false, because if `throw_error`(or `te`) is false, then the try block does not throw an error, so we never reach the return false.
Same for the throwing of other errors, we currently do not have a check that goes wrong on purpose and divides by zero or something.

mateuszbaran:
I think instead of
```julia
                te && throw(ew)
                return false
            end
            throw(e) #not an error we handle here
```
we could just have
```julia
            throw(e) #not an error we handle here
```
which in all reasonable cases will work the same and not leave anything not covered.

kellertuer:
Oh now I chose nearly the first (just renaming (ew) to e again as is was in the very beginning.

mateuszbaran (approved):
OK, that's also fine.

PR #106, opened 2022-05-23
Mutating Geodesics

kellertuer:
A student of mine proposed to maybe also have mutating geodesics. Here is my sketch to these, since it Tok only a few minutes to write these.

But since tests are not 100% working, I would like some feedback whether this is useful, before trying to fix the last few tests with custom point/vector types.

_edit:_ found the bug; when copying the tests to shortest_geodesic I forgot to change one argument. So, this should be done and tested, but whether it is useful and a nice form – let me know.


mateuszbaran:
Sure, these functions make sense.

src/exp_log_geo.jl
```diff
@@ -57,6 +57,36 @@ geodesic(M::AbstractManifold, p, X) = t -> exp(M, p, X, t)
...
+Get the geodesic with initial point `p` and velocity `X` on the [`AbstractManifold`](@ref) `M`.
+A geodesic is a curve of zero acceleration. That is for the curve ``γ_{p,X}: I → \mathcal M``,
+with ``γ_{p,X}(0) = p`` and ``\dot γ_{p,X}(0) = X`` a geodesic further fulfills
+
+```math
+∇_{\dot γ_{p,X}(t)} \dot γ_{p,X}(t) = 0,
+```
+
+i.e. the curve is acceleration free with respect to the Riemannian metric.
+This yields, that the curve has constant velocity that is locally distance-minimizing.
```
mateuszbaran, line 72:
```suggestion
This yields that the curve has constant velocity and is locally distance-minimizing.
```
The curve is locally distance-minimizing, not the velocity.

src/exp_log_geo.jl
```diff
@@ -57,6 +57,36 @@ geodesic(M::AbstractManifold, p, X) = t -> exp(M, p, X, t)
...
+
+This function returns a function `(q,t)` of (time) `t` that mutates `q``.
+
+    geodesic!(M::AbstractManifold, q, p, X, t::Real)
```
mateuszbaran, line 76: I think it would be better to attach these docs to the respective methods below.

src/exp_log_geo.jl
kellertuer: well-spotted, should be changed for the other occurrence as well, since I just copied it from the allocating case.

src/exp_log_geo.jl
```diff
@@ -57,6 +57,36 @@ geodesic(M::AbstractManifold, p, X) = t -> exp(M, p, X, t)
...
+with ``γ_{p,X}(0) = p`` and ``\dot γ_{p,X}(0) = X`` a geodesic further fulfills
+
+```math
+∇_{\dot γ_{p,X}(t)} \dot γ_{p,X}(t) = 0,
+```
+
+i.e. the curve is acceleration free with respect to the Riemannian metric.
+This yields, that the curve has constant velocity that is locally distance-minimizing.
+
+This function returns a function `(q,t)` of (time) `t` that mutates `q``.
```
mateuszbaran, line 74:
```suggestion
Three-argument methods of this function return a function `(q,t)` of (time) `t` that mutates `q`.
```

src/exp_log_geo.jl
kellertuer: Sure, I can do that tomorrow.

mateuszbaran (approved):

PR #115, opened 2022-07-08
Add is_point and is_vector to manifold fallbacks

kellertuer:
For default types we need to first unwrap the data (fall back to arrays) before we come to the embedding dispatch. this PR solves that issue.


kellertuer:
Nevermind. Properly defining `embed` for the types is the even better way to go.

PR #116, opened 2022-07-08
Manifold-dependent allocation

mateuszbaran:
Another point from https://github.com/JuliaManifolds/Manifolds.jl/issues/497 .

### DIFF
--- a/src/ManifoldsBase.jl
+++ b/src/ManifoldsBase.jl
@@ -22,8 +22,6 @@ import LinearAlgebra: dot, norm, det, cross, I, UniformScaling, Diagonal
 import Markdown: @doc_str
 using LinearAlgebra
 
-using StaticArraysCore: SArray
-
 include("maintypes.jl")
 include("numbers.jl")
 include("bases.jl")
--- a/src/ManifoldsBase.jl
+++ b/src/ManifoldsBase.jl
@@ -73,7 +73,7 @@ end
 allocate(::AbstractManifold, a, dims::Tuple) = allocate(a, dims)
 allocate(::AbstractManifold, a, T::Type) = allocate(a, T)
 function allocate(::AbstractManifold, a, T::Type, dim1::Integer, dims::Integer...)
-    return allocate(a, T, dims1, dims...)
+    return allocate(a, T, dim1, dims...)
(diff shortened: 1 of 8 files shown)


mateuszbaran:
Ah, `StaticArraysCore` is not available for Julia 1.0. So now I'm not sure where to put
```julia
function allocate(::PowerManifoldNestedReplacing, x::AbstractArray{<:SArray})
    return similar(x)
end
```

mateuszbaran:
Some parts of nested work with plain base but we extend it in Manifolds.jl anyway. I think I'll do just that.

src/ManifoldsBase.jl
```diff
@@ -36,26 +36,46 @@ include("projections.jl")
...
 allocation and is forwarded to the function `similar`.
+
+It's behavior can be overriden by a specific manifold, for example power manifold with
+nested replacing representation can decide that `allocate` for `Array{<:SArray}` returns
+another `Array{<:SArray}` instead of `Array{<:MArray}`, as would be done by default.
 """
 allocate(a, args...)
 allocate(a) = similar(a)
-allocate(a, dims::Integer...) = similar(a, dims...)
+allocate(a, dim1::Integer, dims::Integer...) = similar(a, dim1, dims...)
```
kellertuer, line 59: Why the `dim1`? that is not documented above.

mateuszbaran: It's there to guarantee that this specialization only applies to calls with at least one integer, it's essentially an implementation detail. I'm not even sure if entirely necessary but at one point it showed up in an ambiguity list.

kellertuer: But aren't the integers anyways collected in `dims...`?

mateuszbaran: Yes, though the difference is that `allocate(a, dims::Integer...)` can be called without any integers (`allocate([1])`) while `allocate(a, dim1::Integer, dims::Integer...)` ensures that there is at least one integer.

kellertuer: ok, if you feel that thats necessary.

kellertuer (approved):

PR #119, opened 2022-07-25
Inverse retraction based distance approximation

mateuszbaran:
Required to complete https://github.com/JuliaManifolds/Manopt.jl/pull/138 .


kellertuer (approved):
I have not yet seen this in literature but it seems nice.
In the PR you reference I think one could also directly use the norm.

mateuszbaran:
What I think is nice about this is that we don't have to explicitly compute the inverse retraction if a manifold provides a faster implementation for a particular case.

PR #120, opened 2022-07-29
Better default for get_basis coordinate eltype

mateuszbaran:
Just a small enhancement to make it possible to use this `get_basis` for quaternionic unitary matrices (see https://github.com/JuliaManifolds/Manifolds.jl/pull/511).

### DIFF
--- a/src/bases.jl
+++ b/src/bases.jl
@@ -319,6 +319,21 @@ function combine_allocation_promotion_functions(::typeof(identity), ::typeof(com
     return complex
 end
 
+"""
+    coordinate_eltype(M::AbstractManifold{M𝔽}, p, 𝔽::AbstractNumbers) where {M𝔽}
+
+Get the element type for 𝔽-field coordinates of the tangent space at a point `p` from
+manifold `M`. This default assumes that usually complex bases of complex manifolds have
+real coordinates but it can be overridden by a more specific method.
+"""
+@inline function coordinate_eltype(::AbstractManifold{M𝔽}, p, 𝔽::AbstractNumbers) where {M𝔽}
+    if M𝔽 === 𝔽
+        return real(number_eltype(p))
+    else
+        return number_eltype(p)
+    end
+end
+
 @doc raw"""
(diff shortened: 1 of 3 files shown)


src/bases.jl
```diff
@@ -444,10 +444,26 @@ function get_basis_diagonalizing end
 function _get_basis(M::AbstractManifold, p, B::DefaultOrthonormalBasis)
     return get_basis_orthonormal(M, p, number_system(B))
 end
+
```
kellertuer, line 447: I think this should be the same as the `eltype` for tangent vectors? If we do have that, that is.
So if we don't I would maybe prefer to introduce that?

mateuszbaran: How would you use `eltype` for tangent vectors? This is `eltype` for coordinates. In the current Quaternion PR I got quaternions as coordinates where real numbers were expected and this is supposed to fix the issue.

kellertuer: Ah then I find the name misleading, since I thought it is the eltype used for the tangent vectors not the coordinates.

So if possible I would prefer these functions for points/vectors/coordinates? Then basis can use the ones it needs (for vectors/coordinates).

mateuszbaran: BTW, the Quaternionic Unitary manifold is a counterexample, eltypes of vectors and coordinates are different (vectors are quaternions while coordinates are real numbers).

kellertuer: that is why I propose to use three eltype functions :)

mateuszbaran: It just needs the function for coordinates, other eltypes are determined automatically. We could maybe call it something like `coordinate_eltype`?

mateuszbaran: We don't currently need functions that return point and tangent vector eltypes.

kellertuer: Ok, then let's rename this (because from the name vector/coordinate is not clear). And when we need, we can add the vector eltype function somewhen later).

mateuszbaran:
I think this is ready to be merged now :slightly_smiling_face:

kellertuer (approved):
Now it looks good :)

PR #126, opened 2022-10-02
Add shooting method

sethaxen:
This PR ports the shooting method code from https://github.com/JuliaManifolds/Manifolds.jl/pull/535. It slightly redesigns the API to better align with how the other (inverse) retractions are implemented here. The code lives in its own file, since it needs access to the `AbstractVectorTransport` type, which isn't available until after `retractions.jl` is imported.

I'm not certain the best way to test this. I could use `DefaultManifold`, but Euclidean is too easy of a case. Is there anything like a minimal sphere implementation in the test suite?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "ManifoldsBase"
 uuid = "3362f125-f0bb-47a3-aa74-596ffd7ef2fb"
 authors = ["Seth Axen <seth.axen@gmail.com>", "Mateusz Baran <mateuszbaran89@gmail.com>", "Ronny Bergmann <manopt@ronnybergmann.net>", "Antoine Levitt <antoine.levitt@gmail.com>"]
-version = "0.13.19"
+version = "0.13.20"
 
 [deps]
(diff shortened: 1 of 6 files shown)


kellertuer:
What do you mean by “too easy”? I think we can only check against the DefaultManifold here.

Unless you define the few function you need on a DefaultSphere locally in the test suite for the shooting method?

sethaxen:
Too easy because any inverse retraction is as good as the actual logarithm on Euclidean, so the initial inverse retraction will solve the problem and there are branches of the shooting code we won't hit.

I think we do need something like a sphere.

kellertuer:
...otherwise you could come up with a retraction that is not exact on the default manifold.

kellertuer:
> Too easy because any inverse retraction is as good as the actual logarithm on Euclidean, so the initial inverse retraction will solve the problem and there are branches of the shooting code we won't hit.
> 
> I think we do need something like a sphere.

I think just the smaller amount of functions you need for a test sphere would be the best idea.

sethaxen:
This is ready for review.

mateuszbaran (approved):

PR #130, opened 2022-10-31
Introduce a `RetractionWithKeywords` to specify / fix keywords on Layer 1

kellertuer:
This is a continuation of https://github.com/JuliaManifolds/Manifolds.jl/issues/547, to introduce retractions (and inverse retractions and vector transports) where you can fix keywords in the retraction types.

This is necessary since keywords only act on layer 3 on the actual implementation – and with this step they are “added” in layer two before “resolving” the retraction type and passing them down to layer 3.

### DIFF
--- a/src/retractions.jl
+++ b/src/retractions.jl
@@ -130,6 +130,31 @@ matrix / matrices for point and tangent vector on a [`AbstractManifold`](@ref)
 struct QRRetraction <: AbstractRetractionMethod end
 
 """
+    RetractionWithKeywords{R<:AbstractRetractionMethod,K} <: AbstractRetractionMethod
+
+Since retractions might have keywords, this type is a way to set them as an own type to be
+used as a specific retraction
+
+## Fields
+
+* `retraction` the retraction that is decorated with keywords
+* `kwargs` th ekeyword arguments
+
+## Constructor
+
+    RetractionWithKeywords(m::T; kwargs...) where {T <: AbstractRetractionMethod}
+
+
+"""
+struct RetractionWithKeywords{T<:AbstractRetractionMethod,K} <: AbstractRetractionMethod
+    retraction::T
+    kwargs::K
+end
+function RetractionWithKeywords(m::T; kwargs...) where {T<:AbstractRetractionMethod}
+    return RetractionWithKeywords{T,typeof(kwargs)}(m, kwargs)
+end
+
+"""
     SoftmaxRetraction <: AbstractRetractionMethod
 
 Describes a retraction that is based on the softmax function.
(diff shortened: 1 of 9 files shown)


kellertuer:
The retraction case should be finished (including tests); If this is a way we want to go, the other two are not much work to do, but I would like to be sure that this is a good approach.

mateuszbaran:
I think this new retraction looks fine, this is a good approach.

mateuszbaran:
Sorry for closing, I've accidentally clicked the wrong button.

kellertuer:
So this should be feature complete, but I have to take a look how to test the final variant, the `VectorTransportWithKeywords`.

src/retractions.jl
```diff
@@ -129,6 +129,35 @@ matrix / matrices for point and tangent vector on a [`AbstractManifold`](@ref)
 """
 struct QRRetraction <: AbstractRetractionMethod end
 
+"""
+    RetractionWithKeywords{R<:AbstractRetractionMethod,K} <: AbstractRetractionMethod
+
+Since retractions might have keywords, this type is a way to set them as an own type to be
+used as a specific retraction.
+Another reason for this type is that we dispatch on the retraction first and only the
+last layer would be implemented with keywords, so this way they can be passed down
```
mateuszbaran, line 138:
```suggestion
last layer would be implemented with keywords, so this way they can be passed down.
```

src/retractions.jl
```diff
@@ -129,6 +129,35 @@ matrix / matrices for point and tangent vector on a [`AbstractManifold`](@ref)
...
+
+## Fields
+
+* `retraction` the retraction that is decorated with keywords
+* `kwargs` the keyword arguments
+
+Note that you can nest this type. Then the most outer specification of a keyword wins.
```
mateuszbaran, line 145:
```suggestion
Note that you can nest this type. Then the most outer specification of a keyword is used.
```

src/retractions.jl
```diff
@@ -275,6 +304,40 @@ function NLSolveInverseRetraction(
...
 end
 
+
+"""
+    InverseRetractionWithKeywords{R<:AbstractRetractionMethod,K} <: AbstractRetractionMethod
+
+Since inverse retractions might have keywords, this type is a way to set them as an own type to be
+used as a specific inverse retraction.
+Another reason for this type is that we dispatch on the inverse retraction first and only the
+last layer would be implemented with keywords, so this way they can be passed down
```
mateuszbaran, line 314:
```suggestion
last layer would be implemented with keywords, so this way they can be passed down.
```

src/retractions.jl
```diff
@@ -275,6 +304,40 @@ function NLSolveInverseRetraction(
...
+
+## Fields
+
+* `inverse_retraction` the inverse retraction that is decorated with keywords
+* `kwargs` the keyword arguments
+
+Note that you can nest this type. Then the most outer specification of a keyword wins.
```
mateuszbaran, line 321:
```suggestion
Note that you can nest this type. Then the most outer specification of a keyword is used.
```

src/retractions.jl
mateuszbaran: Since this feature is documented, it should also be tested.

src/vector_transport.jl
```diff
@@ -282,6 +280,39 @@ struct VectorTransportTo{
...
+used as a specific vector transport.
+Another reason for this type is that we dispatch on the vector transport first and only the
+last layer would be implemented with keywords, so this way they can be passed down.
+
+## Fields
+
+* `vector_transport` the vector transport that is decorated with keywords
+* `kwargs` the keyword arguments
+
+Note that you can nest this type. Then the most outer specification of a keyword wins.
```
mateuszbaran, line 296:
```suggestion
Note that you can nest this type. Then the most outer specification of a keyword is used.
```

mateuszbaran (commented):
I've looked through the PR and I think it only needs tests to be finished.

kellertuer:
I checked the missing lines - the one case (mid point on <=1.1) is called so that is a false positive; where I am not sure whether maybe the 1.1 just did not report properly.

The other – on ValidationManifolds, I checked and hopefully fixed.

mateuszbaran (approved):
OK, looks fine :+1: .

PR #133, opened 2022-11-04
remove unused type variables

### LINKED ISSUE #132: Warnings for unused type variables.
In Julia 1.8.2 we now get a few warnings for unused types in the fallbacks, for example

```
WARNING: method definition for allocate at ManifoldsBase.jl/src/point_vector_fallbacks.jl:27 declares type variable _ but does not use it.
WARNING: method definition for allocate at ManifoldsBase.jl/src/point_vector_fallbacks.jl:28 declares type variable _ but does not use it.
```

wich is a little hard to read since it is in a macro, but I think we have to adapt the `$Twhere` to be meow flexible? Maybe a point and a vector type specifically?

mateuszbaran:
Resolves #132. Manifolds.jl has the same issue in a few places which have to be resolved in another PR.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "ManifoldsBase"
 uuid = "3362f125-f0bb-47a3-aa74-596ffd7ef2fb"
 authors = ["Seth Axen <seth.axen@gmail.com>", "Mateusz Baran <mateuszbaran89@gmail.com>", "Ronny Bergmann <manopt@ronnybergmann.net>", "Antoine Levitt <antoine.levitt@gmail.com>"]
-version = "0.13.23"
+version = "0.13.24"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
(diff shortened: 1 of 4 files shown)


mateuszbaran:
The one new uncovered line looks like a false positive.

kellertuer:
Interesting approach! I had thought we would just split Twhere into Point and TVector types and put them were necessary – but that would be breaking. Is this one a breaking change?

kellertuer:
Besides that, allocate has an unused variable S in there (I removed it on my branch) but you could remove it here as well since it fits this PR very well.

mateuszbaran:
> Interesting approach! I had thought we would just split Twhere into Point and TVector types and put them were necessary – but that would be breaking. Is this one a breaking change?

The problem is we sometimes need the `where` clause and sometimes we don't, and sadly I couldn't find anything that would interpolate to "this is empty" in type variable list after `where` (which would simplify things significantly). Luckily this is not breaking.

> Besides that, allocate has an unused variable S in there (I removed it on my branch) but you could remove it here as well since it fits this PR very well.

Sure, good idea. Weird that it didn't give me a warning.

kellertuer:
Great, because mine would have been breaking.

Though I think in your case there might now be hidden unused variables, since Twhere still contains possibly a point and a vector type but there are functions that use only one of them. I am surprised those do not get recognised as warnings.

kellertuer (approved):
Sad that we seem to get 2 more false positives in the test but the rest looks fine.

PR #134, opened 2022-11-05
Introduce more error modes

### LINKED ISSUE #131: More informative is approx and a non failing/informative ìs_point / is_vector
Since I am currently restructuring our tests, I noticed two things:

1. It is great that `is_point` provides information why is returns false (if activated), but one could maybe even make it non failing?

My idea would be something like

* `is_point(M, p  true)`stays as is to be non-breaking but is the same as
* `is_point(M, p, :error)` throws an error if `p` is not a point on `M`
* `is_point(M, p, :warn)` displays a warning – the test would fail but the test set would continue

Sure symbols are not the best choice here, let me know if you have a better idea how to distinguish both cases 

2. `isapprox(M,p,q)` and `is approx(M,p,X,Y)` are quite hard to read when they fail, so could we maybe extend those functions to act similar to `is_point` that is an error message can be activated that checks first whether it was atol or rtol which made the test fail and (similar to above) either errors or warns? This could be a last positional argument as well for both, I think.

I also started a discussion at https://discourse.julialang.org/t/isapprox-in-tests-and-a-more-precise-result-reason-for-failure/89648 – but then remembered, that we are defining our own `isapprox`es anyways, so with the `M` upfront we can add the new positional argument without “committing type piracy”.

Let me know what you think.

kellertuer:
I.e. that the following methods have symbol positional arguments 

* [x] `is_point`
* [x] `is_vector`
* [x] `isapprox` (with manifolds)
* [x] these are tested (and ambiguities are resolved)

This resolves #131.

I am not yet sure how to do the tests (I am not able to use `@test_logs` here and the actual info/warn string for `isapprox` seems to be a little difficult still.

_edit:_ Oh and I introduced quite a few ambiguities, all but 3 I could resolve, those are special ones with the embedding and might be a little tricky.

### DIFF
--- a/src/ManifoldsBase.jl
+++ b/src/ManifoldsBase.jl
@@ -169,7 +169,7 @@ function check_approx(M::AbstractManifold, p, q; kwargs...)
     s = "The two points $p and $q on $M are not (approximately) equal."
     return ApproximatelyError(s)
 end
-function check_approx(M, p, X, Y; kwargs...)
+function check_approx(M::AbstractManifold, p, X, Y; kwargs...)
     # fall back to classical mode - just that we do not have a reason then
     res = isapprox(X, Y; kwargs...)
     res && return nothing
--- a/src/ManifoldsBase.jl
+++ b/src/ManifoldsBase.jl
@@ -441,8 +441,17 @@ inner(M::AbstractManifold, p, X, Y)
 
 """
     isapprox(M::AbstractManifold, p, q; kwargs...)
+    isapprox(M::AbstractManifold, p, q, error::Symbol; kwargs...)
 
 Check if points `p` and `q` from [`AbstractManifold`](@ref) `M` are approximately equal.
+
+The optional positional argument can be used to get more information for the case that
+the result is false, if the concrete manifold provides such information.
+Currently the following are supported
+* `:error` - throws an error if `isapprox` evaluates to false, providing possibly a more detailed error.
+  Note that this turns `isapprox` basically to an `@assert`.
+* `:info` – prints the information in an `@info`
+* `:warning` – prints the information in an `@warn`
 
 Keyword arguments can be used to specify tolerances.
 """
@@ -450,13 +459,29 @@ isapprox(::AbstractManifold, x, y; kwargs...) = isapprox(x, y; kwargs...)
 
 """
     isapprox(M::AbstractManifold, p, X, Y; kwargs...)
+    isapprox(M::AbstractManifold, p, X, Y, error:Symbol; kwargs...)
 
 Check if vectors `X` and `Y` tangent at `p` from [`AbstractManifold`](@ref) `M` are approximately
 equal.
 
+The optional positional argument can be used to get more information for the case that
+the result is false, if the concrete manifold provides such information.
+Currently the following are supported
+
+* `:error` - throws an error if `isapprox` evaluates to false, providing possibly a more detailed error.
+  Note that this turns `isapprox` basically to an `@assert`.
+* `:info` – prints the information in an `@info`
+* `:warning` – prints the information in an `@warn`
+
+By default these informations are collected by calling [`check_approx`](@ref).
+
 Keyword arguments can be used to specify tolerances.
 """
 isapprox(::AbstractManifold, p, X, Y; kwargs...) = isapprox(X, Y; kwargs...)
+
+function isapprox(M::AbstractManifold, p, X, Y, error::Symbol; kwargs...)
+
+end
 
 
 """
--- a/src/ManifoldsBase.jl
+++ b/src/ManifoldsBase.jl
@@ -450,7 +450,7 @@ Currently the following are supported
 * `:error` - throws an error if `isapprox` evaluates to false, providing possibly a more detailed error.
   Note that this turns `isapprox` basically to an `@assert`.
 * `:info` – prints the information in an `@info`
-* `:warning` – prints the information in an `@warn`
+* `:warn` – prints the information in an `@warn`
 * `:none` (default) – the function just returns `true`/`false`
 
 Keyword arguments can be used to specify tolerances.
@@ -465,7 +465,7 @@ function isapprox(M::AbstractManifold, p, q; error::Symbol = :none, kwargs...)
             s = "$(typeof(ma)) with $(ma.val)\n$(ma.msg)"
         end
         (error === :info) && @info s
-        (error === :warning) && @warn s
+        (error === :warn) && @warn s
         return false
     end
     return true
@@ -494,7 +494,7 @@ Currently the following are supported
 * `:error` - throws an error if `isapprox` evaluates to false, providing possibly a more detailed error.
   Note that this turns `isapprox` basically to an `@assert`.
 * `:info` – prints the information in an `@info`
-* `:warning` – prints the information in an `@warn`
+* `:warn` – prints the information in an `@warn`
 * `:none` (default) – the function just returns `true`/`false`
 
 By default these informations are collected by calling [`check_approx`](@ref).
@@ -511,7 +511,7 @@ function isapprox(M::AbstractManifold, p, X, Y; error::Symbol = :none, kwargs...
             s = "$(typeof(mat)) with $(mat.val)\n$(mat.msg)"
         end
         (error === :info) && @info s
-        (error === :warning) && @warn s
+        (error === :warn) && @warn s
         return false
     end
     return true
@@ -540,7 +540,7 @@ is `nothing` or an error.
 A more precise way can be set using a symbol as the optional parameter, where
 ' `:error` is the same as setting `throw_error=true`
 ' `:info` displays the error message as an `@info`
-* `:warning` displays the error message as a `@warning`
+* `:warn` displays the error message as a `@warning`
 
 all other symbols are equivalent to `throw_error=false`.
 """
@@ -561,14 +561,14 @@ function is_point(M::AbstractManifold, p, error::Symbol; kwargs...)
     if mps !== nothing
         s = "$(typeof(mps)) with $(mps.val)\n$(mps.msg)"
         (error === :info) && @info s
-        (error === :warning) && @warn s
+        (error === :warn) && @warn s
         return false
     end
     mpe = check_point(M, p; kwargs...)
     if mpe !== nothing
         s = "$(typeof(mpe)) with $(mpe.val)\n$(mpe.msg)"
         (error === :info) && @info s
-        (error === :warning) && @warn s
+        (error === :warn) && @warn s
         return false
     end
     return true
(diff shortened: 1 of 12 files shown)


mateuszbaran:
No problem, I've fixed the ambiguities :slightly_smiling_face: .

kellertuer:
ah, stupid me, sure the bool has to be more unambiguous then as well :) Thanks for the help.

I will think next about the string to display for the Isapprox and after that check for the tests; I did some info/warn checks in Manopt but for now I was not able to adapt them.

kellertuer:
I got a little stuck thinking about the `isapprox` function here, because different from `is_point` where the internal `check_point` delivers a message, the lower level (undecorated or implemented at concrete manifold level) `isapprox`es not only do their own isapprox implementation (which we can not recreate on the abstract level afterwards to check what the reason was), they also do not return any reasoning. So for the abstract case it might be hard to provide @info/@warn things, for the concrete cases it would be a lot of updates in Manifolds.jl

We could change to an internal `is_approx` (similar to the internal `check_point`( which would mean not so much change in Manifolds but still a change (to introduce reasons).

Also I still did not yet get tests for @info/@warn to work today either.

mateuszbaran:
> We could change to an internal `is_approx` (similar to the internal `check_point`( which would mean not so much change in Manifolds but still a change (to introduce reasons).

I think that's a reasonable direction. Maybe `check_approx` would be a good name for that new function?

kellertuer:
`check_approx` is a very good name idea, this way we could introduce that as nonbreaking  here (have an idea already) and update the functions over on Manifolds step by step.

kellertuer:
I sketched `check_approx` including a fallback to the (old/previous) `isapprox` so that the default is backwards-compatible (but we can update them to `check_approx`es to provide more information. I am just wondering: Should we do a new Error Type for that? I feel `AssertionError` is too vague and would maybe define `ApproxError` or something like that – maybe even two – one for points one for tangents. Any ideas for good names?

mateuszbaran:
Sure, we can have a special error type for that. `ApproxError` would be fine.

kellertuer:
Uh, the is approx has a large ambiguity problem. Compare

`isapprox(M, p, q, error::Symbol)`
and
`isapprox(M::PowerManifold, p, X, Y)`

I do not yet know how to resolve that. Probably we can only do the error as a keyword? This would be inconsistent to the other ones and changing all to keywords would be a breaking change.

mateuszbaran:
Yes, that's an issue. There is no good way to resolve such ambiguity and `error` needs to be a keyword argument. It's inconsistent but we can live with that, and add to the list of changes for the next breaking release.

kellertuer:
Ok, that sounds like an idea.

kellertuer:
Now my only problem left is that currently the (old) default fallback results in a stack overflow since it is “running in circles”. I have to check how to best resolve that since we do not have a new positional argument to dispatch on for the new – more informative – case.

kellertuer:
I think I will also try to remove the 4 warnings the tests currently print by using `@test_logs (:warn,) ...` if I can get that to work.
Besides Tests this should be ready for review.

src/ManifoldsBase.jl
```diff
@@ -440,34 +440,110 @@ Compute the inner product of tangent vectors `X` and `Y` at point `p` from the
...
+By default these informations are collected by calling [`check_approx`](@ref).
+
 Keyword arguments can be used to specify tolerances.
 """
-isapprox(::AbstractManifold, p, X, Y; kwargs...) = isapprox(X, Y; kwargs...)
+function isapprox(M::AbstractManifold, p, X, Y; error::Symbol = :none, kwargs...)
+    mat = check_approx(M, p, X, Y; kwargs...)
+    if mat !== nothing
+        (error === :error) && throw(mat)
+        if isnan(mat.val)
```
mateuszbaran, line 508: I think we could always print `val` even if it's `NaN` for consistency?

src/ManifoldsBase.jl
```diff
@@ -440,34 +440,110 @@ Compute the inner product of tangent vectors `X` and `Y` at point `p` from the
...
+        return false
+    end
+    return true
+end
+
+function check_approx(M, p, X, Y; kwargs...)
+    # fall back to classical mode - just that we do not have a reason then
+    res = isapprox(X, Y; kwargs...)
+    res && return nothing
+    v = norm(X - Y)
```
mateuszbaran, line 524:
```suggestion
    v = norm(M, p, X - Y)
```
I think that's more correct in general.

src/ManifoldsBase.jl
```diff
@@ -480,8 +556,29 @@ function is_point(M::AbstractManifold, p, throw_error = false; kwargs...)
...
+        (error === :warning) && @warn s
+        return false
+    end
+    return true
+end
+
+
 """
     is_vector(M::AbstractManifold, p, X, throw_error = false, check_base_point=true; kwargs...)
+    is_vector(M::AbstractManifold, p, X, error::Symbol, check_base_point=true; kwargs...)
```
mateuszbaran, line 581:
```suggestion
    is_vector(M::AbstractManifold, p, X, error::Symbol, check_base_point::Bool=true; kwargs...)
```
Since we are working on this part anyway, we could add `::Bool` here to be more precise.

test/validation_manifold.jl
```diff
@@ -35,6 +35,12 @@ end
         @test ManifoldsBase.check_size(A, x2, v2) === ManifoldsBase.check_size(M, x, v)
         @test ManifoldsBase.check_size(A, x2) === ManifoldsBase.check_size(M, x)
     end
+    @testset "Is_point / is_vector error." begin
```
mateuszbaran, line 38:
```suggestion
    @testset "is_point / is_vector error." begin
```

mateuszbaran (commented):
I think this looks quite nice, I have just a few minor points.

src/ManifoldsBase.jl
kellertuer: My fear is that the nan might confuse a used

src/ManifoldsBase.jl
kellertuer: nice idea.

kellertuer: The problem with this one is that on the manifold where we can properly test this fallback, said norm is not defined

mateuszbaran: What is the problem with defining that norm?

kellertuer: The manifold is called `NonManifold` and we test all MethodErrors with that.

src/ManifoldsBase.jl
mateuszbaran: It might, but so might lack of any number. How about changing the message to something like `"$(typeof(ma)) (value: $(ma.val))\n$(ma.msg)"`?

src/ManifoldsBase.jl
mateuszbaran: But then we can just use a different manifold to test `check_approx`?

kellertuer: Sure, we can implement a whole new Manifold with that. All just a question of effort (compared to deleting one line).

mateuszbaran: I can cover that with a test then.

src/ManifoldsBase.jl
kellertuer: Then I still find `Value: NaN` irritating to the user, since it is actually not `NaN` but absence of information (at least information that can be provided).

mateuszbaran: OK, then let's leave this as is.

src/ManifoldsBase.jl
kellertuer: Let me just first finish to do the remaining coverage – maybe I just saw one where that can easily be added as well (in domain error we might have a fitting manifold).

My conceptional problem is, that I would like to provide errors without much stuff present, here we would say: For errors to work, please implement a norm, otherwise we error while we error.

mateuszbaran: Hm, we can always wrap norm computation in try/catch then.

kellertuer: ...or keep it simple. I do not have a strong preference here, but I would either use `norm(X-Y)` or not provide a number (since there is not good default to do so).

mateuszbaran: To me the only good default is `norm(M, p, X - Y)` (if it's implemented) so I prefer no number over `norm(X-Y)` (which may be sometimes different from `norm(M, p, X - Y)`).

kellertuer:
Note that I changed one thing in the (old) tests: Those tests that threw warnings are now checked to issue the warning we expect.

src/ManifoldsBase.jl
kellertuer: You are right the actual norm might be misleading especially on quotient.
So we either
a) provide no number
or
b) try/catch norm and provide a number if that suceeds, otherwise no number.

I tend little bit to a) for simplicity, but b) might be nicer in information sometimes (and when we error, performance should not be that much of a problem). You tend to b)?

mateuszbaran: Yes, I prefer b) because it gives additional information which should always be correct but I'm also fine with a). Performance indeed should not be a problem here, this method should not be used in performance-critical code.

kellertuer: Ok, I implemented b).

mateuszbaran: Thanks!

src/ManifoldsBase.jl
```diff
@@ -144,6 +144,44 @@ level, whether its decorated or not. Any negative value deactivates this depth l
...
+"""
+function check_approx(M::AbstractManifold, p, q; kwargs...)
+    # fall back to classical approx mode - just that we do not have a reason then
+    res = isapprox(p, q; kwargs...)
+    # since we can not assume distance to be implemented, we can not provide a default value
+    res && return nothing
+    s = "The two points $p and $q on $M are not (approximately) equal."
+    return ApproximatelyError(s)
+end
+function check_approx(M, p, X, Y; kwargs...)
```
mateuszbaran, line 172:
```suggestion
function check_approx(M::AbstractManifold, p, X, Y; kwargs...)
```

src/ManifoldsBase.jl
```diff
@@ -144,6 +144,44 @@ level, whether its decorated or not. Any negative value deactivates this depth l
...
+with the difference that is returns an [`ApproximatelyError`](@ref) if the two elements are
+not approximately equal, containting a more detailed description/reason.
+If the two elements are approximalely equal, this method returns `nothing`.
+
+This method is an internal function and is called by `isapprox` whenever the user specifies
+an `error=` keyword therein.
+"""
+function check_approx(M::AbstractManifold, p, q; kwargs...)
+    # fall back to classical approx mode - just that we do not have a reason then
+    res = isapprox(p, q; kwargs...)
```
mateuszbaran, line 166: Should it be `isapprox(p, q)` or `isapprox(M, p, q)`?

mateuszbaran (commented):
I have two last small points, everything else looks good :slightly_smiling_face: .

src/ManifoldsBase.jl
kellertuer: Well, `isapprox(M, p, q)` internally calls `check_approx`, so that would not work.

This call is similar to the default we had before for `isapprox(M,p,q)` which also called this.

mateuszbaran: OK, I see.

mateuszbaran (approved):

PR #137, opened 2022-11-22
More method specializations for `allocate_result`

mateuszbaran:
We have an unfortunate side-effect of our trait system. Method recursion prevents some type inference from happening around, for example, `allocate_result`. This can cause massive slow-downs, for example the current Manifolds.jl does 32 allocations to copy a point on `SpecialOrthogonal(3)`. This PR solves this partially -- with these changes it's now just 4 -- but we will have to introduce specialized methods to work around inefficiencies in some key parts. I don't really see how to improve it further generically.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "ManifoldsBase"
 uuid = "3362f125-f0bb-47a3-aa74-596ffd7ef2fb"
 authors = ["Seth Axen <seth.axen@gmail.com>", "Mateusz Baran <mateuszbaran89@gmail.com>", "Ronny Bergmann <manopt@ronnybergmann.net>", "Antoine Levitt <antoine.levitt@gmail.com>"]
-version = "0.13.25"
+version = "0.13.26"
 
 [deps]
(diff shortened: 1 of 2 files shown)


kellertuer:
Interesting; I do not understand enough of the implications here, I fear.

mateuszbaran:
I don't completely understand it either. I've just tried benchmarking `copy` again and it went to 1 allocation :thinking: . Julia's optimizer can sometimes be surprisingly nondeterministic.

kellertuer:
Maybe we could even ask that on discourse whether someone can take a look and give us a hint where we could improve that?

mateuszbaran:
I just did that :smiley: https://discourse.julialang.org/t/extremely-slow-invoke-when-inlined/90665

mateuszbaran:
That is, my question is for another issue which is even less clear to me. This thing with `allocate_result` is less troubling for me.

kellertuer:
But cool that you work on these technical details :)

mateuszbaran:
I just like when things work fast :slightly_smiling_face:

kellertuer:
That is the same for me, but I do not have learned about all the technical details you always figure out ;)

PR #138, opened 2022-11-22
Don't inline `invoke` in a generic fallback.

mateuszbaran:
Workaround for https://discourse.julialang.org/t/extremely-slow-invoke-when-inlined/90665 .

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,6 +13,9 @@ jobs:
       matrix:
         julia-version: ["1.0", "1.6", "1.7", "1.8"]
         os: [ubuntu-latest, macOS-latest, windows-latest]
+        exclude:
+          - os: macOS-latest
+            julia-version: "1.0"
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


mateuszbaran:
I will tag 0.13.26 after merging this.

mateuszbaran:
Ref. https://github.com/JuliaLang/julia/pull/47671

kellertuer:
Nice! You even managed to get an update to Julia with that. Super!

kellertuer (approved):

mateuszbaran:
CI keeps failing on Julia 1.0/macOS without a clear error message :confused: .

kellertuer:
How general where `Varargs` possible in Julia 1.0?

mateuszbaran:
Since the code works on Ubuntu and the Vararg change passed CI earlier, I'd say that it's certainly not the issue here.

kellertuer:
Oh, yes, I did not notice that; hm, to me the message looks a little bit like Pkg not being able to get the package, at least on windows I had a similar message once that vanished after a while.

PR #139, opened 2022-11-25
Check for having to call `get_basis` on a basis

mateuszbaran:
This is the thing I wrote about here: https://github.com/JuliaManifolds/Manopt.jl/pull/167#issue-1453906466 .


kellertuer:
Will that already be used here in Base somewhere or is it just useful over on the LM-Alg?

mateuszbaran:
This can be useful for any algorithm that needs to call `get_vector` or `get_coordinates` on arbitrary bases. We just didn't really care before because usually `DefaultOrthonormalBasis` works fine.

kellertuer:
Cool. I was just wondering whether it might even be useful somewhere in here already?

mateuszbaran:
Maybe it could already be useful for something, I haven't given it much thought.

PR #141, opened 2022-12-31
Introduce rand(M) and rand!(M, p) to the API

kellertuer:
This PR adapts the existing methods from Manifolds.jl to ManifoldsBase.jl.

Things to check / discuss
* [x] should we make both functions aware of decorators? Especially for the `rand(rng, M)` variants this might be challenging, but I thing that might be cool to have.
* [x] Should we also provide a default implementation for the Power manifold?
* [x] ~Should we also provide an implementation in the embedding (rand in embedding & project)?~

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "ManifoldsBase"
 uuid = "3362f125-f0bb-47a3-aa74-596ffd7ef2fb"
 authors = ["Seth Axen <seth.axen@gmail.com>", "Mateusz Baran <mateuszbaran89@gmail.com>", "Ronny Bergmann <manopt@ronnybergmann.net>", "Antoine Levitt <antoine.levitt@gmail.com>"]
-version = "0.13.27"
+version = "0.13.28"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
--- a/Project.toml
+++ b/Project.toml
@@ -6,6 +6,7 @@ version = "0.13.27"
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Markdown = "d6f4376e-aef5-505a-96c1-9c027394607a"
+Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 
 [compat]
 DoubleFloats = ">= 0.9.2"
--- a/src/DefaultManifold.jl
+++ b/src/DefaultManifold.jl
@@ -132,6 +132,21 @@ function parallel_transport_to!(::DefaultManifold, Y, p, X, q)
     return copyto!(Y, X)
 end
 
+function Random.rand!(::DefaultManifold, pX; σ = one(eltype(pX)), vector_at = nothing)
+    pX .= randn(size(pX)) .* σ
+    return pX
+end
+function Random.rand!(
+    rng::AbstractRNG,
+    ::DefaultManifold,
+    pX;
+    σ = one(eltype(pX)),
+    vector_at = nothing,
+)
+    pX .= randn(rng, size(pX)) .* σ
+    return pX
+end
+
 function riemann_tensor!(::DefaultManifold, Xresult, p, X, Y, Z)
     return fill!(Xresult, 0)
 end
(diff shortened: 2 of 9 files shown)


mateuszbaran:
Nice, I think this is a good idea.

> * should we make both functions aware of decorators? Especially for the `rand(rng, M)` variants this might be challenging, but I thing that might be cool to have.

I think this can wait, especially for the variant with `rng`. That would be quite complicated to support.

> * Should we also provide a default implementation for the Power manifold?

Sure, you can copy it from Manifolds.jl.

> * Should we also provide an implementation in the embedding (rand in embedding & project)?

That would require making `rand` aware of decorators, which would be quite a lot of work.

kellertuer:
Ok, we can add the power part here as well;

For the decorator-aware part – for the one starting with `M` that should be simple, for the ones starting with the rin, that is more work, I think, but also worth it if you think of the SPDs or something similar with a lot of metrics? Randomness – at least in my mind – is independent of the metric.

kellertuer:
Uff, I took. a short look at the Power rands – that is a hundred lines of code! Some inheriting from Random (which we have here already) some from Distributions (which we don't) and probably also for several different power styles. I will have to read this hundred lines very very very very carefully to check which ones we can copy here and which we can't.

mateuszbaran:
> For the decorator-aware part – for the one starting with `M` that should be simple, for the ones starting with the rin, that is more work, I think, but also worth it if you think of the SPDs or something similar with a lot of metrics? Randomness – at least in my mind – is independent of the metric.

Randomness isn't entirely independent of the metric (if you care about probability density) but I get your point. My only concern is that it may be complicated.

> Uff, I took. a short look at the Power rands – that is a hundred lines of code! Some inheriting from Random (which we have here already) some from Distributions (which we don't) and probably also for several different power styles. I will have to read this hundred lines very very very very carefully to check which ones we can copy here and which we can't.

I am somewhat familiar with that code so I can handle that.

kellertuer:
For the M-upfront variant it should be our normal decorator macro, but you are right, for the other it might be more complicated.

kellertuer:
Would it be a solution to (by default) pass `rand(ring, M, ...)` to a `rand_rngdeco(M, rng...)` that is decorated with the nested trait and its fall back is to call `rand(rng, M_undeco)` on the undecorated manifold again?
That would be a little trick, but maybe the easiest to do? That way one can still implement metric-specific variants for example.

mateuszbaran:
Hm, I've taken a look and it seems that the new trait system shouldn't care that manifold is not the first argument so I'd try decorating without permuting arguments.

kellertuer:
Oh, are you sure? That would be neat, if that just worked. Can you try? You know that system better than me.

kellertuer:
It does not seem to be so easy, just adding

```julia
@trait_function rand(M::AbstractDecoratorManifold; kwargs...)
@trait_function rand!(M::AbstractDecoratorManifold, p; kwargs...)
@trait_function rand(rng::AbstractRNG, M::AbstractDecoratorManifold; kwargs...)
@trait_function rand!(rng::AbstractRNG, M::AbstractDecoratorManifold, p; kwargs...)
```

Even brass the DefaultManifold, and I am not sure why, since it does report 
```julia
MethodError: no method matching rand!(::DefaultManifold{Tuple{3}, ℝ}, ::Vector{Float64})
```

but actually that _is_ implemented in line 135 in DefaultManifold (besides the kwargs not seem to be carried on though I used them above)?

So – probably I am again missing something in our magic decorator.

kellertuer:
Ha! I forgot the prefixes _and_ to import `rand!` – now the tests are running but I am not sure we have coverage yet nor am I sure how to test the decorated ones.

mateuszbaran:
Ah, nice, you've added the dispatch. I will try adding a test using the Pass-through trait.

mateuszbaran:
I had to fix our decorator system a bit but now `rand!` and `rand` should work properly.

kellertuer:
Nice!

mateuszbaran (approved):
I think this is fine now.

mateuszbaran:
Are you working on adapting Manifolds.jl to this small change (primarily removal of the rand methods for PowerManifold)? If not I can do that.

kellertuer:
That would be great, I‘ll be a little more busy with semester preparations starting tomorrow.

PR #143, opened 2023-01-14
Add `is_flat` function

mateuszbaran:
This is needed to correctly implement injectivity radius for tangent bundle, which is then needed to fix Manopt line search on tangent bundle.


mateuszbaran:
Now I just need to figure out for every manifold we have in Manifolds.jl if it's flat.

kellertuer (approved):
While I do not have a clue how this would be used or defined, the definition and docs look fine for me.

mateuszbaran:
It's already used and defined in https://github.com/JuliaManifolds/Manifolds.jl/pull/567 . Is there something unclear?

mateuszbaran:
I just have to go once again over manifolds there because I didn't yet check if some special cases of parametrized manifolds reduce to something flat.

kellertuer:
Ah, then maybe I was expecting more. But we can surely do that later when we have the need for principal curvatures (then this function would check if all of them are zero)

mateuszbaran:
Maybe there will be something more in the future, who knows.

Issue #144, opened 2023-01-16
A (slightly) more precise `default_vector_transport_method`


kellertuer:
I would like to shortly check with you about an extension of the default methods. This came up in JuliaManifolds/Manopt.jl#201

The scenario is as follows:

Assume we have a manifold (like Grassmann) with two representations (Stiefel and Projection),
and that depending on the representation we have different functions implemented. In the given example

* Stiefel has a `ProjectionVectorTransport`
* projection does not have a method (would be the same as before dispatch to ParallelTransport and error)

But it might also be that we want to define different defaults per representation.

# Proposed way

I propose to handle this as follows

1. Introduce a `default_vector_transport_method(M::AbstractManifold, ::Type)` to dispatch on different representation types
2. defaulting this to calling the current one (i.e. ignoring the type) that way we are backwards compatible
3. Introduce a nicer interface by `default_vector_transport_method(M::AbstractManifold; representation=AbstractArray)` having the type as keyword argument.

Let me know what you think, I do not think this is hard to do, just how to best do the interface is a discussion. The same would also be done for retractions and their inverses of course.

kellertuer:
The reason I would like to dispatch on the type and not the point itself is, that the transport should only depend on how `p` is represented (so its type) but never its actual values (i.e. where you are at on `M`).

mateuszbaran:
Sorry for not commenting on this earlier but I don't exactly understand the motivation. Parallel transport on Grassmann is quite expensive computationally so to me it seems unlikely it's ever preferable to projection transport, regardless of point representation.

kellertuer:
It is not about Parallel transport.

The thins is that

`default_vector_transport_method(M)` should return two different things in Grassmann

* _if_ it is a Stiefel/Array representation, then is should return `ProjectionTransport()`
* _if_ it is a ProjectionRepresentation it should not be defined (that is it defaults to PT that is not defined and errors)

In general, a manifold with different representations (take Hyperbolic for example) might have a “favourite” default depending on the representation.

kellertuer:
So if we come up with the `BaranTransport` for Projection representation on Grassman (but do not implement ProjectionTransport, since we do not find a paper for that) then Grassmann with ProjectionRepresentation should default to `BaranTransport()`while we want to keep the Stiefel default as is.

Currently there is no default yet on Grassman, but when I tried to define it, I noticed this difficulty

mateuszbaran:
[...]

PR #145, opened 2023-01-18
Introduce the idea of the more-precise default functions from (#144).

kellertuer:
Let me know what you think.

Maybe `default_X_method(M, typeof(p))` looks a little clumsy, but I feel to emphasise that it should never depend on the **values** of `p` really only on its type. We should not change the default “locally” but based on representation.

The application example from Manifolds.jl is the Grassmann manifold, where we should only set a default for the Array/Stiefel representation (to be `ProjectionRetraction`) but not for the Projection representation, where no vector transport is implemented.

This is non-breaking, since the 2-argument version falls back to the 1-argument one – but making use of this more-precise dispatch requires a bit of patching in Manifolds and Manopt.

### DIFF
--- a/src/PowerManifold.jl
+++ b/src/PowerManifold.jl
@@ -376,6 +376,9 @@ functions defined for the power manifold, meaning that this is used elementwise.
 function default_retraction_method(M::PowerManifold)
     return default_retraction_method(M.manifold)
 end
+function default_retraction_method(M::PowerManifold, t::Type)
+    return default_retraction_method(M.manifold, t)
+end
 
 @doc raw"""
     default_inverse_retraction_method(M::PowerManifold)
@@ -386,6 +389,9 @@ functions defined for the power manifold, meaning that this is used elementwise.
 function default_inverse_retraction_method(M::PowerManifold)
     return default_inverse_retraction_method(M.manifold)
 end
+function default_inverse_retraction_method(M::PowerManifold, t::Type)
+    return default_inverse_retraction_method(M.manifold, t)
+end
 
 @doc raw"""
     default_vector_transport_method(M::PowerManifold)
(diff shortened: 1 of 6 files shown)


src/PowerManifold.jl
```diff
@@ -376,6 +376,9 @@ functions defined for the power manifold, meaning that this is used elementwise.
 function default_retraction_method(M::PowerManifold)
     return default_retraction_method(M.manifold)
 end
+function default_retraction_method(M::PowerManifold, t::Type)
+    return default_retraction_method(M.manifold, t)
```
mateuszbaran, line 380:
```suggestion
    return default_retraction_method(M.manifold, eltype(t))
```
At least for nested representations this should pass `eltype`.

mateuszbaran: Similarly for other power manifold methods.

src/retractions.jl
```diff
@@ -347,20 +347,36 @@ struct SoftmaxInverseRetraction <: AbstractInverseRetractionMethod end
 
 """
     default_inverse_retraction_method(M::AbstractManifold)
+    default_inverse_retraction_method(M::AbstractManifold, ::Type{T}) where {T})
```
mateuszbaran, line 350:
```suggestion
    default_inverse_retraction_method(M::AbstractManifold, ::Type{T}) where {T}
```

mateuszbaran:
Everything else looks fine here :+1: .

src/PowerManifold.jl
kellertuer: Oh, well spotted :) I was not even aware that `eltype` works on array types as well.

kellertuer:
I also already ran downstream tests – Manifolds still works on this – but we can refine in a few places (where we use the default and for Grassmann as discussed).

mateuszbaran (approved):
Other than these two minor things this looks good :slightly_smiling_face: .

src/PowerManifold.jl
```diff
@@ -386,6 +389,9 @@ functions defined for the power manifold, meaning that this is used elementwise.
 function default_inverse_retraction_method(M::PowerManifold)
     return default_inverse_retraction_method(M.manifold)
 end
+function default_inverse_retraction_method(M::PowerManifold, t::Type)
+    return default_inverse_retraction_method(M.manifold, t)
```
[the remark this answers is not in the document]
kellertuer: Ah, I must have missed those two.

PR #146, opened 2023-01-24
Fused retractions

mateuszbaran:
Here I'm working on https://github.com/JuliaManifolds/Manifolds.jl/issues/569 .


kellertuer:
Looking forward to seeing how that turns out. I hope we had good ideas to both improve performance and keeping it nonbreaking :)

mateuszbaran:
I've worked a bit on this today and `retract_pade(M::AbstractManifold, p, X, ::Number)` makes me confused due to the two numbers at the end. Or one. It's hard to notice that it actually is well-defined. I think it would be much more clear if we didn't extract number from `PadeRetraction` there and just passed the `PadeRetraction` object.

kellertuer:
It is just one number, since the pade-retraction has an approximation order `n` that we extract, when we move from level 2 (dispatch on retraction type) to level 3 (retraction-type functions like `retract_pade`.

But sure if you want to go for fusing stuff, then it would have two arguments that are numbers in the end `n::Int, t::Real`.
Note that currently `retract_caley` defaults to `n=1`.

Can you elaborate on the argument why that would not be well-defined? Is it ok if the fix `n` to be integer?

My argument _for_ the current form ist that `retract_pade` that has a `PadeRetraction{n}()` as its last (or with fusion second to last) argument doubles the information that it is a pade-retraction. It also makes the interplay with Caley more complicated. Also the type really just has an integer in its parameters.
If possible I would like to keep the `n` there as a last (non-fused) or second-to-last (fused) argument.

mateuszbaran:
> But sure if you want to go for fusing stuff, then it would have two arguments that are numbers in the end `n::Int, t::Real`.
> Note that currently `retract_caley` defaults to `n=1`.

Why that order? Right now we have `t` right after `X` in `exp` and `retract`, before retraction method.

> Can you elaborate on the argument why that would not be well-defined? Is it ok if the fix `n` to be integer?

As long as `n` doesn't have a default value it is well-defined. If it did have a default value for `n`, we couldn't be sure if the user wanted to specify `n` or `t` by an integer after `X`. We have retraction methods as optional arguments so expecting `n` to maybe have a default value is something I had done until I realized it couldn't work with optional `t` as well.

> My argument _for_ the current form ist that `retract_pade` that has a `PadeRetraction{n}()` as its last (or with fusion second to last) argument doubles the information that it is a pade-retraction. It also makes the interplay with Caley more complicated. Also the type really just has an integer in its parameters.
> If possible I would like to keep the `n` there as a last (non-fused) or second-to-last (fused) argument.

Note that we have different expectations of what `retract_pade(M, p, X, 2, 3)` is supposed to mean. `retract_pade(M, p, X, 2, PadeRetraction(3))` (or `retract_pade(M, p, X, PadeRetraction(3), 2)`) would make the intention clear. We can use a different type instead of `PadeRetraction` if you prefer (like `Val(3)`), just something that is not `<:Number`.

kellertuer:
> > But sure if you want to go for fusing stuff, then it would have two arguments that are numbers in the end `n::Int, t::Real`.
> > Note that currently `retract_caley` defaults to `n=1`.
> 
> Why that order? Right now we have `t` right after `X` in `exp` and `retract`, before retraction method.

I do not have a strong preference here, I was just thinking: It's added at the end – but you are right it's “at the end but before the retraction” and since `n` specifies the retraction your idea of doing `t, n` is also very reasonable. Let's go with your choice then, since mine was really only to add `t` at the end (which is not what we do). Or maybe not, see next point.

> 
> > Can you elaborate on the argument why that would not be well-defined? Is it ok if the fix `n` to be integer?
> 
> As long as `n` doesn't have a default value it is well-defined. If it did have a default value for `n`, we couldn't be sure if the user wanted to specify `n` or `t` by an integer after `X`. We have retraction methods as optional arguments so expecting `n` to maybe have a default value is something I had done until I realized it couldn't work with optional `t` as well.

This might actually be an argument to putting `t` last, since I do not expect `n` to have a default value, it is a _mandatory_ argument of the `retract_pade` and only `t` has a default value. It is mandatory, since it is always a given parameter of `PadeRetraction`.

> > My argument _for_ the current form ist that `retract_pade` that has a `PadeRetraction{n}()` as its last (or with fusion second to last) argument doubles the information that it is a pade-retraction. It also makes the interplay with Caley more complicated. Also the type really just has an integer in its parameters.
> > If possible I would like to keep the `n` there as a last (non-fused) or second-to-last (fused) argument.
> 
> Note that we have different expectations of what `retract_pade(M, p, X, 2, 3)` is supposed to mean. `retract_pade(M, p, X, 2, PadeRetraction(3))` (or `retract_pade(M, p, X, PadeRetraction(3), 2)`) would make the intention clear. We can use a different type instead of `PadeRetraction` if you prefer (like `Val(3)`), just something that is not `<:Number`.

For me, the order is clearer after reading this. the `n` should be the 2 is your first notation, since the `t` is last in the `retract_typeofretraction()` functions in level 3. the `n` for me is a mandatory parameter to I am still for doing

`retract_pade(M, p, X, n, t=1.0)` where only the `t` is optional. `t` can be real or integer, `n` is always integer. And I think with a proper documentation there is no need to clarify this in the documentation, so I do not see the need to overcomplicate things by introducing types here.

mateuszbaran:
Sorry but after reading your response I'm only more confident that we do need a type wrapper for `n` there -- it's just too easy to make a mistake by switching the order of `t` and `n`, and you suggest an order that is inconsistent with `retract`.

kellertuer:
Hm, I am still not that convinced, in principle Pole and Schield do the same with the number of steps (and even the NLSolve, though only for the inverse).
My main argument is that we also do not type p and X and in the same manner for `retract_pade` the n is a mandatory integer variable added _before_ the `t`.
For example the embedded retraction actually needs certain type, there we have that, but for `n` I only see that we can restrict it to integer. Still, since it is mandatory, I think there is no confusion (same as the thing that we can not confuse X (on the circle for example) with `t`. And sure this is only for `retract_pade` on level 3, I am not speaking about Level 1 or 2 where we keep the type for sure.

to be precise, the signatures we have are (in my model)

* `pade_retract(M, p, X, n, t=1.0)`, so the first integer (after `X` with also might be an integer) it `n` followed by the optional `t`
* `retract_pade!(M, q, p, X, n, t=1.0)` same here

in practice these are not exported and not called by a user directly, so the only point where this has an effect is implementing these. There I would prefer to keep it in this simplicity maybe?
On the other hand with your argument we also would have to type q,p,X because they could be confused in order?

mateuszbaran:
> Hm, I am still not that convinced, in principle Pole and Schield do the same with the number of steps (and even the NLSolve, though only for the inverse).

Pole and Schild are for vector transport IIRC, not retractions?

> in practice these are not exported and not called by a user directly, so the only point where this has an effect is implementing these. There I would prefer to keep it in this simplicity maybe?
> On the other hand with your argument we also would have to type q,p,X because they could be confused in order?

Our entire interface is designed around not constraining `q`, `p` and `X` so the possibility of confusing them is something we just have to accept. I've made the error multiple times, so this design choice isn't something I particularly like. Until IIRC Julia 1.5 it was also motivated by performance but that's no longer an issue.

If you really hate my suggestion then I won't push on this more but I won't like this interface either.

kellertuer:
> > Hm, I am still not that convinced, in principle Pole and Schield do the same with the number of steps (and even the NLSolve, though only for the inverse).
> 
> Pole and Schild are for vector transport IIRC, not retractions?

A yes mixed them up, we called them `pole_ladder` instead of `vector_transport_to_pole` or something but basically they have the same thing.

> 
> > in practice these are not exported and not called by a user directly, so the only point where this has an effect is implementing these. There I would prefer to keep it in this simplicity maybe?
> > On the other hand with your argument we also would have to type q,p,X because they could be confused in order?
> 
> Our entire interface is designed around not constraining `q`, `p` and `X` so the possibility of confusing them is something we just have to accept. I've made the error multiple times, so this design choice isn't something I particularly like. Until IIRC Julia 1.5 it was also motivated by performance but that's no longer an issue.
> 
> If you really hate my suggestion then I won't push on this more but I won't like this interface either.

First of all I like the untyped `p,q,X` since it allows for arbitrary things to be used with the packages, for example Arrays or StaticArrays without having to introduce wrapper types like `MPointArray` or something. I feel that might lead to overtyping and too many end users having to think about introducing the right types instead of things “just working” (if used in right order). Sure this is not a black-white situation but a spectrum, but I feel the current state is a good mix. Don't you?
And since we have a very consistent order of arguments, I think the confusion should not be so large. And I think the same about the `n`.

Though in the end, I mean for this PR, we just have to decide on one thing. I do not have a strong opinion here, so if you feel that a type for the `n` or passing down the retraction is really necessary, I would prefer passing down the retraction type, since that avoids introducing yet-another-type.
For me personally `n` here as the oder of pade has a quite concrete meaning so just keeping it an integer would be my preferred way.

IN short: My preferred way is to not type and keep the `n` as in the signatures I wrote above, but if you feel a type is necessary since two numbers as arguments might be confusing (I very much think that is not the case), then let's pass down the `PadeRetraction`.

mateuszbaran:
> Sure this is not a black-white situation but a spectrum, but I feel the current state is a good mix. Don't you?
> And since we have a very consistent order of arguments, I think the confusion should not be so large. And I think the same about the `n`.

I generally like our interface, I just wanted to point out that there are some drawbacks of not wrapping things. There are also drawbacks of wrapping things but a single layer of wrapping is rarely an issue in my experience.

Part of the issue with `retract_pade` is that its order or arguments is inconsistent with `retract`. Of course order of arguments for two different functions doesn't have to be consistent but they are somewhat related.

> IN short: My preferred way is to not type and keep the `n` as in the signatures I wrote above, but if you feel a type is necessary since two numbers as arguments might be confusing (I very much think that is not the case), then let's pass down the `PadeRetraction`.

OK, I think I will pass down `PadeRetraction` then.

kellertuer:
> > Sure this is not a black-white situation but a spectrum, but I feel the current state is a good mix. Don't you?
> > And since we have a very consistent order of arguments, I think the confusion should not be so large. And I think the same about the `n`.
> 
> I generally like our interface, I just wanted to point out that there are some drawbacks of not wrapping things. There are also drawbacks of wrapping things but a single layer of wrapping is rarely an issue in my experience.

Ok, I feel wrapping might be less usable for an end-user who just has arrays and wants to use those. So that is not a computational drawback but a usability one, that is might feel more complicated.
 
> Part of the issue with `retract_pade` is that its order or arguments is inconsistent with `retract`. Of course order of arguments for two different functions doesn't have to be consistent but they are somewhat related.

Could you elaborate a bit on that since I feel if “retraction goes last if it is there” and “t is before that if it is there” are the assumptions we do, then this is consistent.

> 
> > IN short: My preferred way is to not type and keep the `n` as in the signatures I wrote above, but if you feel a type is necessary since two numbers as arguments might be confusing (I very much think that is not the case), then let's pass down the `PadeRetraction`.
> 
> OK, I think I will pass down `PadeRetraction` then.

Ok, as I said not my favourite one, but one I am ok with, we have the embedded_retraction which does that as well because it needs a retraction there, so that is not really inconsistent. And actually one could dispatch on `PadeRetraction{1}` if needed for Caley? Also nice.

kellertuer:
> > > > Sure this is not a black-white situation but a spectrum, but I feel the current state is a good mix. Don't you?
> > > > And since we have a very consistent order of arguments, I think the confusion should not be so large. And I think the same about the `n`.
> > > 
> > > 
> > > I generally like our interface, I just wanted to point out that there are some drawbacks of not wrapping things. There are also drawbacks of wrapping things but a single layer of wrapping is rarely an issue in my experience.
> > 
> > 
> > Ok, I feel wrapping might be less usable for an end-user who just has arrays and wants to use those. So that is not a computational drawback but a usability one, that is might feel more complicated.
> 
> Sure, I think we should change that part of our interface.
> 
> > > Part of the issue with `retract_pade` is that its order or arguments is inconsistent with `retract`. Of course order of arguments for two different functions doesn't have to be consistent but they are somewhat related.
> > 
> > 
> > Could you elaborate a bit on that since I feel if “retraction goes last if it is there” and “t is before that if it is there” are the assumptions we do, then this is consistent.
> 
> Yes, these are the assumptions I have in mind but in your suggestion retraction-related information (`n`) goes before `t`.
> 

Retraction (type) goes last, not retraction information – but well, as I said let's keep the retraction type also for the dispatch argument.

> > And actually one could dispatch on `PadeRetraction{1}` if needed for Caley? Also nice.
> 
> Yes, begin able to dispatch on `n` is a small benefit of not unwrapping, so we don't even need `retract_caley` then (which has a typo BTW...)

Oh! In deed, let's deprecate the typo one then. I think `retract_caley` should be kept, though merely for a user who has `caley` in mind, wants to overwrite that without knowing/thinking about that it is a special case of pade. So, keeping `caley` is mainly for the users convenience again (that it is another name for pade(1)).

src/ManifoldsBase.jl
```diff
@@ -425,6 +425,35 @@ See also: [`EmbeddedManifold`](@ref), [`project!`](@ref project!(M::AbstractMani
...
+end
+"""
+    embed_project(M::AbstractManifold, p, X)
+
+Embed vector `X` tangent at `p` from manifold `M` an project it back to tangent space
+at `p`. For points from that tangent space this is identity but in case embedding is
+defined for tagent vectors from outside of it, this can serve as a way to for example remove
+numerical innacuracies caused by some algorithms.
+"""
+function embed_project(M::AbstractManifold, p, X)
```
mateuszbaran, line 446: This should fix https://github.com/JuliaManifolds/Manifolds.jl/issues/522 . I've added it here because I'm definitely expecting this PR to be breaking anyway so it's going to be large-ish anyway.

mateuszbaran: I'm also going to use this as a ManifoldsBase.jl replacement for `retract!` and `project_tangent!` in Optim.jl: https://github.com/JuliaNLSolvers/Optim.jl/blob/master/src/Manifolds.jl .

kellertuer: I am not sure why we need this new function, but ok;

With the `retract!` from Optim.jl be careful that the retract there is not our retract (but basically “just” a projection) if I remember the discussion correctly from a longer discussion with Antoine (since he used the topological retraction in there if I remember correctly)

kellertuer: For the breaking-ness – I had hoped this would be possible in a non-breaking way, but sure if it breaks, then it breaks. I would then just like to be a bit more precise in the release what actually brakes and how to fix things (we did not do that much before, but I feel it might be nice also for us to remember).

mateuszbaran: > I am not sure why we need this new function, but ok;
> 
> With the `retract!` from Optim.jl be careful that the retract there is not our retract (but basically “just” a projection) if I remember the discussion correctly from a longer discussion with Antoine (since he used the topological retraction in there if I remember correctly)

If I understand right, Optim's `retract is `embed_project` for points and `project_tangent` is `embed_project` for tangent vectors -- if you think I'm wrong, then I'd like to know :slightly_smiling_face: .

> For the breaking-ness – I had hoped this would be possible in a non-breaking way, but sure if it breaks, then it breaks. I would then just like to be a bit more precise in the release what actually brakes and how to fix things (we did not do that much before, but I feel it might be nice also for us to remember).

Sure, I will make a list of changes when I get Manifolds.jl running.

mateuszbaran:
I've changed `DefaultManifold` to store size in a field to make it do less compilation. Actually very little can be optimized by statically knowing the dimension.

mateuszbaran:
By the way, I'm up to about Stiefel in updating Manifolds.jl to these changes. I will make a PR to Manifolds.jl when tests pass locally.

kellertuer:
If you say that thats easier to handle, sure the DefaultManifold can be changed.

Nice, I assume it is a bit of work to get the new versions fast?

mateuszbaran:
> Nice, I assume it is a bit of work to get the new versions fast?

Most of the time it's either easy or there doesn't seem to be much room for improvement over just multiplying the vector by `t`. At the moment the biggest problem is that I'm not sure how to multiply `StiefelFactorization` by `t`. Yes, it's defined as an `MPoint` but it's also used to represent tangent vectors.

kellertuer:
I can check tomorrow, but I thinking's mainly some caching (maybe for Cholesky?).

kellertuer:
Oh yes that is quite unfortunate that the `StiefelFactorization` is an `MPoint` but also used as a tangent vector, it should probably just be struct? At first sight, I do not see direct how the multiplication with t works here and would have to check the ZimmermannHüper paper more closely.

mateuszbaran:
Yes, I'd prefer `StiefelFactorization` to just be a struct. I'd expect the scaling to be just multiplication of the `Z` part by `t` but I'm not fully sure.

kellertuer:
Then lets just make it a struct and as far as I see it might indeed just be the Z, yes.

mateuszbaran:
I just discovered that #134 made `isapprox` incredibly slow -- something like 35 microseconds when actual `isapprox` on arrays takes about 60 ns. My last commit got it down to about 1.2 us but to make it acceptable I need to change the interface of `function check_approx(M::AbstractManifold, p, q; kwargs...)` a bit -- we can't have it generate a string when all we need is a yes/no response. Do you have a preferred way to handle this @kellertuer ?

mateuszbaran:
BTW, I discovered the issue when I was investigating why statistics.jl tests in Manifolds.jl take so long to finish (about 20 minutes on my computer). A large part of that was calling `isapprox` in median calculation.

kellertuer:
> I just discovered that #134 made `isapprox` quite slow -- something like 35 microseconds when actual `isapprox` on arrays takes about 60 ns. My last commit got it down to about 1.2 us but to make it acceptable I need to change the interface of `function check_approx(M::AbstractManifold, p, q; kwargs...)` a bit -- we can't have it generate a string when all we need is a yes/no response. Do you have a preferred way to handle this @kellertuer ?

I am often amazed what you find :)
Since that is an internal function sure we can change it. But to my (though untrained) eye, the new shortcut-return should already the point that no string is generated?

I do not have a preferred way of handling this, no, if you have a good idea, feel free to propose it here.

mateuszbaran:
> I am often amazed what you find :)

Thanks. I can show you how I do it if you want :). It's not that complicated.

> Since that is an internal function sure we can change it. But to my (though untrained) eye, the new shortcut-return should already the point that no string is generated?

This shortcut skips creating one string, but `check_approx(M::AbstractManifold, p, q; kwargs...)` still makes one.

I'm not sure what the best solution here is. Currently manifolds implement `isapprox` ignoring the new interface. We probably need the fast yes/no implementations to just coexist next to the ones providing errors. The logic is going to be doubled but I feel that the cost of not doubling it and providing full functionality is too large.

`is_point`/`check_point` functions may have a similar issue but they are less likely to occur in a performance-critical code.

kellertuer:
I thought we even introduced the `check_` functions to split something already.

I am not so much a fan of doubling, but sure if that is necessary for performance, then let‘s do that.
If it is not too much work, I would prefer doing the same for point/vector as well, because that would be consistent.

mateuszbaran:
There wouldn't be that much doubling in `isapprox` because it has few methods and they are relatively simple. Doing something similar for `is_point`/`is_vector` would be much more doubling though, and the speed gains would be less important. So I'm not sure it's a good idea at this time.

kellertuer:
That‘s why I phrased it carefully, then the inconsistency is ok and we just adapt isapprox, of course.

mateuszbaran:
Manifolds.jl tests pass for me locally now with these changes, so I think you can start looking at this :slightly_smiling_face: . I will still investigate a bit more if these are all the changes we need at the moment, since it's breaking, and will make a short list of changes.

src/ManifoldsBase.jl
```diff
@@ -506,19 +535,23 @@ Currently the following are supported
...
 function isapprox(M::AbstractManifold, p, q; error::Symbol = :none, kwargs...)
-    ma = check_approx(M, p, q; kwargs...)
-    if ma !== nothing
-        (error === :error) && throw(ma)
-        if isnan(ma.val)
-            s = "$(typeof(ma))\n$(ma.msg)"
-        else
-            s = "$(typeof(ma)) with $(ma.val)\n$(ma.msg)"
+    if error === :none
+        return isapprox(p, q; kwargs...)
```
kellertuer, line 539: I do not like this. Especially for posts on manifolds that require information from the manifold still (I do not have an example in mind but well)
or that like FixedRankMatrices have a very special form – I would still expect them to implement the check? Or is this intended to be the new switch of logic?

src/ManifoldsBase.jl
```diff
@@ -542,19 +575,23 @@ By default these informations are collected by calling [`check_approx`](@ref).
...
 function isapprox(M::AbstractManifold, p, X, Y; error::Symbol = :none, kwargs...)
-    mat = check_approx(M, p, X, Y; kwargs...)
-    if mat !== nothing
-        (error === :error) && throw(mat)
-        if isnan(mat.val)
-            s = "$(typeof(mat))\n$(mat.msg)"
-        else
-            s = "$(typeof(mat)) with $(mat.val)\n$(mat.msg)"
+    if error === :none
+        return isapprox(X, Y; kwargs...)
```
kellertuer, line 579: same as above, just a bit worse, since equality of two  vectors might depend on the point?

kellertuer:
Will check the fusion soon, for now just 2 comments on the change in isapprox.

src/ManifoldsBase.jl
mateuszbaran: This is kind of new and old logic. New, because `isapprox` can't call `check_approx` by default (it's quite slow). Old, because right now all manifolds implement `isapprox` anyway instead of `check_approx` so there isn't really much to change.

src/ManifoldsBase.jl
mateuszbaran: This is similar thing to point comparison, let's discuss it there.

src/ManifoldsBase.jl
mateuszbaran: I'm definitely open to having a different interface but one constraint is that we can't create any error objects in the fast/default case.

kellertuer: I though about this a bit yesterday evening. What I do not like in the following form is

`check_approx(M,p,q; kwargs...)` does the same as `isapprox(p,q; kwargs...)` in these lines, but the first allocates a string which we want to get rid of.
Note that the first has much more information (M, for tangent vectors even M and p) so that seems for „doing the same” a bit strange.

I am not the biggest fan of doubling logix here, but maybe that is needed, since one works on providing more information in the string along the way.

a) My idea would be an internak function that has the same signature as `check_approx` but just returns true/false, maybe `_isapprox(M,p,q;kwargs...)`?

we could also
b) have two signatures of `check_approx` something like `check_approx(M, p, q, r; kwargs...)`where `r` is something to dispatch on determining the return type? So maybe even `r=String` and `r=Bool`? The disadvanbtage of this is, it would be quite breaking and all current checks need a new `r::Type{String}`. On the other hand this might even allow for other check types later as well (no clue which ;))

mateuszbaran: I think I prefer the `_isapprox` solution so I will go with that if you are OK with it. `r::Type{String}` seems like too much complexity added to a relatively simple thing.

kellertuer: ok, I felt like the type approach would be non-breaking and easier to adapt in Manifolds, but I do not have a strong preference.

mateuszbaran: There is not much work either way actually :slightly_smiling_face: .

mateuszbaran: I've added `_isapprox`. It's not ideal because our traits cause type instability when using `isapprox` but I think we have to accept that.

kellertuer: does the other idea (the dispatch in check) avoid that or does that have the same issue?

mateuszbaran: I've just spent some time digging into it and the issue is that Julia really dislikes optimizing through our trait mechanism (see this if you want to dig why: https://github.com/JuliaLang/julia/blob/cd955e15fa6f55cd24a7b759d64b4ba72595e1bb/base/compiler/typelimits.jl#L17 ). It's not a problem if we don't care too much about performance or use traits just a little but in the case I've been checking (`isapprox(Euclidean(2), [2, 3], [3, 4])`) it is quite complicated.

mateuszbaran: Ah, I've found a solution, though it's quite ugly. I will post here later about it.

kellertuer: 😱ugly code!

We will see, maybe we find something nicer in the end

mateuszbaran: It seems to work, and the ugly part looks like that:
```julia
for trait_type in [
    TraitList{<:IsDefaultMetric},
    TraitList{<:IsDefaultConnection},
    TraitList{IsMetricManifold},
    TraitList{IsConnectionManifold},
]
    @eval begin
        ManifoldsBase.@next_trait_function $trait_type isapprox(
            M::AbstractDecoratorManifold,
            p,
            q;
            kwargs...,
        )
        ManifoldsBase.@next_trait_function $trait_type isapprox(
            M::AbstractDecoratorManifold,
            p,
            X,
            Y;
            kwargs...,
        )
    end
end
```
But we may need that for more functions and more traits, maybe even all of them.

kellertuer: Which basically skips certain traits? Isn't it better to make is approx transparent to these?

mateuszbaran: No, it's functionally equivalent to what we have now. It just makes more "copies" of the go-to-the-next-trait method so that we don't have method recursion.

What do you mean by making approx transparent to these? Not adding these traits to `isapprox` dispatch?

kellertuer: Well preeviously some trait not doing anything on a function is what we called them “transparent”, right? And I thought what you are introducing there was actually the default behaviour already? But I might have to check.

kellertuer: I just remember that “pass through for default” was not so complicated as in your code.

mateuszbaran:
I've added a changelog, what do you think?

src/ManifoldsBase.jl
kellertuer: Ah, I too often think in the “just work” regime of code. So my intuition here was correct, just not my compiler ideas. But well, if this is necessary then lets leave it at that.

kellertuer:
> I've added a changelog, what do you think?

Looks good, thanks for starting this. We should do this in the other packages as well, at least for breaking changes/releases.
And the changelog contains the one criticism I see in the nicer, faster variants – the user has to think of this and implement a “slightly more complicated version”. Should we adapt the tutorial to that? Se https://juliamanifolds.github.io/ManifoldsBase.jl/stable/example.html#manifold-tutorial-fn and maybe even the technical note before?

mateuszbaran:
> We should do this in the other packages as well, at least for breaking changes/releases.

Yes, that would be nice. At least this is a start :slightly_smiling_face: .

> And the changelog contains the one criticism I see in the nicer, faster variants – the user has to think of this and implement a “slightly more complicated version”. Should we adapt the tutorial to that? Se https://juliamanifolds.github.io/ManifoldsBase.jl/stable/example.html#manifold-tutorial-fn and maybe even the technical note before?

Sure, I forgot about it. I will modify the tutorial.

docs/src/example.md
```diff
@@ -31,13 +31,15 @@ There are only two small technical things we need to explain at this point befor
...
 
-Second, this interface usually provides both an allocating and a mutating variant of each function, for example for the [`exp`](@ref)onential map [implemented below](@ref manifold-tutorial-fn) this interface provides `exp(M,p,X)` to compute the exponential map and `exp!(M, q, p, X)` to compute the exponential map in the memory provided by `q`, mutating that input.
+Second, this interface usually provides both an allocating and a mutating variant of each function, for example for the [`exp`](@ref)onential map [implemented below](@ref manifold-tutorial-fn) this interface provides `exp(M, p, X)` to compute the exponential map and `exp!(M, q, p, X)` to compute the exponential map in the memory provided by `q`, mutating that input.
 the convention is, that the manifold is the first argument -- in both function variants -- the mutating variant then has the input to be mutated in second place, and the remaining parameters are again the same (`p`and `X` here).
 We usually refer to these two variants of the same function as the allocating (`exp`) function and the mutating (`exp!`) one.
```
kellertuer, line 36: since we are at it, I think in Manopt I have switched to calling the `!` variants the in-place variants (instead of mutating),

docs/src/example.md
```diff
@@ -31,13 +31,15 @@ There are only two small technical things we need to explain at this point befor
...
 
 The convention for this interface is to __document the allocation function__, which by default allocates the necessary memory and calls the mutating function. So the convention is to just __implement the mutating function__, unless there is a good reason to provide an implementation for both.
 For more details see [the design section on mutating and non-mutating functions](@ref mutating-and-nonmutating)
 
+For performance reasons scaled variants of retractions and exponential maps are expected to be implemented. Scale is specified by an optional argument that comes after the tangent vector and by default it is equal to 1. You can also add a method that does not take the scaling factor `t` but it is not required.
```
kellertuer, line 41: we could add the signatures here
```suggestion
For performance reasons scaled variants of retractions `retraction!(M, q, p, X, t, m)` and exponential maps `exp!(M, q, p, X, t) are should to be implemented. Scale is specified by an optional argument that comes after the tangent vector and by default it is equal to 1. The variant without scaling, e.g. `exp!(M, q, p, X)`, can be implemented as well if there is a good reason like performance, but the default fallback of this variant is to call the previous one with `t=1`.
```

src/ManifoldsBase.jl
```diff
@@ -163,19 +163,23 @@ not approximately equal, containting a more detailed description/reason.
 If the two elements are approximalely equal, this method returns `nothing`.
 
 This method is an internal function and is called by `isapprox` whenever the user specifies
-an `error=` keyword therein.
+an `error=` keyword therein. `_isapprox` is another related internal function. It is
```
kellertuer, line 166: Should we maybe add a doctoring to `_isapprox` as well?

src/ManifoldsBase.jl
```diff
@@ -542,19 +579,40 @@ By default these informations are collected by calling [`check_approx`](@ref).
...
-        return false
+        return true
     end
-    return true
+end
+
+"""
+    _isapprox(M::AbstractManifold, p, q; kwargs...)
+
+See documentation of `check_approx`.
```
kellertuer, line 604: If we keep it this way, please with ``[`check_approx`](@ref)``, but I would prefer at least a bit of docs here as well (that it is a faster true/false variant of check).

src/exp_log_geo.jl
```diff
@@ -30,7 +34,7 @@ The result is saved to `q`.
 See also [`exp`](@ref).
```
kellertuer, line 34: one could add a small technical note here as well, that the most general one to implement is the scaled one?

src/nested_trait.jl
```diff
@@ -258,6 +258,27 @@ macro invoke_maker(argnum, type, sig)
 end
 
 
+macro next_trait_function(trait_type, sig)
```
kellertuer, line 261: could we document this a bit?

src/exp_log_geo.jl
```diff
@@ -17,11 +17,15 @@ function exp(M::AbstractManifold, p, X)
     exp!(M, q, p, X)
     return q
 end
-exp(M::AbstractManifold, p, X, t::Real) = exp(M, p, t * X)
+function exp(M::AbstractManifold, p, X, t::Number)
```
kellertuer, line 20: We could actually implement both `exp`s also as

```
function exp(M::AbstractManifold, p, X, t::Number=one(eltype(X)))
```

and even advice this generally – or would that kill performance?

kellertuer (commented):
This looks nearly finished, I just have a few small comments, one that might reduce codeines maybe with the default for t?

docs/src/example.md
mateuszbaran: Sure, I will rename it.

src/ManifoldsBase.jl
mateuszbaran: OK, I will expand the docs a bit.

src/exp_log_geo.jl
mateuszbaran: I'm not sure actually, what would be the benefit here?

kellertuer: I am also not 100% sure but it would make this change a bit easier for cases with or without t? I you do not feel that this is useful we can also leave that out.

mateuszbaran: Well, the current variant definitely works and I don't see how this change would make anything significantly easier.

src/nested_trait.jl
mateuszbaran: I've added some documentation

src/exp_log_geo.jl
mateuszbaran: I've added a note, is that enough?

src/ManifoldsBase.jl
mateuszbaran: I've added a docstring there.

docs/src/example.md
mateuszbaran: I've renamed `mutating` to `in-place`, but for now I've left `non-mutating`.

kellertuer: that's fine with me. I even called the “non-mutating” one “allocating”, but I think that is fine giving it a negative (non-) name here.

src/nested_trait.jl
kellertuer: I actually do not understand much of what is documented. That is fine with me, but one could add an example maybe?

kellertuer (approved):
I'll approve this already since the last few changes seem to be mainly refs :)

src/nested_trait.jl
kellertuer: We have the mentioned file twice (in source and test) and in a not-so-detailed scrolling it was hard to find sections?

mateuszbaran: How would you prefer to refer to that code?

kellertuer: Hm, well, my main problem is, that I still do not directly see how to use it, if the macro is used within an eval (kind of like in a complicated environment overall. I would prefer a line like

```
Defining/using the Macro like

         @next_trait_function TraitList{IsEmbeddedManifold} f(M::AbstractDecoratorManifold, p, q)

bring the benefit, that you obtain more gummi bears... (or whatever the macro does)
```
Since I still have not yet understood the magic of this macro and the docs read like it basically does nothing but is still important for the compiler

mateuszbaran: Well, it does nothing other than nicely asking the compiler to try harder optimizing some calls :wink: . It is not intended for normal users, because before attempting to use it you should check if it makes sense at all. The easiest way is by going through unoptimized Cthulhu output with remarks turned on, which is fairly hard to interpret. I think the only way someone other than me can use it is by adding a new trait or a new function to one of the two `@eval` blocks (either the one in ManifoldsBase.jl or a similar one in Manifolds.jl) and looking if it helps. Explaining the benefit here is fairly hard and maybe not worth the effort because hopefully some time in the future compiler won't need this kind of help.

kellertuer: Ok, so my gummi bears was not that far away, you definitely need them once you have to use these ;)

But maybe a careful phrasing of your first sentence might help anyone here? Also the note that it affects certain traits and that there is one block at the end of the decorator_traits file where this is used. Maybe that would be a good choice to add these details? (and mention that its `src/decorator_traits.jl` at the end, to make the reference more precise).

mateuszbaran: I've updated the docs a bit, feel free to suggest a change if you have an idea.

kellertuer: OK, that's fine with me :)

mateuszbaran:
I think everything looks fine now and this can be merged :slightly_smiling_face: .

PR #148, opened 2023-02-16
Introduce change_representer already in ManifoldsBase

### LINKED ISSUE #147: Were we not careful enough with ManifoldsBase 0.14?
I just noticed that since ManifoldsBase 0.14 we have quite a few inexact errors with NaNs?

see

https://github.com/JuliaManifolds/ManifoldsBase.jl/actions/runs/4187041219/jobs/7256354396#step:5:354

kellertuer:
* Introduce `change_representer`already in ManifoldsBase so it can be nicely used in Manopt
* also fixes #147 removing debug output from test run.

We still might need tests for change_representer.

### DIFF
--- a/test/metric.jl
+++ b/test/metric.jl
@@ -1,5 +1,6 @@
 using Test
 using ManifoldsBase
+import ManifoldsBase: inner, change_representer, change_metric
 
 @testset "Metrics" begin
     @test ManifoldsBase.EuclideanMetric() isa ManifoldsBase.AbstractMetric
--- a/test/metric.jl
+++ b/test/metric.jl
@@ -1,6 +1,5 @@
 using Test
 using ManifoldsBase
-import ManifoldsBase: inner, change_representer, change_metric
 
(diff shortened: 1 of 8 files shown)


mateuszbaran:
I think this is a good idea.

kellertuer:
Code coverage misses 2 lines from Julia 1.0 because the tests report 

```
Codecov: Error validating uploader: Signed digest did not match
```

test/metric.jl
```diff
@@ -1,5 +1,6 @@
 using Test
 using ManifoldsBase
+import ManifoldsBase: inner, change_representer, change_metric
```
mateuszbaran, line 3: This import doesn't seem to be used.

kellertuer: ah yes, I first thought I write the tests there but when I then added the code directly to the DefaultManifold, I forgot to remove that line again.

mateuszbaran (approved):

PR #158, opened 2023-06-30
Simplify the tutorial and move to quarto.

kellertuer:
This PR simplifies the existing tutorial on how to define a manifold and moves it to a quarto notebooks as well.

I noticed two things this PR could also resolve

* it seems for `exp` it is enough to define `exp!(M, q, p, X)` the `exp!(M, q, p, X, t)` falls back to this. for retract it is the other way around, one has to define `retract_method!(M, q, p, X, t)`
* I noticed that the method to implement for a retraction is not always clear, we could add such a note to each `AbstractRetractionMethod` subtype which lower level function corresponds to the high level type?

Besides that this needs a careful check that the new CI runs fine and that all examples and references are correct.
The example itself was not changed much.

ToDo
* [x] references in the new example seem to not yet work (the `@ref`s) I am not 100% sure why since I made that the same way as over on the other packages.


kellertuer:
A future idea would be to do a tutorial for traits with an embedding idea, and maybe one for metrics or such.

kellertuer:
Hm, locally everything works, but the documenter CI says IJulia is missing the 1.9 kernel? Not yet sure why...the first set seems to be a bit difficult every now and then of quarto+Julia maybe.

kellertuer:
Hm, I am a bit lost here, all config files are exactly the same as for Manopt.jl and IJulia us built directly before we switch to quarto, still IJulia seems to not have Julia 1.9 in its kernels. To me a total mystery.

kellertuer:
Further investigations, directly before it says the kernel is not present it lists the kernel as present see the currently added warning at https://github.com/JuliaManifolds/ManifoldsBase.jl/actions/runs/5431263051/jobs/9877519967#step:9:709

So. The kernel is installed but not found. Hæ?

kellertuer:
Finally managed to set up the CI correctly. Now the only thing left is to find out why `@ref`s in the tutorial do not yet work.

kellertuer:
This is finished besides the discussion on the default difference between retract and exp

mateuszbaran:
Sure, I think this can use Quarto as well.

> * it seems for `exp` it is enough to define `exp!(M, q, p, X)` the `exp!(M, q, p, X, t)` falls back to this. for retract it is the other way around, one has to define `retract_method!(M, q, p, X, t)`

IIRC I wanted to make `exp!(M, q, p, X)` fall back to `exp!(M, q, p, X, t)` but it would be technically a braking change so we are in a somewhat inconsistent state now.

> * I noticed that the method to implement for a retraction is not always clear, we could add such a note to each `AbstractRetractionMethod` subtype which lower level function corresponds to the high level type?

Sure, that is a good idea.

kellertuer:
But can we fix that without breaking, that is having both work the same?

This is not so urgent since it mainly affects people _implementing_ manifolds, not _using_ them, but it would be great to have a unified interface.

For the Retraction types, I will add that to the docs then.

kellertuer:
I added a note to all retractions and inverse retractions which functions are related to them. I also moved the discussion about exp/retract to a new issue – and did a thorough check about the current state there.

So this PR should be ready to go :)

mateuszbaran:
> But can we fix that without breaking, that is having both work the same?

As far as I know, we can't.

kellertuer:
Well, then with breaking ;) Still unifying them would help users I think.

mateuszbaran:
I agree, we should unify them next time we make a breaking change here.

mateuszbaran:
image
There seems to be something wrong with formulas in the new tutorial.

kellertuer:
Then let's find a good unification in the issue and do that before the next breaking change :)

kellertuer:
Ups, it seems I missed that `\mid` was not recognised. should be fixed now (i.e. works locally now).

tutorials/implement-a-manifold.qmd
```diff
@@ -0,0 +1,283 @@
...
+
+
+The first two functions we want to define are those to check points and tangent vectors for our manifold.
+Let's first clarify what the tangent space looks like. The directions “we can walk into” from a point $p\in \mathbb S_r^d$ are all $X$ that are orthogonal to $p$, which is the plane/vector space tangent to the sphere. Formally
+
+```math
+T_p\mathbb S_r^d :=
+\bigl\{
+    X \in \mathbb R^{d+1}
+    \big\mid
```
mateuszbaran, line 127:
```suggestion
    \big|
```

mateuszbaran (approved):

kellertuer:
THanks, since this does not introduce a new feature is it ok if we leave it on master until the next feature comes along?
Or should I register this as a new version?

mateuszbaran:
I think it's better to register since stable docs will then include this change.

PR #160, opened 2023-07-03
Add interface for optionally static size

mateuszbaran:
Basic infrastructure for solving https://github.com/JuliaManifolds/Manifolds.jl/issues/625 .

### DIFF
--- a/src/DefaultManifold.jl
+++ b/src/DefaultManifold.jl
@@ -7,7 +7,7 @@ to build one's own manifold. It is a simplified/shortened variant of `Euclidean`
 
 This manifold further illustrates how to type your manifold points and tangent vectors. Note
 that the interface does not require this, but it might be handy in debugging and educative
-situations to verify correctness of involved variabes.
+situations to verify correctness of involved variables.
 
 # Constructor
 
@@ -25,13 +25,7 @@ struct DefaultManifold{𝔽,T} <: AbstractManifold{𝔽}
     size::T
 end
 function DefaultManifold(n::Vararg{Int}; field = ℝ, parameter::Symbol = :field)
-    if parameter === :field
-        size = n
-    elseif parameter === :type
-        size = TypeParameter(n)
-    else
-        throw(ArgumentError("Parameter can be either :field or :type. Given: $parameter"))
-    end
+    size = wrap_type_parameter(parameter, n)
     return DefaultManifold{field,typeof(size)}(size)
 end
 
--- a/src/DefaultManifold.jl
+++ b/src/DefaultManifold.jl
@@ -21,12 +21,12 @@ Arguments:
 - `parameter`: whether a type parameter should be used to store `n`. By default size
   is stored in a field. Value can either be `:field` or `:type`.
 """
-struct DefaultManifold{𝔽,T<:AbstractManifoldParameter} <: AbstractManifold{𝔽}
+struct DefaultManifold{𝔽,T} <: AbstractManifold{𝔽}
     size::T
 end
 function DefaultManifold(n::Vararg{Int}; field = ℝ, parameter::Symbol = :field)
     if parameter === :field
-        size = FieldParameter(n)
+        size = n
     elseif parameter === :type
         size = TypeParameter(n)
     else
@@ -150,7 +150,7 @@ function Base.show(io::IO, M::DefaultManifold{𝔽,<:TypeParameter}) where {𝔽
         "DefaultManifold($(join(get_parameter(M.size), ", ")); field = $(𝔽), parameter = :type)",
     )
 end
-function Base.show(io::IO, M::DefaultManifold{𝔽,<:FieldParameter}) where {𝔽}
+function Base.show(io::IO, M::DefaultManifold{𝔽}) where {𝔽}
     return print(io, "DefaultManifold($(join(get_parameter(M.size), ", ")); field = $(𝔽))")
 end
 
--- a/src/DefaultManifold.jl
+++ b/src/DefaultManifold.jl
@@ -9,11 +9,20 @@ This manifold further illustrates how to type your manifold points and tangent v
 that the interface does not require this, but it might be handy in debugging and educative
 situations to verify correctness of involved variabes.
 """
-struct DefaultManifold{𝔽,T<:NTuple{N,Int} where {N}} <: AbstractManifold{𝔽}
+struct DefaultManifold{𝔽,T<:AbstractManifoldSize} <: AbstractManifold{𝔽}
     size::T
 end
-function DefaultManifold(n::Vararg{Int}; field = ℝ)
-    return DefaultManifold{field,typeof(n)}(n)
+function ManifoldsBase.DefaultManifold(
+    n::Vararg{Int};
+    field = ManifoldsBase.ℝ,
+    static = false,
+)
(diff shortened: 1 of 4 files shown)


src/maintypes.jl
```diff
@@ -36,3 +36,31 @@ matrix internally, it is possible to use [`@manifold_element_forwards`](@ref) an
...
+"""
+struct StaticSize{T} <: AbstractManifoldSize end
+StaticSize(t::NTuple) = StaticSize{t}()
+
+"""
+    RTSize{TS<:NTuple{N,Int} where N}
+
+Runtime size of a manifold. 
+"""
+struct RTSize{TS<:NTuple{N,Int} where {N}} <: AbstractManifoldSize
```
kellertuer, line 61: I think `RuntimeSize` is not too bad here.

Though maybe another pair of possible names might be `ParameterSize` and `Fieldsize` or something like that?

src/DefaultManifold.jl
```diff
@@ -9,11 +9,20 @@ This manifold further illustrates how to type your manifold points and tangent v
...
-struct DefaultManifold{𝔽,T<:NTuple{N,Int} where {N}} <: AbstractManifold{𝔽}
+struct DefaultManifold{𝔽,T<:AbstractManifoldSize} <: AbstractManifold{𝔽}
     size::T
 end
-function DefaultManifold(n::Vararg{Int}; field = ℝ)
-    return DefaultManifold{field,typeof(n)}(n)
+function ManifoldsBase.DefaultManifold(
+    n::Vararg{Int};
+    field = ManifoldsBase.ℝ,
+    static = false,
```
kellertuer, line 18: We could maybe – for consistency, do this as a keyword? It will be a keyword for Euclidean anyways.

Also just `static` might not be so informative as a keyword, maybe we can find a nicer name?

kellertuer (commented):
Nice that you started this.

While I think this is quite technical and should not concern beginners, it might be nice for advanced usages.
A small benchmark illustrating when which is nicer might be nice as well (if that can be well-illustrated)

mateuszbaran:
> While I think this is quite technical and should not concern beginners, it might be nice for advanced usages.
> A small benchmark illustrating when which is nicer might be nice as well (if that can be well-illustrated)

Yes, it's definitely not for beginners.
Benchmarks would be hard at the moment because all good examples live in Manifolds.jl. I think rotation matrices would make a great example for a benchmark.

src/maintypes.jl
mateuszbaran: I think I prefer `RuntimeSize`.

src/DefaultManifold.jl
mateuszbaran: This is a keyword now (notice the semicolon).

Other possible names would be perhaps `static_size`, `specialize_for_size`, `specialize_size` or something like that? Nothing sounds particularly convincing.

src/maintypes.jl
```diff
@@ -36,3 +36,31 @@ matrix internally, it is possible to use [`@manifold_element_forwards`](@ref) an
...
+    size::TS
+end
+
+getsize(::StaticSize{T}) where {T} = T
+getsize(S::RTSize) = S.size
```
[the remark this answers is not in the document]
kellertuer: I am also not 100% sure about the `get_size_parameter(M)` method but checking that the field is zero in the implementation seems to me like a magic number and I would prefer a dispatch variant for accessing this, though in order to use it you have to implement it per manifold.
It will make code that uses the parameter more readable.

src/maintypes.jl
kellertuer: Runtime and static are more like meta-implications of this difference in storing the parameter.

I think I'd prefer names that not address the meta level directly?

One is a parameter the other a field and that is the difference. Or phrased differently. I read runtime but I have to think field – I read static but I have to think parameter. This one step in thinking is what I do not like in this naming so much.

mateuszbaran: I see, that's a good point. I will change it to `ParameterSize` and `FieldSize`.

mateuszbaran:
Sure, I will add a clear explanation with an example in the Manifolds.jl PR that converts key manifolds to `AbstractSize`.

src/maintypes.jl
kellertuer: Besides the Size name ;) But `ParameterParameter` sounds silly. But `ParameterX` and `FieldX` are good, we just need a better name for `X`.

kellertuer: Hm though these parameters are not directly the dimension it is what they store so maybe `X` as `Dimension` would be best, though to be precise it is actually DimensionInformation that is stored?

src/DefaultManifold.jl
mateuszbaran: Hm, then maybe we should also use `TypeParameter` and `FieldParameter` names for the types?

kellertuer: That sounds good :)

src/maintypes.jl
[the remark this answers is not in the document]
mateuszbaran: Do you mean this: `return length(size) == 0 ? 1 : *(size...) * real_dimension(𝔽)` by "checking that the field is zero"? I can get rid of that using `prod` function.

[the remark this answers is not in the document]
kellertuer: Yes that line has this “magic zero” I referred to.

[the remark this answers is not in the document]
mateuszbaran: Latest commit removed it :slightly_smiling_face:

src/DefaultManifold.jl
```diff
@@ -8,12 +8,31 @@ to build one's own manifold. It is a simplified/shortened variant of `Euclidean`
...
 """
-struct DefaultManifold{𝔽,T<:NTuple{N,Int} where {N}} <: AbstractManifold{𝔽}
+struct DefaultManifold{𝔽,T<:AbstractManifoldParameter} <: AbstractManifold{𝔽}
     size::T
 end
-function DefaultManifold(n::Vararg{Int}; field = ℝ)
-    return DefaultManifold{field,typeof(n)}(n)
+function DefaultManifold(n::Vararg{Int}; field = ℝ, parameter::Symbol = :field)
+    if parameter === :field
+        size = FieldParameter(n)
```
kellertuer, line 29: one could even just use a number or an attar here, right?

Though then the  `get_parameter` will be a bit more difficult (and the current constraint on `T` has to be let go of).

mateuszbaran: Hm, technically we could have
```julia
get_parameter(::TypeParameter{T}) where {T} = T
get_parameter(P) = P
```
i.e. everything other than `TypeParameter` would be stored without any wrapper. That would also be fine.

kellertuer: Yes, something like that. Do you think that might be confusing? I think that is an easier solution and even better `size` would really be an int or array again (for field type) :)

mateuszbaran: I've implemented your idea and it works well, thanks :slightly_smiling_face: . I like it.

kellertuer: Nice! I like the result as well – and it was more by accident that I saw that idea.

kellertuer (approved):

PR #161, opened 2023-07-07
Make showing cached bases more robust

mateuszbaran:
A tiny improvement. I've discovered this issue when working with new basis types. Having an error in printing is somewhat irritating so I think having a less informative text is still better.


kellertuer (approved):
A bit hacked, but I see what you mean and agree with your ideas.

PR #167, opened 2023-09-07
Unify exp and retract, log and inverse_retract

### LINKED ISSUE #159: Unify `exp` and `retract`
as noted in #158, I saw that currently our default dispatch for `exp` and `retract` differ, when it comes to treating the last positional argument `t`.
Here I would like to give a thorough overview about the state and think/discuss about the solution

## exp and retract

`exp` follows the scheme “allocate first”, i.e.both with and without `t` they allocate and “pass over” to `exp!`

`retract` differs as
* `retract(M, p, X, m)` calls `retract(M, p, X, 1.0, m)`
* `retractM, p, t, m)` passes to `_retract` (level 2) which dispatches on the method and `retract_method` allocates and calls `retract_method!`– but they do not exist without `t`.

## exp! and retract!

We have that
```
exp!(M::AbstractManifold, q, p, X, t::Number) = exp!(M, q, p, t * X)
```

So `exp!` fuses `t` and `X` by default. This is nice for the user since implementing `exp!(M, q, p, X)` is all they need to do, no need to think about `t`.

For `retract`this is the other way around. `retract!(M, q, p, X, m)` adds `1.0` as second to last element, i.e. calls `retract!(M::AbstractManifold, q, p, X, t::Number, method)` – and this passes down to level 2 and 3 – `_retract!` and `retract_method!`, which never exist without `t`.

## Summary

* For `exp` it is _necessary_ to implement `exp!(M, q, p, X)` and nice (maybe speedup?) to implement `exp!(M, q, p, X, t)`, but this method alone is not enough.
* For `retract` it is _necessary_ to implement `retract_method!(M, q, p, X, t)`, implementing the one without `t` is not possible.

## Own opinion

I feel that `exp!` is nice the way it is and an easy entry level. I feel that `retract!`compared to that is a bit unexpected. So I would prefer to keep `exp` as is; maybe if there is a way to keep the current state _and_ that _also_ `exp!(M, q, p, X, t)` alone would suffice, that would be the most neat one

Maybe this would also be the best for retract – but I have no good idea how to accomplish that _either_ of them would suffice.
This would at least mean that the “without `t`”variant has to dispatch through level 2 and 3 (a bit more of boiler plate code).

kellertuer:
The unification concerns the “dispatch flow”

Now both
* allocate first, dispatch on eat ! method
* exp and retract fuse `t` and `X` “last”

This way

1) a performance aware user can implement the `t`, `X` variant, otherwise the fused version with just X is enough to implement
2) We only have a layer-2-dispatch-tree for the inlace variant; this reduces code _and_ is nice for the user, since when they want to implement the allocating variant, they anyways have the specific `M` and `method` types, such that there should not be problems with ambiguities.

This is till WIP, since I have to check that all tests are fine till and a bit that manifolds.jl also still works fine ;)
This resolves #159.


Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "ManifoldsBase"
 uuid = "3362f125-f0bb-47a3-aa74-596ffd7ef2fb"
 authors = ["Seth Axen <seth.axen@gmail.com>", "Mateusz Baran <mateuszbaran89@gmail.com>", "Ronny Bergmann <manopt@ronnybergmann.net>", "Antoine Levitt <antoine.levitt@gmail.com>"]
-version = "0.14.11"
+version = "0.15"
```
mateuszbaran, line 4:
```suggestion
version = "0.15.0"
```

kellertuer:
I even found a small bug where some vector fall back of retract accidentally returned X instead of q (though it took me an hour to narrow it down).

Good points:
* The tests are running fine again
* the amount of code is reduced drastically
* one can still define individual retractions – just one has to be precise in parameter types if one skips a layer (and for the allocating ones these layers no longer exist, so one has to be precise)
* For Manopt this update is non-breaking
* For Manifolds.jl it seems to break only some of the number-manifolds, where we used the now-removed `retract`-layers (Circle, Euclidean)

But besides the small break on allocating manifolds, this seems to not only unify code but also simplify our code base!

Still ToDo
* [ ] check and update documentation
* [ ] check test coverage.

test/default_manifold.jl
```diff
@@ -753,8 +752,8 @@ Base.size(x::MatrixVectorTransport) = (size(x.m, 2),)
         @test X3 == log(M, p, q)
         @test log!(M, X3, p, q) == log(M, p, q)
         @test X3 == log(M, p, q)
-        @test inverse_retract(M, p, q, CustomDefinedInverseRetraction()) == -2 * Y
-        @test distance(M, p, q, CustomDefinedInverseRetraction()) == 2.0
+        @test inverse_retract(M, p, q, CustomDefinedInverseRetraction()) == -Y
+        @test distance(M, p, q, CustomDefinedInverseRetraction()) == 1.0
```
mateuszbaran, line 756: I think you should maybe modify `CustomDefinedInverseRetraction` to keep the old test the same?

mateuszbaran:
I think this looks quite good. It looks like updating Manifolds.jl later won't be much of a problem :slightly_smiling_face: .

mateuszbaran:
BTW, if we make a breaking release of ManifoldsBase.jl maybe you could also propose a change to https://github.com/JuliaManifolds/ManifoldsBase.jl/pull/119 that you wanted? Ref. https://github.com/JuliaManifolds/Manifolds.jl/issues/438

kellertuer:
Sure, will recheck the test tomorrow and yes, we can surely also introduce an inverse retraction based distance approximation, just that I am missing a neat name for it.

test/default_manifold.jl
kellertuer: took me a while to see that I probably implemented the mutating variant wrongly, but I got it back from the allocating one I deleted, so sure, returning to the modified form is better for sure.

kellertuer:
I thought a bit about the arguments with `distance`, the main problem I have is, that I do not have a good idea for a name. So we could consider distance as high-level as e.g. the algorithms in Manopt.jl and keep its name and the argument (though I said something different before); just because if there is no good name, then there is no use in renaming it.

kellertuer:
@mateuszbaran you introduced a specific function definition once for `mid_point` (and only that) to work on zero-dimensional arrays on Julia pre 1.1.

None of the tests ran through that, I even tried to trigger this function in a Julia 1.0 session but could not. I removed it in the last commit. Can you check that this is ok or otherwise write a test to cover that case?

mateuszbaran:
IIRC it was due to some bug on Julia 1.0. I think that method could be removed and Julia requirement bumped to 1.6 here?

kellertuer:
Since it was a `mid_point!` function we did not loose coverage due to my rework (it might have skipped a few allocating ones due to the allocate first strategy), so my interpretation is, that the bug might be fixed.
What du you think?

But I am also fine to use this breaking change to remove support for 1.0 – through it is nice to support the interface even on very old versions, none of our other packages supports 1.0 anymore, either.

mateuszbaran:
ManifoldsBase.jl isn't very useful without other libraries so I think it's fair to remove support for 1.0. After all, old versions that work with 1.0 will still be available.

kellertuer:
Ok, been there, done that.

Should we merge this and wait for a few more breaking things before registering 0.15? I could work on the metric manifold next (just have to remember, where I stopped and why).

mateuszbaran:
I wouldn't merge it until we are closer to releasing Manifolds.jl 0.9. Maybe after I finish and you review the key parts of my PR? We don't really need this merged earlier, and merging it will just make tagging new versions of ManifoldsBase.jl 0.14.x more complicated.

kellertuer:
Ok, sounds fair.
If we have a few of these, we could think of an interims `dev-0.15` branch, but you are right for one PR we can also just wait.

I can check your PR soon-ish – but for now it is still a draft, right?

mateuszbaran:
Yes, my PR is still a draft but some parts are more or less finished. I think you can check the changes related to optional static size and removal of `ProductRepr`. It's really just one pattern applied to all manifolds which should be relatively easy to review and would cover a lot of code affected by that PR. I've also started a NEWS file which you can also check :slightly_smiling_face: . Other parts still need more work (by my current estimate they should be mostly ready by the end of September).

kellertuer:
Great, will do.

Did you see that I added the `TypeParameter` to the docs here? Please check and improve if I missed something :)

mateuszbaran:
Thanks!

I haven't checked those docs yet, but I will :slightly_smiling_face: .

docs/src/types.md
```diff
@@ -58,3 +58,14 @@ Modules = [ManifoldsBase]
 Pages = ["numbers.jl"]
 Order = [:type, :function]
 ```
+
+## [Type Parameter](@id type-parameter)
+
+An [`AbstractManifold`](@ref) usually has a type parameter that determines e.g. its siye or
+more precisely the [`manifold_dimension`](@ref). The [`TypeParameter`](@ref ManifoldsBase.TypeParameter) offers the flexibility
+to have this parameter either as type parameter or a field
```
mateuszbaran, line 66:
```suggestion
Concrete [`AbstractManifold`](@ref)s usually correspond to families of manifolds that are parameterized by some numbers, for example determining their [`manifold_dimension`](@ref). Those numbers can either be stored in a field or as a type parameter of the structure. The [`TypeParameter`](@ref ManifoldsBase.TypeParameter) offers the flexibility
to have this parameter either as type parameter or a field.
```

kellertuer: Nice, yours sounds better! We could take the chance to explain advantages of each approach?

As a parameter we can dispatch on it but it might lead to more (pre)compilations
As a field we avoid recompilations, but we can not dispatch.

mateuszbaran: Good idea, I think your explanation is good. Except maybe that type parameter definitely leads to more compilation :wink: . I guess you could use exp on rotation matrices as an example.

kellertuer:
This was reasonably easy to merge. Let me know when you have checked this on Manifolds.jl

mateuszbaran:
Something broke with `Circle`. `inverse_retract(M::Circle{ℝ}, p::Float64, q::Float64, m::LogarithmicInverseRetraction)` used to work but now it tries to allocate (which obviously fails). Shouldn't it forward to `log` instead of `inverse_retract!`?

mateuszbaran:
There is a similar issue with `retract`, for example `retract(M::Euclidean{Tuple{}, ℝ}, p::Float64, X::Float64, m::ExponentialRetraction)` calls `retract!` instead of `exp`.

kellertuer:
Ah interesting, that is a case we did not take into account. But then maybe we have to reintroduce `_retract` that dispatches on the method (either passing to exp or allocating and to retract!)

mateuszbaran (approved):
I've adapted Manifolds.jl PR to this change and tests are passing.

kellertuer:
Cool! Then I will merge this now into main – then merge main into the last PR :)

PR #168, opened 2023-09-22
fix tangent vector allocations (nearly)

kellertuer:
We still have one problem:

I tried to define `allocate_result(M::AbstractManifold, ::typeof(rand), p)` (commented out for now) but that is ambigous to the decorator path of allocations. The decorator path however would always pass for FixedRank to the Embedding. So we would need the specific allocation anyways?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "ManifoldsBase"
 uuid = "3362f125-f0bb-47a3-aa74-596ffd7ef2fb"
 authors = ["Seth Axen <seth.axen@gmail.com>", "Mateusz Baran <mateuszbaran89@gmail.com>", "Ronny Bergmann <manopt@ronnybergmann.net>", "Antoine Levitt <antoine.levitt@gmail.com>"]
-version = "0.14.11"
+version = "0.14.12"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
--- a/src/ManifoldsBase.jl
+++ b/src/ManifoldsBase.jl
@@ -854,21 +854,33 @@ Optionally a random number generator `rng` to be used can be specified. An optio
 
 """
 Random.rand(M::AbstractManifold)
+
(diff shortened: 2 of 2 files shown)


kellertuer:
Ah I think I found a way, it was mainly TangentSpaceAtPoint where I will define in a next PR on Maniflds.jl

```
allocate_result(M::TangentSpaceAtPoint, ::typeof(rand)) = zero_vector(M.fiber.manifold, M.point)
```

and then we have to check whether the allocation that include `p` would be nice to have

```
allocate_result(M::AbstracManifold, ::typeof(rand), p) = zero_vector(M, p)
```

but that one (though nice) includes a fight of ambiguities that at least I am not able to resolve (with line 91 in decorator_trait.jl)

mateuszbaran:
Yes, `allocate_result(M::AbstracManifold, ::typeof(rand), p) = zero_vector(M, p)` would lead to a lot of ambiguities. The only sane solution I can think of is adding an `if f === rand` to `allocate_result(M::AbstractManifold, f, x...)`, so that there is no dispatch. But it's ugly so not great either.

kellertuer:
Most will not notice, because currently that is dispatched to the embedding and a default p of representation size allocated there. But this is the one thing failing here; for now defining that in FixedRank (and TangentSpaceAtPoint) does resolve this as well.

kellertuer:
The three lines added here are the same as on the other PR, where it seems a bug on 1.1 was fixed so our new method for mid point does no longer seem necessary.

mateuszbaran:
I would worry about these lines, soon we will just drop Julia 1.0 and those lines anyway.

kellertuer:
yes sure it will be fixed with the other still open PR.

kellertuer:
Just ran the Manifold tests completely and what does fail are PositiveNumbers, some interplay between rand! and the rug it seems? Not sure. I fear I again do not fully understand why that fails over on Manifolds.

The question is then, do we check and fix that here or do we adopt PositiveNumbers?

Issue #135, opened 2022-11-09
Change function signature of `is_` functions to be consistent with `isapprox`.


kellertuer:
To be consistent with the newly planned error mode, these two positional arguments of `is_point` (for the first) and `is_vector` these two should become keyword arguments.
This should also be done for the internal `check_` functions also for consistency with the `isapprox` and `check_approx` functions,
see also discussions in #134.

kellertuer:
...somehow we did not do this in 0.14.0 ;)

mateuszbaran:
I forgot to check :/.

PR #169, opened 2023-09-30
Define Meta Manifolds already in ManifoldsBase

kellertuer:
This allows the user to even use a TangentSpaceAtPoint when only depending on ManifoldsBase.
I would like to have this for defining proper/nice/generic subproblems in tangent spaces in Manopt.jl (e.g. for the current trust region rework).

This is breaking (hence also for 0.15) in the sense that with this version otherwise the VectorBundle is defined twice (which kills precompiation).

# To do
* [x] moved the `ArrayPartition` sections to a weak dependency with `RecursiveArrayTools`.
* [x] We now have a first (non-standard-library) dependency here_ Requires.jl, but I think that is ok
* [x] check documentation
* [x] introduce DocumenterCitations here as well
* [x] bump Documenter to 1.0
* [x] check test coverage


mateuszbaran:
Hmm... since it's breaking, maybe let's move the implementation of fiber bundles from https://github.com/JuliaManifolds/Manifolds.jl/pull/642 here? And Product manifolds as well.

kellertuer:
Sure, why not. I was not aware you worked on the Fibre Bundle in that PR.

mateuszbaran:
It's in the PR's title :smile:

mateuszbaran:
Will you make the update or prefer me to do that?

kellertuer:
Since those are your changes (and it is harder for me to track them down in your PR), maybe you could add them here? Feel free to add product and power as much as possible here, too, I'd love to have them here as well.

mateuszbaran:
OK, sure, I will do it then.

kellertuer:
What do you think about the extension here? Or should the few functions using `ArrayPartition`s stay in `Manifolds` instead, where we have that dependency anyways?

mateuszbaran:
I prefer to have that extension here.

kellertuer:
Nice, I do not have a strong preference, but I agree, that this puts all VectorBundle-Fun in one place. Then Iwe should add the Revise dependency here as well (that's what's missing still).

kellertuer:
I just spend quite a bit of time with the docs. Citations should be fine already – but the docs still have 8 doctorings that we never included somewhere. I only fixed about a few and these are still missing. Test coverage – I started with that but I am not yet sure how far I got.

kellertuer:
Maybe we should also rethink the docs a bit when fixing the remaining ones – the manifolds section is by now very long – and might be better to be restructured.

mateuszbaran:
Maybe let's update docs when new things are ready? I'm in the middle of moving metamanifolds here.

kellertuer:
Sure, that's one of the reasons I waited with fixing :)

ext/ManifoldsBaseRecursiveArrayToolsExt.jl
```diff
@@ -3,92 +3,72 @@ module ManifoldsBaseRecursiveArrayToolsExt
...
-    return throw(DomainError(s, "unknown component $s on $M."))
-end
-
-"""
-    setindex!(p::ArrayPartition, val, M::VectorBundle, s::Symbol)
-    p[M::VectorBundle, s] = val
-
-Set the element(s) at index `s` of a point `p` on a [`VectorBundle`](@ref) `M` to `val` by
-using the symbols `:point` and `:vector` for the base and vector component, respectively.
+include("ProductManifoldRecursiveArrayToolsExt.jl")
```
kellertuer, line 71: could we maybe either put these two or even all three files in a subfolder? For a short moment I thought these are now 3 extensions while it is only one. I think in Manifolds.jl we did a folder of the name of the Extension.

But cool that also product comes here!

mateuszbaran: Sure, I will move them to a folder.

kellertuer: I will also take the chance to review a few (currently a bit short) doc strings in the parts we move.

But I really like that with this PR – ManifoldsBase is: A nice interface for manifolds and a set of meta manifolds you can directly use as well. Super cool.

mateuszbaran: Cool, thanks. I think moving those parts is a good idea.

test/test_sphere.jl
```diff
@@ -51,6 +114,38 @@ function ManifoldsBase.project!(::TestSphere, Y, p, X)
...
 
+function Random.rand!(M::TestSphere, pX; vector_at = nothing, σ = one(eltype(pX)))
+    if vector_at === nothing
+        project!(M, pX, randn(eltype(pX), representation_size(M)))
+    else
+        n = σ * randn(eltype(pX), size(pX)) # Gaussian in embedding
+        project!(M, pX, vector_at, n) #project to TpM (keeps Gaussianness)
+    end
+    return pX
+end
```
mateuszbaran, line 125: We've recently had a discussion about not needing the `rng`-free rand but I don't remember what was the conclusion. It seems to be needed now though?

kellertuer: I think we o have the fallback from rig-free to rng-included ones – I feel a bit like exp/exp! here. A beginner might implement exp (or the rng-free rand!) the advanced the other two and we should always do the second (and half beginners boost their code?). So I would switch this?

mateuszbaran: What do you mean by switching this?

kellertuer: Implement the rng-version here ;)

mateuszbaran: Both are needed actually.

kellertuer: Oh, O had hoped we reduced that to one of the versions to reduce duplicate code.

mateuszbaran: We did not, that would require more work. Ref. https://github.com/JuliaManifolds/Manifolds.jl/issues/652 .

kellertuer: Then I remembered that wrongly (sorry, for today I am entering braindead mode by now, but teaching for the next project is nearly prepared ;) )

mateuszbaran: No problem :slightly_smiling_face: . I guess we will just accept this duplication for now.

kellertuer:
It seems for now that flat and sharp were defined in Manifolds.jl only, shall we move them over as well?

mateuszbaran:
> It seems for now that flat and sharp were defined in Manifolds.jl only, shall we move them over as well?
> 
> _edit1:_ Same for flat and the RiesRepresenterCotangent stuff...

Yes, my mistake, I've moved too much here. `flat`, `sharp` and cotangent stuff should not be here.

kellertuer:
I would not mind having them here, but sure we can also move them back, since they are only needed for the cotangent parts and those are not yet so mature I think, that they should be moved here. Could you move that back?

mateuszbaran:
I've moved them back to Manifolds.jl

mateuszbaran:
List of documentation issues looks sane now :slightly_smiling_face:

kellertuer:
Yes, and I also have a idea (that I started as well) how to restructure the docs – and I will check the docs we moved whether they are detailed enough; hopefully will find a bit of time tomorrow for that.

Thanks :)

mateuszbaran:
Cool, I'll check that when it's ready :slightly_smiling_face:

kellertuer:
Woah! We made it to all-lines-covered? That's awesome.

kellertuer:
I reworked retractions already, so that all generic dispatch and types are (as before) in the retractions file; I will do the same for the inverse retractions and the vector transport.

There are two missing Docs Errors that I could not yet narrow down and it seems the Fiber.jl docs are included twice somewhere and I can not see where that happens? Will have to check for that as well.

When there are no errors left I would still like to check those docs that are really short and provide a bit more details there, For example 

```

"""
    TensorProductType(spaces::VectorSpaceType...)

Vector space type corresponding to the tensor product of given vector space
types.
"""
struct TensorProductType{TS<:Tuple} <: VectorSpaceType
    spaces::TS
end
```

I have no clue what this type is for and that is does, the first line is not the type but the constructor and the sentence in the docs does not help me to understand this either.

mateuszbaran:
Sounds good. `TensorProductType` is still experimental so I will move it back to Manifolds.jl.

src/VectorFiber.jl
```diff
@@ -0,0 +1,279 @@
...
+"""
+TangentSpaceAtPoint(M::AbstractManifold, p) = VectorSpaceAtPoint(M, TangentFiber, p)
+
+"""
+    TangentSpace(M::AbstractManifold, p)
+
+Return a [`TangentSpaceAtPoint`] representing tangent
+space at `p` on the [`AbstractManifold`](@ref) `M`.
+"""
+TangentSpace(M::AbstractManifold, p) = TangentSpaceAtPoint(M, p)
```
kellertuer, line 77: I currently find the use of `TangentSpace` at least misleading.
It is used here as a short hand for `TangentSpaceAtPoint` (why were we so precise with that name anyways?) but it is also used as

```
const TangentSpace = TangentSpaceType()
```

And I am not yet sure what `TangentSpaceType is necessary for. Could we remove this naming double use somehow?

kellertuer: Also I carefully read the doc string of `VectorSpaceType` (very beginning of bases.jl) and I am not yet sure I understand why we need that and what it is necessary for (even less with the const in the last comment).

src/VectorBundle.jl
```diff
@@ -0,0 +1,461 @@
+
+"""
+    VectorBundle{𝔽,TVS,TM,TVT}
```
kellertuer, line 3: To me the whole generality of Fiber bundle is not yet so clear, so here saying that it is just a special case is maybe not helping a user if it does not help me?

I would first prefer a short description of what this actually is before noting that it is a special case of another type.

src/VectorBundle.jl
```diff
@@ -0,0 +1,461 @@
...
+end
+
+function default_vector_transport_method(B::TangentBundle)
+    default_vt_m = default_vector_transport_method(B.manifold)
+    return FiberBundleProductVectorTransport(default_vt_m, default_vt_m)
+end
+
+"""
+    distance(B::BundleFibers, p, X, Y)
+
```
kellertuer, line 69: While maybe slightly outdates (the function defines this for VectorBundleFibers) now after vector bundle being a special case of fiber bundles – I am actually finally lost on Vector bundle fibers What are these?

src/VectorBundle.jl
```diff
@@ -0,0 +1,461 @@
...
+Distance between vectors `X` and `Y` from the vector space at point `p`
+from the manifold `B.manifold`, that is the base manifold of `M`.
+"""
+distance(B::VectorBundleFibers, p, X, Y) = norm(B, p, X - Y)
+
+function get_basis(M::TangentBundleFibers, p, B::AbstractBasis{<:Any,TangentSpaceType})
+    return get_basis(M.manifold, p, B)
+end
+
+function get_coordinates(M::TangentBundleFibers, p, X, B::AbstractBasis)
```
kellertuer, line 79: These should also all be documented. I can surely help, but would first have to under stand
vector bundles
fiber bundles
vector bundle fibres and all these...

src/VectorBundle.jl
```diff
@@ -0,0 +1,461 @@
...
+function inner(B::FiberBundle, p, X, Y)
+    px, Vx = submanifold_components(B.manifold, p)
+    VXM, VXF = submanifold_components(B.manifold, X)
+    VYM, VYF = submanifold_components(B.manifold, Y)
+    # for tangent bundle Vx is discarded by the method of inner for TangentSpaceAtPoint
+    # and px is actually used as the base point
+    return inner(B.manifold, px, VXM, VYM) + inner(FiberAtPoint(B.fiber, px), Vx, VXF, VYF)
+end
+
+function _inverse_retract(M::FiberBundle, p, q, ::FiberBundleInverseProductRetraction)
```
kellertuer, line 147: here, the corresponding `inverse_retract` should be documented.

src/VectorBundle.jl
```diff
@@ -0,0 +1,461 @@
...
+function project(B::BundleFibers, p, X)
+    Y = allocate_result(B, project, p, X)
+    return project!(B, Y, p, X)
+end
+
+function project!(B::TangentBundleFibers, Y, p, X)
+    return project!(B.manifold, Y, p, X)
+end
+
+function _retract(M::VectorBundle, p, X, t::Number, ::FiberBundleProductRetraction)
```
kellertuer, line 262: This should be documented.

src/VectorBundle.jl
```diff
@@ -0,0 +1,461 @@
...
+        xp,
+        Xp + VXF,
+        VXM,
+        B.vector_transport.method_point,
+    )
+    copyto!(B.manifold, xq, xqt)
+    return q
+end
+
+function retract_sasaki!(B::TangentBundle, q, p, X, t::Number, m::SasakiRetraction)
```
kellertuer, line 302: the corresponding retract should be documented.

src/VectorBundle.jl
```diff
@@ -0,0 +1,461 @@
...
+        copyto!(B.manifold, X_k, p_kp1, X_kp1)
+        copyto!(B.manifold, Y_k, p_kp1, Y_kp1)
+        copyto!(B.manifold, Z_k, p_kp1, Z_kp1)
+    end
+    copyto!(B.manifold, xq, p_k)
+    copyto!(B.manifold, Xq, xq, X_k)
+    return q
+end
+
+function representation_size(B::TangentBundleFibers)
```
kellertuer, line 336: This should be documented

src/VectorBundle.jl
```diff
@@ -0,0 +1,461 @@
...
+Base.show(io::IO, vb::TangentBundle) = print(io, "TangentBundle($(vb.manifold))")
+
+@inline function allocate_result(B::TangentBundleFibers, f::typeof(zero_vector), x...)
+    return allocate_result(B.manifold, f, x...)
+end
+
+"""
+    fiber_bundle_transport(fiber::FiberType, M::AbstractManifold)
+
+Determine the vector tranport used for [`exp`](@ref exp(::FiberBundle, ::Any...)) and
```
kellertuer, line 355: this one I do not fully understand – how much is this different from a default_vector_transport_method? And why is the name without `_method`?

src/VectorBundle.jl
```diff
@@ -0,0 +1,461 @@
...
+[`log`](@ref log(::FiberBundle, ::Any...)) maps on a vector bundle with fiber type
+`fiber` and manifold `M`.
+"""
+fiber_bundle_transport(::VectorSpaceType, ::AbstractManifold) = ParallelTransport()
+
+function vector_space_dimension(B::TangentBundleFibers)
```
kellertuer, line 361: These should be documented.

src/VectorBundle.jl
```diff
@@ -0,0 +1,461 @@
...
+    return manifold_dimension(B.manifold)
+end
+function vector_space_dimension(B::VectorBundleFibers)
+    return vector_space_dimension(B.manifold, B.fiber.fiber)
+end
+
+function vector_transport_direction(M::VectorBundle, p, X, d)
```
kellertuer, line 368: no method required?

src/VectorFiber.jl
```diff
@@ -0,0 +1,279 @@
+
+"""
```
kellertuer, line 2: I do (also here) not fully understand what this is for. The doc string is super short and does not help – What do we need this type for?

src/VectorFiber.jl
```diff
@@ -0,0 +1,279 @@
...
+end
+
+const TangentFiberType = VectorSpaceFiberType{TangentSpaceType}
+
+const TangentFiber = VectorSpaceFiberType{TangentSpaceType}(TangentSpace)
+
+"""
+    VectorBundleFibers{TVS,TM}
+
+Alias for [`BundleFibers`](@ref) when the fiber is a vector space.
```
kellertuer, line 23: does BundleFibers still exist? (You might notice I am a bit lost in the zoo of types involving the 3 words bundle-fiber-vector and arbitrary permutations of these and sometimes added with a `Type` at the end).

src/VectorFiber.jl
```diff
@@ -0,0 +1,279 @@
...
+
+"""
+    TangentSpaceAtPoint{𝔽,M}
+
+Alias for [`VectorSpaceAtPoint`](@ref) for the tangent space at a point.
+"""
+const TangentSpaceAtPoint{𝔽,M} =
+    VectorSpaceAtPoint{𝔽,M,TangentSpaceType} where {𝔽,M<:AbstractManifold{𝔽}}
+
+"""
```
kellertuer, line 63: Should be documented as a constructor of the type above?

src/VectorFiber.jl
```diff
@@ -0,0 +1,279 @@
...
+    distance(M::TangentSpaceAtPoint, p, q)
+
+Distance between vectors `p` and `q` from the vector space `M`. It is calculated as the norm
+of their difference.
+"""
+function distance(M::TangentSpaceAtPoint, p, q)
+    return norm(M.fiber.manifold, M.point, q - p)
+end
+
+function embed!(M::TangentSpaceAtPoint, q, p)
```
kellertuer, line 93: the corresponding `embed` should be documented.

src/VectorFiber.jl
```diff
@@ -0,0 +1,279 @@
...
+calculated as their sum.
+"""
+exp(::TangentSpaceAtPoint, ::Any, ::Any)
+
+function exp!(M::TangentSpaceAtPoint, q, p, X)
+    copyto!(M.fiber.manifold, q, p + X)
+    return q
+end
+
+fiber_dimension(M::TangentBundleFibers) = manifold_dimension(M.manifold)
```
kellertuer, line 113: This should be documented.

src/VectorFiber.jl
```diff
@@ -0,0 +1,279 @@
...
+
+function Base.show(io::IO, ::MIME"text/plain", TpM::TangentSpaceAtPoint)
+    println(io, "Tangent space to the manifold $(base_manifold(TpM)) at point:")
+    pre = " "
+    sp = sprint(show, "text/plain", TpM.point; context = io, sizehint = 0)
+    sp = replace(sp, '\n' => "\n$(pre)")
+    return print(io, pre, sp)
+end
+
+function vector_transport_to!(
```
kellertuer, line 245: this should be documented (the corresponding `vector_transport_to`=

kellertuer (commented):
Sorry for so many comments but I am a bit lost between
fiber bundles, bundle fibres, vector bundles and vector bundle fibers and their correspoding types... it also seems you have either switched order of these words somewhen or added Fiber before bundle or so at some point?

src/VectorBundle.jl
mateuszbaran: Good point. What do you think about re-using old `VectorBundle` docs here? Or do you want more details?

Fiber bundles is not something commonly encountered in numerical applications but I think it's a good abstraction with a lot of supporting theory: unit tangent bundle, principal bundles, normal bundles, etc, with rich enough theory of metrics and connection forms that useful stuff can be done in many cases.

kellertuer: Then let's mention a bit more details here, I would prefer either a short description of what that is, and/or a reference?

src/VectorBundle.jl
mateuszbaran: On trivial fiber bundles (and basically all fiber bundles we want to work on are trivial) operations generally decompose into manifold-part and fiber-part. `BundleFibers` is used to the fiber-part computations. It's a bit of abuse of notation actually and I'm not sure it's the right structure but it works well enough so far. So, here we don't consider `BundleFibers` as a manifold exactly, but the fiber attached to point `p` in the fiber bundle, and `X` and `Y` are fiber-projections of a point on the fiber bundle.

mateuszbaran:
> Sorry for so many comments but I am a bit lost between fiber bundles, bundle fibres, vector bundles and vector bundle fibers and their correspoding types... it also seems you have either switched order of these words somewhen or added Fiber before bundle or so at some point?

Hm, I can try to explain but I'm not sure how much you understand the old `VectorBundle` design? The only change is that I made `VectorBundle` a special case of a more general `FiberBundle`, for cases where fibers are not vector spaces.

src/VectorBundle.jl
mateuszbaran: I will work on the docs over the next few days. I will try to make it comprehensible.

kellertuer: I am happy to help, for the ones I noted here, I just felt a bit lost.

Thanks for working on those, I think this is a great opportunity to extend them a bit.

src/VectorFiber.jl
mateuszbaran: `TangentSpaceType` is used to distinguish tangent bundle from other vector bundles, and tangent spaces at point from other spaces (at point). We could remove `TangentSpace` constant if you think that would be more clear but `TangentSpaceType` is necessary (though it could be renamed if needed). This particular method is definitely not needed.

Renaming `TangentSpaceAtPoint` to `TangentSpace` is possible but that would require also renaming other `AtPoint` types. I'm not sure if new names would be more or less confusing.

mateuszbaran: Do you think it would be better to rename `TangentSpaceType` to `TangentSpace` and replace `TangentSpace` with ``TangentSpace()`?

kellertuer: Hm, I think I would maybe do the opposite, rename the `AtPoint` things to without that, so we have a (user-facing) `TangentSpace` type that represents – a tangent space.

`TangentSpaceType` is something more technical / design decision, not that public, so I think there it is fine to have the more complicated name?
And the reason why `TangentSpace=TangentSpaceType()` is what we currently have is also something I do not understand especially if that later appears as a type but is actually an instance?

mateuszbaran: I see, I will try that.

kellertuer:
I carefully read the code for the Fiber bundle (to not be to special already when looking at the tangent bundle) and tried extending the docs. I ran into a few questions, since I feel we store a few tings 2-3 times

1. Why do we need to store `FiberBundle.type` is it not enough to have that as a parameter of the fiber bundle? For dispatch we would just need it as a parameter.
2. Why do we beed to store a single fiber (I would prefer to rename `BundleFibers` to just `Fiber`) 
3. Why does the `fiber` _again_ store the manifold and the type? Is that not double?

If we remove the `fiber` field from `FiberBundle` the fiber bundle would still contain the same information – it knows its type and its manifold. Or in other words, the only thing that 3 does is introduce redundancy (no inconsistency, since you took the same bound type).

Also in the fiber bundle the fiber type is in the field `type` while in the `BundleFiber` it is the field `fiber` ? 

So I would maybe do 
´``Julia
struct FiberBundle{
    𝔽,
    TF<:FiberType,
    TM<:AbstractManifold{𝔽},
    TVT<:FiberBundleProductVectorTransport,
} <: AbstractManifold{𝔽}
    manifold::TM
    vector_transport::TVT
end
```

and 

```
struct Fiber{TF<:FiberType,TM<:AbstractManifold}
    manifold::TM
end
``` 

But here even further – isn't `Fiber` (like `TangentSpace`) also a manifold? And should it therefore not also have a `𝔽` field upfront?

I am still not sure why the fiber bundle has to store its vector transport (we also do not do that for other manifolds). So we could maybe even leave that out as well? any vector transport call would either use the default (where I would store the one you currently want to store in the fiber bundle) or specify another one.

I hope this is reasonable – and if so, I am happy to rework this and write the docs, because if my concerns here are right, I think I understand the concept :)

**Comment by mateuszbaran** [comment]
> If we remove the `fiber` field from `FiberBundle` the fiber bundle would still contain the same information – it knows its type and its manifold. Or in other words, the only thing that 3 does is introduce redundancy (no inconsistency, since you took the same bound type).

IIRC some redundancy is there because it was easy to just do M.fiber rather than having a code that reconstructs the object and worry about performance of said reconstruction. Maybe it's no longer an issue, I will investigate.

> I am still not sure why the fiber bundle has to store its vector transport (we also do not do that for other manifolds).

This is actually a bit underdeveloped part. Essentially, we want some structure on a fiber bundle -- at least a connection. This "vector transport" is actually about specifying what connection should be used. It's usually restricted to principal and vector bundles (see https://ncatlab.org/nlab/show/connection+on+a+bundle but please don't scroll too far down :sweat_smile:  ) but there may be valid examples of connections outside of these two types.

> _edit:_ Hm, now I am confused again. Why do we have `FiberBundle` (the large thing) and `BundleFibers` (which I thought would be a single one?) but also `FiberAtPoint`? I am now again lost what this “middle type” is for. T

I will try to simplify this part -- I don't like it either.

**Comment by kellertuer** [comment]
> > If we remove the `fiber` field from `FiberBundle` the fiber bundle would still contain the same information – it knows its type and its manifold. Or in other words, the only thing that 3 does is introduce redundancy (no inconsistency, since you took the same bound type).
> 
> IIRC some redundancy is there because it was easy to just do M.fiber rather than having a code that reconstructs the object and worry about performance of said reconstruction. Maybe it's no longer an issue, I will investigate.

I am happy to help with docs, but for that I would like to know why and when redundancy is necessary, to me it seems like the above idea with 2 types is not only clearer and closer to there manifolds, but also simpler in many places.

> 
> > I am still not sure why the fiber bundle has to store its vector transport (we also do not do that for other manifolds).
> 
> This is actually a bit underdeveloped part. Essentially, we want some structure on a fiber bundle -- at least a connection. This "vector transport" is actually about specifying what connection should be used. It's usually restricted to principal and vector bundles (see https://ncatlab.org/nlab/show/connection+on+a+bundle but please don't scroll too far down 😅 ) but there may be valid examples of connections outside of these two types.

Hm but manifolds we also equip with a metric a connection a vector transport and so on … all of that implicitly or with the `default_` functions. Why is that then not possible for fiber bundles?
Or phrased differently – is there a good reason to deviate from the default design pattern _all_ of our manifolds and meta manifolds follow until now?
 
> > _edit:_ Hm, now I am confused again. Why do we have `FiberBundle` (the large thing) and `BundleFibers` (which I thought would be a single one?) but also `FiberAtPoint`? I am now again lost what this “middle type” is for. T
> 
> I will try to simplify this part -- I don't like it either.

Then let‘s not do it and remove this „middle type“

**Comment by mateuszbaran** [comment]
> Hm but manifolds we also equip with a metric a connection a vector transport and so on … all of that implicitly or with the `default_` functions. Why is that then not possible for fiber bundles?
> Or phrased differently – is there a good reason to deviate from the default design pattern _all_ of our manifolds and meta manifolds follow until now?

One way to look at it is that it's mostly to not bother with a decorator struct for changing the connection. I'm really not excited about having to design another decorator just to be able to change the connection on fibers when there is a very simple alternative.

**Comment by kellertuer** [comment]
Hm,
because with the two types I sketched above I think I could easily implement most functions that are there in the new form.

On the other hand, the connection decorator will be needed at some point in time (analogously to the metric decorator) anyways.
And then there is my German-ness that likes to stick to design rules and ideas – and with these your idea of “this is the easy way to do this for now” conflicts a bit. But I do agree there is a bit of a tradeoff.

**Comment by kellertuer** [comment]
I feel my approach might be minimalistic (though sticking more to the other manifolds in struct design).

If you feel this is too much work on your side, I could reduce the current PR to what I think should suffice (including docs and testing) and we go from there add what you think is missing.

**Comment by mateuszbaran** [comment]
If you want it done quickly, then perhaps the best way to go about it is to keep fiber and vector bundles in Manifolds.jl for now? Fibers and tangent spaces are non-controversial so if that would be enough for Manopt then it might be a good compromise. And of course this PR also brings `ProductManifold` to ManifoldsBase.jl.

**Comment by kellertuer** [comment]
I am not super in a hurry, just that we have PRs open that depend on this.
But sure I would even be fine just having the tangent space here (though also having tangent bundle then would be nice) and we could keep the fibers in Manifolds.jl for now?

Though that would also split a few things I think. But since the design seems to not be finished, maybe really leaving out fibers here is a good idea – what I actually need is the tangent space – and it would be nice to just have it as `TangentSpace(M,p)` then.

**Comment by mateuszbaran** [comment]
> But sure I would even be fine just having the tangent space here (though also having tangent bundle then would be nice) and we could keep the fibers in Manifolds.jl for now?

It would be great if we could have `FiberAtPoint` (or just call it `Fiber`) here, and make `TangentSpace` a special case of `Fiber`. This would save us another rework when fiber bundles are finished. `BundleFibers` and `FiberBundle` wouldn't be included here.

> Though that would also split a few things I think. But since the design seems to not be finished, maybe really leaving out fibers here is a good idea – what I actually need is the tangent space – and it would be nice to just have it as `TangentSpace(M,p)` then.

OK, `TangentSpace` name is free now so we can use it instead of `TangentSpaceAtPoint`.

**Comment by kellertuer** [comment]
Ok, that we can do, but then also TangentBundle would not yet be moved here (since it is a special case of the VectorBundle)? That would also be fine with me then.

**Comment by mateuszbaran** [comment]
Yes, `TangentBundle` would remain in Manifolds.jl for now.

**Comment by kellertuer** [comment]
That's fine with me. I will rework this PR – hopefully tomorrow evening.

Before then registering 0.15, should we also fix #135 (maybe also the parts of https://github.com/JuliaManifolds/Manifolds.jl/issues/630 that can be resolved here)?

**Comment by mateuszbaran** [comment]
OK. I will do a bit of work on this PR today too.

#135 sounds like an easy change so it would be a good idea. I'd postpone https://github.com/JuliaManifolds/Manifolds.jl/issues/630 though because figuring out the right tolerances can be tricky.

**Comment by kellertuer** [comment]
Ok, fair enough :)

**Comment by mateuszbaran** [comment]
I did the rework. I've kept `fiber_type` in `Fiber` because it doesn't take any memory when it's an empty object anyway, and it's easier to pass it somewhere.

**Comment by mateuszbaran** [comment]
I haven't properly updated docs and `show` methods yet though.

**Comment by kellertuer** [comment]
> I did the rework. I've kept fiber_type in Fiber because it doesn't take any memory when it's an empty object anyway, and it's easier to pass it somewhere.

I want to check, whether I like all those places ;) But sure, if that is reasonable, we can keep it as a field, though I do like it as only a parameter ,maybe a bit more.

> I haven't properly updated docs and show methods yet though.

Sure, I can work on those tomorrow :)

File: src/VectorFiber.jl [hunk 20]
```diff
@@ -0,0 +1,246 @@
...
+    pre = " "
+    sp = sprint(show, "text/plain", TpM.point; context = io, sizehint = 0)
+    sp = replace(sp, '\n' => "\n$(pre)")
+    return print(io, pre, sp)
+end
+
+function vector_transport_to!(M::TangentSpace, Y, p, X, q, ::AbstractVectorTransportMethod)
+    return copyto!(M.manifold, Y, p, X)
+end
+
```
kellertuer, line 240:
```suggestion
```


I would remove this generic definition, unless there is a good reason. This would introduce ambiguities as soon as someone wants to actually define a specific vector transport I think?

src/VectorFiber.jl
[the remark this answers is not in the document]
mateuszbaran: I called it `VectorFiber` because not everything here relates to `TangentSpace` and I didn't feel like making another file for other vector fibers. On the other hand, putting everything in `Fiber.jl` felt like too much stuff in one place. Feel free to adjust it as you like.

src/VectorFiber.jl
mateuszbaran: OK, I don't exactly remember why it was needed and maybe it isn't now.

src/VectorFiber.jl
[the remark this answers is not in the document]
kellertuer: Hm, ok. I was thinking maybe putting everything of Fiber and VectorFiber into `Fiber.jl` and calling the rest here `TangentSpace.jl` might increase discoverability of the tangent space code, but in the end it is probably just a matter of taste.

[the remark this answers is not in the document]
mateuszbaran: OK, your idea also makes sense -- TangentSpace is the most important fiber so it deserves its own file.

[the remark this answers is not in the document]
kellertuer: Nice. Thanks for agreeing :)

src/VectorBundle.jl
mateuszbaran: This was removed from this PR.

src/VectorBundle.jl
mateuszbaran: This was removed from this PR.

src/VectorFiber.jl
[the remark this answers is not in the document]
mateuszbaran: Well, turns out we can't move `VectorSpaceFiber` to `Fiber.jl` because `VectorSpaceFiber` needs stuff from `bases.jl` but `bases.jl` needs stuff from `Fiber.jl`. I've made another file just for `VectorSpaceFiber`.

kellertuer:
...and it is even non-breaking? That's nice, though I forgot what the braking one is waiting for (besides maybe also resolving the one issue we mentioned already above).

I think a few variables could be unified and will check more documentation hopefully later today.

mateuszbaran:
This could maybe be considered technically non-breaking but new features here require breaking changes in Manifolds.jl to be usable and we drop Julia 1.0 here. I'd prefer to release this as a part of v0.15.

kellertuer:
hm oh, but you just changed the version from 15.0 to 14.x?

mateuszbaran:
Sorry, didn't mean to push it :/ . It was meant to be local to make `resolve` work in my environment.

kellertuer:
That is fine, then you just confused me with that version number – I am of course also fine with a 15.0 version here.

src/retractions.jl
```diff
@@ -1047,6 +1074,15 @@ retract_softmax!(M::AbstractManifold, q, p, X, t::Number)
 
 function retract_softmax! end
 
+"""
+    retract_sasaki!(M::AbstractManifold, q, p, X, t::Number, m::SasakiRetraction)
+
+Compute the in-place variant of the [`SasakiRetraction`](@ref) `m`.
+"""
+retract_pade!(M::AbstractManifold, q, p, X, t::Number, m::SasakiRetraction)
```
kellertuer, line 1082:
```suggestion
retract_sasaki!(M::AbstractManifold, q, p, X, t::Number, m::SasakiRetraction)
```

kellertuer: Or do I miss something here? The docs say `sasaki` as well, and it is similar to Padé though (since it also still passes down the retraction type),

kellertuer:
I worked a bit further on the docs, but I am again lost on quite a last reference left to `[`TangentBundle`](@ref)`  for the `SasakiRetraction` I am unsure how to rephrase this.

It also seems we introduce the `CotangentVectorSpaceType` but no `CotangentSpace` is that intentional? I think we could also introduce the cotangent space if we have the vector space type for the fiber already?

Those are the last two errors left before the docs render. Then I will check the strings further and which to add maybe

mateuszbaran:
> I worked a bit further on the docs, but I am again lost on quite a last reference left to `[`TangentBundle`](@ref)` for the `SasakiRetraction` I am unsure how to rephrase this.

Either remove the reference or link to Manifolds.jl docs?

> It also seems we introduce the `CotangentVectorSpaceType` but no `CotangentSpace` is that intentional? I think we could also introduce the cotangent space if we have the vector space type for the fiber already?

`CotangentSpaceType` has been here for some time but I actually don't need `CotangentSpace` so I didn't think about it. Feel free to add it.

kellertuer:
Hm remo

> > I worked a bit further on the docs, but I am again lost on quite a last reference left to `[`TangentBundle`](@ref)` for the `SasakiRetraction` I am unsure how to rephrase this.
> 
> Either remove the reference or link to Manifolds.jl docs?

Hm, removing makes the doc string a bit unreadable. So probably then the link over for Manifolds, though I am not the largest fan of that.

> 
> > It also seems we introduce the `CotangentVectorSpaceType` but no `CotangentSpace` is that intentional? I think we could also introduce the cotangent space if we have the vector space type for the fiber already?
> 
> `CotangentSpaceType` has been here for some time but I actually don't need `CotangentSpace` so I didn't think about it. Feel free to add it.

Ah ok :) Then let's also introduce that just for completeness.

kellertuer:
I think I fixed and updated the docs.

One thing we could unify a bit is how we address points / tangent vectors on the tangent space, since a point on `TpM` could be a `p` (but we do not want to use names double) or an `X`, similarly for the second choice `X` would already be taken, then maybe `Z,Y` for tangent vectors to the tangent space (since they are also tangent vectors / points again)?

mateuszbaran:
We could also maybe use more than one character for variable name in case of tangent spaces? I have an even worse problem for vector bundles and using single letters doesn't work at all.

kellertuer:
What? There exist more letters than p, q, X, Y, Z? Not that I have heard of! ;)

I would like to avoid small greek ones, since they are usually cotangent vectors, but besides that I am open do ideas for sure. Do you have something in mind? `A` and `B` for tangents to tangent vectors could work? Since they are still tangent vectors I would maybe stay with capital letters? `V` and `W` would also be nice I think.

mateuszbaran:
I will shock you even more: you can have argument names in Julia that are *two* or even *three* letters long! Such as `pX`, `p_M` or `X_T`.

mateuszbaran:
For example, how would you write `riemann_tensor!` signature for `TangentSpace` without indices and confusion?

kellertuer:
Probably `W1`, `W2`, `W3` as tangent vectors to `X` ? (in math then with ``W_1``)

But I am open to suggestions with other indices.

mateuszbaran:
I see. I was thinking about `X_T`, `Y_T`, `Z_T` to be more consistent with existing methods but `W1`, `W2`, `W3` works too.

kellertuer:
In math that would have to be an `\mathrm{T}` (since it is not a variable), sure we could do that. I would prefer numbering them, I think.

mateuszbaran:
OK, let's use numbers then.

kellertuer:
Nice, will try to find time tomorrow (since for today I started sketching `TangentSpaceModelproblems` already to make TrustRegions (and probably the ARC solver) nicer.

kellertuer:
I unified the notation. We should check that retractions (inverse retractions, vector transports) are properly passed down to the single factors in the product manifold (see linked issue from 0.14 / Manifolds)? But besides that this should be good to go :)

kellertuer:
I would like to sketch a small roadmap.
Once the is PR is done (just bumped test coverage hopefully) I would like to

1) Merge this first
2) merge it into the other PR and check whether we can introduce an “automatic product retraction pass down” (as discussed in https://github.com/JuliaManifolds/Manifolds.jl/issues/659) or at least for a few retractions (we could just generate a n-copy-productretraction)
3) Maybe even introduce `×` for retractions as well.
4) Merge that PR
5) Resolve https://github.com/JuliaManifolds/ManifoldsBase.jl/issues/135 with a third PR
6) register 0.15 🥳

mateuszbaran:
Sounds good :+1: . I more thing I'd like to do before merging this is making all tests pass (locally) in the Manifolds.jl rework PR.

kellertuer:
Ah, that sounds like a very good idea 👍

In the mean time I can do the error-keyword-rework already for #135 – that also mainly seems to be a few lines of code and a bit of test adaptions.

mateuszbaran:
Sure, that should be fairly simple.

kellertuer:
> Sure, that should be fairly simple.

Well, ± ambiguities and a few internals for embedded manifold or such that did change. but I think I am nearly there.

mateuszbaran:
I've updated `PowerManifold` so that it can also be optionally static. I opted to use field-size by default because for this manifold it almost never makes sense to store the size statically, IMO.

kellertuer:
I think I found a way to introduce retractions with non-product retractions as discussed in https://github.com/JuliaManifolds/Manifolds.jl/issues/659, the only “disadvantage” might be, that when margin this into https://github.com/JuliaManifolds/ManifoldsBase.jl/pull/167. we might have to repeat these for the `retract` as well.

This might be slightly bridging our layers, but layer one says “Note that all other parameters of a function should be as least typed as possible for all parameters besides the manifold.” so that is what I did here for the abstract retraction case and put it on layer 1; currently I also moved the `ProductRetraction` cases to layer 1, but maybe that is wrong and they should be back on Layer 2.

this is still missing a bit of testing, but I will first check which lines are missing, since also the new `cross` functions (analogue to product manifold one can now also create product retractions, inverse retractions, vector transports) and new show functions have to be tested.

mateuszbaran:
> I think I found a way to introduce retractions with non-product retractions as discussed in [JuliaManifolds/Manifolds.jl#659](https://github.com/JuliaManifolds/Manifolds.jl/issues/659), the only “disadvantage” might be, that when margin this into #167. we might have to repeat these for the `retract` as well.

OK, if this doesn't cause any ambiguities then I'm fine with it.

kellertuer:
> > I think I found a way to introduce retractions with non-product retractions as discussed in [JuliaManifolds/Manifolds.jl#659](https://github.com/JuliaManifolds/Manifolds.jl/issues/659), the only “disadvantage” might be, that when margin this into #167. we might have to repeat these for the `retract` as well.
> 
> OK, if this doesn't cause any ambiguities then I'm fine with it.

That is exactly why I moved them from L2 to L1, on L2 they caused too many (non-ignorable) ambiguities. I will then write a few tests now :)

kellertuer:
Oh, two of the function definitions I do not yet hit but I thought I would have. I will only have time tomorrow to check for these as well.

mateuszbaran:
Cool. With this latest fix, test are passing in Manifolds.jl PR. I will continue with docs and coverage there.

kellertuer:
Thanks for fixing one of my a-bit-too-fast copied method. I just extended (hopefully finally) test coverage.

I think now this should be good to go :)

kellertuer:
We have one false-positive uncovered line left, so code coverage is as best as possible.

mateuszbaran:
Yes, I think this is pretty much ready and we can work on releasing 0.15.0. I think first, we need to merge all three breaking PRs into one branch and check if the Manifolds.jl PR needs some additional work.

mateuszbaran (approved):

kellertuer:
Cool. Then let's start with this one, then retractions (since we might want to check that this still works with the new dispatch)  – and then the error PR.

PR #171, opened 2023-10-30
Fix checks on TpM

kellertuer:
I noticed that `is_point` and `is_vector` were a bit too vague (accepting too much as correct points/vectors).

This PR fixes that by passing points/vectors to the vector check.


mateuszbaran (approved):
LGTM :+1:

PR #175, opened 2023-11-25
Fix retarction with t on power manifold

kellertuer:
In our memory-allocate-first approach now for retractions and that one can use retractions directly on Power manifold (and they are defined here) we missed to pass on `retract(M, p, X, t, m)` (and its corresponding `retract!`. This resolves https://github.com/JuliaManifolds/Manopt.jl/issues/330.

I still have to check test coverage, besides that this is good to go.


mateuszbaran (approved):
LGTM :+1:

PR #178, opened 2023-12-13
Introduce embedded vector transport

kellertuer:
as proposed in https://github.com/JuliaManifolds/Manifolds.jl/issues/624
this can be done quite generic.
This PR also changes the vector transport dispatch path a bit so that it is the same way as retractions and inverse retractions (allocate earlier) which does not require any test change, so it should be nonbreaking.

This PR still needs a bit of testing, but doc strings should be complete already.

### DIFF
--- a/src/vector_transport.jl
+++ b/src/vector_transport.jl
@@ -663,13 +663,13 @@ function _vector_transport_along!(
 end
 
 @doc raw"""
-    vector_transport_along_embedded!(M, Y, p, X, c, m::AbstractVectorTransportMethod; kwargs...)
+    vector_transport_along_embedded!(M::AbstractManifold, Y, p, X, c, m::AbstractVectorTransportMethod; kwargs...)
 
 Compute the vector transport of `X` from ``T_p\mathcal M`` along the curve `c`
 using the vector transport method `m` in the embedding and projecting the result back on
 the corresponding tangent space. The result is computed in place of `Y`.
 """
-vector_transport_along_embedded!(M, Y, p, X, c, ::AbstractVectorTransportMethod)
+vector_transport_along_embedded!(M::AbstractManifold, Y, p, X, c, ::AbstractVectorTransportMethod)
 
 function vector_transport_along_embedded! end
 
@@ -1034,7 +1034,7 @@ function _vector_transport_direction!(
 end
 
 @doc raw"""
-    vector_transport_direction_embedded!(M::AbstractManifold, Y, p, X, d, m::AbstractRetractionMethod)
+    vector_transport_direction_embedded!(M::AbstractManifold, Y, p, X, d, m::AbstractVectorTransportMethod)
 
 Compute the vector transport of `X` from ``T_p\mathcal M`` into the direction `d`
 using the [`AbstractRetractionMethod`](@ref) `m` in the embedding.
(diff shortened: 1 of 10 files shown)


kellertuer:
Decided to stay slightly longer than I try to usually this week to finish testing here as well. I think it works quite fine – and vector transports are closer to retractions now also in allocations – before they were somewhere in between, whether the default would allocate and call the in-place one or not – now it allocates as early as retractions as well. And none of the tests broke, and we have less code \o/

kellertuer:
Hm we seem to loose 9 lines of coverage somewhere which seems to be totally unrelated – will have to check that. But feature and test wise this is ready to review.

src/vector_transport.jl
```diff
@@ -1075,6 +1033,24 @@ function _vector_transport_direction!(
...
+@doc raw"""
+    vector_transport_direction_embedded!(M::AbstractManifold, Y, p, X, d, m::AbstractRetractionMethod)
+
+Compute the vector transport of `X` from ``T_p\mathcal M`` into the direction `d`
+using the  of the [`AbstractRetractionMethod`](@ref) `m` in th embedding.
+
+The default implementataion requires one allocation for the points and tangent vectors in the
+embedding and the resulting point, but the final projection is performed in place of `Y`
+"""
+function vector_transport_direction_embedded!(M, Y, p, X, d, m)
```
mateuszbaran, line 1045:
```suggestion
function vector_transport_direction_embedded!(M::AbstractManifold, Y, p, X, d, m::AbstractVectorTransportMethod)
```

src/vector_transport.jl
```diff
@@ -697,6 +650,28 @@ function _vector_transport_along!(
...
+    X,
+    c::AbstractVector,
+    m::EmbeddedVectorTransport;
+    kwargs...,
+)
+    return vector_transport_along_embedded!(M, Y, p, X, c, m.vector_transport; kwargs...)
+end
+
+@doc raw"""
+    vector_transport_along_embedded!(M, Y, p, X, c, m::AbstractVectorTransportMethod; kwargs...)
```
mateuszbaran, line 666:
```suggestion
    vector_transport_along_embedded!(M::AbstractManifold, Y, p, X, c, m::AbstractVectorTransportMethod; kwargs...)
```

src/vector_transport.jl
```diff
@@ -697,6 +650,28 @@ function _vector_transport_along!(
...
+
+Compute the vector transport of `X` from ``T_p\mathcal M`` along the curve `c`
+using the vector transport method `m` in the embedding and projecting the result back on
+the corresponding tangent space. The result is computed in place of `Y`.
+"""
+vector_transport_along_embedded!(M, Y, p, X, c, ::AbstractVectorTransportMethod)
```
mateuszbaran, line 672:
```suggestion
vector_transport_along_embedded!(M::AbstractManifold, Y, p, X, c, ::AbstractVectorTransportMethod)
```

src/vector_transport.jl
```diff
@@ -1075,6 +1033,24 @@ function _vector_transport_direction!(
     )
 end
 
+@doc raw"""
+    vector_transport_direction_embedded!(M::AbstractManifold, Y, p, X, d, m::AbstractRetractionMethod)
```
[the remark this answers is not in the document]
kellertuer: Ups, very well spotted!

mateuszbaran (approved):
Did you check if all tests in Manifolds.jl pass? If yes, that this LGTM.

kellertuer:
Ah that is maybe a good idea to do ;)

kellertuer:
It seems to “nearly work” just that the allocating manifolds like the circle need a bit of a tweak. But I can fix that on my next PR in Manifolds as well then.

kellertuer:
The fix is tone in the `atol=` PR, so this should be good to go. I will wait for the other PR though, it seems fine to put them together in one new version.

PR #184, opened 2024-03-12
Sectional curvature

mateuszbaran:
Adapted from Manopt.jl, with added functions for lower and upper bounds.

### DIFF
--- a/src/ManifoldsBase.jl
+++ b/src/ManifoldsBase.jl
@@ -151,12 +151,12 @@ function are_linearly_independent(
     p,
     X,
     Y;
-    eps_abs::Real = sqrt(eps(number_eltype(X))),
+    atol::Real = sqrt(eps(number_eltype(X))),
 )
     norm_X = norm(M, p, X)
     norm_Y = norm(M, p, Y)
     innerXY = inner(M, p, X, Y)
-    return norm_X > eps_abs && norm_Y > eps_abs && !isapprox(abs(innerXY), norm_X * norm_Y)
+    return norm_X > atol && norm_Y > atol && !isapprox(abs(innerXY), norm_X * norm_Y)
 end
 
 """
--- a/src/ManifoldsBase.jl
+++ b/src/ManifoldsBase.jl
@@ -153,8 +153,7 @@ function are_linearly_independent(M::AbstractManifold, p, X, Y)
     innerXY = inner(M, p, X, Y)
     return norm_X > eps_abs &&
            norm_Y > eps_abs &&
-           !isapprox(innerXY, norm_X * norm_Y) &&
-           !isapprox(innerXY, -norm_X * norm_Y)
+           !isapprox(abs(innerXY), norm_X * norm_Y)
 end
 
 """
--- a/src/ManifoldsBase.jl
+++ b/src/ManifoldsBase.jl
@@ -922,6 +922,52 @@ function riemann_tensor(M::AbstractManifold, p, X, Y, Z)
     return riemann_tensor!(M, Xresult, p, X, Y, Z)
 end
 
+@doc raw"""
+    sectional_curvature(M::AbstractManifold, p, X, Y)
+
+Compute the sectional curvature of a manifold ``\mathcal M`` at a point ``p \in \mathcal M``
+on two linearly independent tangent vectors at ``p``.
+The formula reads
+```math
+
+    \kappa_p(X, Y) = \frac{⟨R(X, Y, Y), X⟩_p}{\lVert X \rVert^2_p \lVert Y \rVert^2_p - ⟨X, Y⟩^2_p}
+
+```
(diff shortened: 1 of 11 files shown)


kellertuer:
This looks very nice! Thanks for adapting that.

If the test coverage is fine, I am ok with this.

mateuszbaran:
Question: should `sectional_curvature` return 0 for non-linearly independent vectors `X`, `Y` or be undefined? Making it 0 makes it much easier to define `sectional_curvature` on power and product manifolds but makes it harder to properly define the generic fallback that uses `riemann_tensor`.

kellertuer:
Why is that harder for the Riemannian tensor?
I would have thought it is either undefined or zero.

mateuszbaran:
The formula `inner(M, p, R, X) / ((norm(M, p, X)^2 * norm(M, p, Y)^2) - (inner(M, p, X, Y)^2))` is `0 / 0` when `X` and `Y` are linearly dependent. We can either detect this situation and return 0 (which requires adding a cut-off due to numerical inaccuracy) or say "don't use linearly dependent vectors".

kellertuer:
I think the clearer way would be “we do not use linearly dependent vectors”?

mateuszbaran:
That's how it currently works.

kellertuer:
Ah, yes. I see. And it is also well-documented that way. I think it is ok to keep it that way.

mateuszbaran:
Or we can just leave that not implemented for now.

mateuszbaran:
It doesn't work for product manifolds though. You may have two vectors that are linearly independent, but their projections on one or more of the component manifolds are not linearly independent, so that needs to be handled somewhere since sectional curvature for them is well-defined.

mateuszbaran:
Yes, that's what would be needed.

src/ManifoldsBase.jl
```diff
@@ -140,6 +140,23 @@ Compute the angle between tangent vectors `X` and `Y` at point `p` from the
...
     return acos(real(inner(M, p, X, Y)) / norm(M, p, X) / norm(M, p, Y))
 end
+
+"""
+    are_linearly_independent(M::AbstractManifold, p, X, Y)
+
+Check is vectors `X`, `Y` tangent at `p` to `M` are linearly independent.
+"""
+function are_linearly_independent(M::AbstractManifold, p, X, Y)
+    eps_abs = sqrt(eps(number_eltype(X)))
```
kellertuer, line 150: would it maybe make sense to make this a `tolerance` keyword argument?

besides that this function should stay internal (not exported) since it is only an internal helper.

src/ManifoldsBase.jl
```diff
@@ -140,6 +140,23 @@ Compute the angle between tangent vectors `X` and `Y` at point `p` from the
...
+    norm_X = norm(M, p, X)
+    norm_Y = norm(M, p, Y)
+    innerXY = inner(M, p, X, Y)
+    return norm_X > eps_abs &&
+           norm_Y > eps_abs &&
+           !isapprox(innerXY, norm_X * norm_Y) &&
+           !isapprox(innerXY, -norm_X * norm_Y)
```
kellertuer, line 157:
```suggestion
           !isapprox(abs(innerXY), norm_X * norm_Y)
```

would this not be a bit easier to read?

kellertuer (commented):
Overall this looks fine to me. Just two small remarks.

src/ManifoldsBase.jl
mateuszbaran: That's a good idea. I've kept the name to indicate that it's absolute tolerance (we might also want to specify relative tolerance for the `isapprox` test).

src/ManifoldsBase.jl
```diff
@@ -140,6 +140,25 @@ Compute the angle between tangent vectors `X` and `Y` at point `p` from the
...
+    are_linearly_independent(M::AbstractManifold, p, X, Y)
+
+Check is vectors `X`, `Y` tangent at `p` to `M` are linearly independent.
+"""
+function are_linearly_independent(
+    M::AbstractManifold,
+    p,
+    X,
+    Y;
+    eps_abs::Real = sqrt(eps(number_eltype(X))),
```
kellertuer, line 154: would maybe `atol` work as well?

mateuszbaran: Sure, done.

kellertuer:
This looks good so far. I am just missing the good old code coverage – did we miss to update the GitHub action? I remember something like this in a recent PR either on Manopt or Manifolds as well that I had to specifically add the secret (again) to the action.

kellertuer:
ah cf. https://github.com/JuliaManifolds/Manopt.jl/pull/357/commits/5d7646fbd0650da15d8d4bbe997fde29b4763af7

kellertuer (approved):

PR #197, opened 2024-08-02
Improved performance of power manifold creation and more

mateuszbaran:
The ManifoldsBase.jl improvements that are needed for better performance of IPM.

Note: constraints are creating new PowerManifolds at every access and these constructors were type-unstable before these changes, creating significant slowdowns.

`get_component(M::ProductManifold, p, i)` with integers is still slow and I didn't really want to sprinkle everything with `Base.@constprop :aggressive ` to try to improve it.


kellertuer (approved):
I trust that you know that this is faster ;)

PR #201, opened 2024-08-27
Add helpful message when `ProductManifold` is used without RAT.jl

mateuszbaran:
I had to change `representation_size` for `ProductManifold` for this but I think it doesn't make that much sense anyway.


kellertuer:
One could consider this breaking since it introduces an error in some cases; but I think instead of nothing being more careful with that is more like improve the user experience, so that is fine with me even in a nonbreaking sense.

kellertuer (approved):

mateuszbaran:
Yes, it's technically breaking but it's a part that no one should rely on.

PR #202, opened 2024-08-29
Refactored error message when no RAT.jl

mateuszbaran:
A small improvement to #201 to make it easier to handle another similar issue in Manifolds.jl, regarding `identity_element`.


kellertuer (approved):
Well-spotted.

Issue #204, opened 2024-09-27
get_vector failing for product manifold


Gertian:
Hi,

Firstly thanks to the devs for this wonderful package ! However, when using it today I ran into some issues which, as far as I understand, comes from the get_vector function failing in the context of product manifolds. However the error message is very vague and not really helpful for an uneducated end-user such as myself. 

The full code in question is below. Obviously I have tested if ``calc_energy(p, Γ_ref)`` works well.  My apologies in advance if this is just me being stupid and or if this should have been posted as an issue in one of the other packages such as `manifolds.jl`.

```julia
using Revise
#optimize on a manifold
using Manopt
using Manifolds
using RecursiveArrayTools
#finite differnces
using FiniteDifferences
using ManifoldDiff
#LinearAlgebra packages
using LinearAlgebra
using SkewLinearAlgebra #for efficient pfaffian !


#helper function that takes Q ∈ SO(2n) to a covariance matrix
function Q_to_Γ(Q)
    Σ = zero(Q)
    for i in 0:n-1
        Σ[2*i+1:2*i+2, 2*i+1:2*i+2] = [0 1;-1 0]
    end
    return Q*Σ*transpose(Q)
end
#helper function that takes a matrix and returs it skew-symmetric part.
function skewsymmetric(A) #TODO include some checks
    return 0.5*(A - transpose(A))
end

#set some paramters
L = 3
ph = 1
n = L*ph 
r = 3 
#make a generic coveriance matrix
Γ_ref = Q_to_Γ(rand(Rotations(2*n)))

#define the manifold
M = ProductManifold(Euclidean(r, field = ℂ ), PowerManifold(Rotations(2*n), NestedPowerRepresentation(), r))



#Helper function that takes in a set of prefactors, covariance matrices (in the form of a point ∈ manifold) and reference covariance matrix and spits out the energy : 
t=1. 
function calc_energy(p, Γ_ref; λs = p.x[1] , Γs = map(Qi -> Q_to_Γ(Qi), p.x[2]))
    #first we need to calculate all overlaps with the reference state. Due to gauge freedom these are all real and positive. Hence we can safely take sqrt(det) !
    gas = map(Γa -> 2^(-n/2) * det(Γa + Γ_ref)^(1/4), Γs)
    #now, we can make the matrics Δab cfr Brevyi : 
[...]

PR #205, opened 2024-09-30
Basis number system swap

mateuszbaran:
The solution of #204 we discussed today. The code seems to work locally but maybe documentation needs to be updated a bit more.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [0.15.17] unreleased
+## [0.15.17] 04/10/2024
 
 ### Changed
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.15.17] unreleased
+
+### Changed
+
+* **Mildly breaking**: the number system parameter now corresponds to the coefficients standing in front of basis vectors in a linear combination instead of components of a vector. For example, `DefaultOrthonormalBasis() == DefaultOrthonormalBasis(ℝ)` of `DefaultManifold(3, field=ℂ)` now has 6 vectors, and `DefaultOrthonormalBasis(ℂ)` of the same manifold has 3 basis vectors.
+
 ## [0.15.16] 13/09/2024
 
 ### Changed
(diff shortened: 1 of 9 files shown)


src/bases.jl
```diff
@@ -48,7 +48,7 @@ Abstract type that represents a basis of vector space of type `VST` on a manifol
 a subset of it.
 
 The type parameter `𝔽` denotes the [`AbstractNumbers`](@ref) that will be used
-for the vectors elements.
+as coefficients in linear combinations of vectors.
```
kellertuer, line 51:
```suggestion
as coefficients in linear combinations of the basis vectors.
```
maybe a bit more precise this way? (he other places accordingly of course)

test/default_manifold.jl
```diff
@@ -636,11 +636,12 @@ using ManifoldsBaseTestUtils
         MC = ManifoldsBase.DefaultManifold(3; field = ManifoldsBase.ℂ)
         p = [1.0im, 2.0im, -1.0im]
         CB = get_basis(MC, p, DefaultOrthonormalBasis(ManifoldsBase.ℂ))
+        @test CB.data == [[1.0, 0.0, 0.0], [0.0, 1.0, 0.0], [0.0, 0.0, 1.0]]
         @test CB.data isa Vector{Vector{ComplexF64}}
```
kellertuer, line 640: Is this test then still correct? For complex coefficients upfront, the eltype of the basis should be real (the `==` test above is even vs real basis vectors)

kellertuer:
In general I can check how well this is described in the docs page about bases and improve that somewhen tomorrow.

test/default_manifold.jl
mateuszbaran: At the moment `get_basis` returns vectors with complex eltype. It makes some sense from the point of view of type stability when operations are performed on those vectors but since all imaginary parts are 0, they could be real vectors.

At the moment it works like this because `p` has complex eltype and the code for producing basis vectors takes the eltype from `p`. Would you prefer to change it?

mateuszbaran:
> You were right today, a few things seem to indeed get nicer. Can you also check how much actually breaks with Manifolds.jl downstream?

Sure, I will make another PR to Manifolds.jl with adjustments.

> In general I can check how well this is described in the docs page about bases and improve that somewhen tomorrow.

Thanks!

kellertuer:
Besides the type of real coefficients being complex (see point above) this in general looks good to me.

test/default_manifold.jl
mateuszbaran: I've found a reasonable way to address this issue I think.

mateuszbaran:
OK, I've addressed that issue.

NEWS.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.15.17] unreleased
```
kellertuer, line 8: Maybe enter a date here then.

kellertuer (approved):

test/default_manifold.jl
kellertuer: Oh, a very fancy fix, nice.

NEWS.md
mateuszbaran: Sure, done

PR #206, opened 2024-10-18
Introduce r-norms on manifolds with components

kellertuer:
For manifolds that consist of components, like product, power and default manifold, this introduces an additional parameter to distance to specify the outer norm to be used after computing the single components distances; this especially also covers the 1- and ∞-norm cases. 
It also introduces a `has_components` function to be able to check whether a manifold can compute such an (outer) norm.

# 🛣️🗺️
* [x] `ProductManifold`
* [x] `PowerManifold`
* [x] `DefaultManifold`
* [x] adapt the idea also to `norm`
* [x]  Documentation 
* [x] News and version number 
* [x] Testing / Code coverage


src/DefaultManifold.jl
```diff
@@ -57,7 +57,7 @@ function check_approx(M::DefaultManifold, p, X, Y; kwargs...)
     return ApproximatelyError(v, s)
 end
 
-distance(::DefaultManifold, p, q) = norm(p - q)
+distance(::DefaultManifold, p, q, r::Real = 2.0) = norm(p - q, r)
```
mateuszbaran, line 60:
```suggestion
distance(::DefaultManifold, p, q, r::Real = 2) = norm(p - q, r)
```
LinearAlgebra.jl has an integer default here so maybe that's a good idea.

src/DefaultManifold.jl
```diff
@@ -138,7 +140,7 @@ end
 
 number_system(::DefaultManifold{𝔽}) where {𝔽} = 𝔽
 
-norm(::DefaultManifold, p, X) = norm(X)
+norm(::DefaultManifold, p, X, r = 2.0) = norm(X, r)
```
mateuszbaran, line 143:
```suggestion
norm(::DefaultManifold, p, X, r::Real = 2) = norm(X, r)
```

src/PowerManifold.jl
```diff
@@ -530,20 +530,103 @@ function default_vector_transport_method(M::PowerManifold, t::Type)
...
+    isinf(r) && return _distance_Inf(M, p, q)
+    (r == 1) && return _distance_1(M, p, q)
+    return _distance_r(M, p, q, r)
+end
+function distance(
+    M::AbstractPowerManifold,
+    p,
+    q,
+    ::LogarithmicInverseRetraction,
+    r::Real = 2.0,
```
mateuszbaran, line 559:
```suggestion
    r::Real = 2,
```

src/PowerManifold.jl
```diff
@@ -530,20 +530,103 @@ function default_vector_transport_method(M::PowerManifold, t::Type)
...
+    return distance(M, p, q, r)
+end
+
+@doc "$(_doc_distance_pow)"
+function distance(
+    M::AbstractPowerManifold,
+    p,
+    q,
+    m::AbstractInverseRetractionMethod,
+    r::Real = 2.0,
```
mateuszbaran, line 570:
```suggestion
    r::Real = 2,
```

src/PowerManifold.jl
```diff
@@ -1092,23 +1182,46 @@ function mid_point!(M::PowerManifoldNestedReplacing, q, p1, p2)
...
 Compute the norm of `X` from the tangent space of `p` on an
-[`AbstractPowerManifold`](@ref) `M`, i.e. from the element wise norms the
-Frobenius norm is computed.
+[`AbstractPowerManifold`](@ref) `M`, i.e. from the element wise norms `r`-norm is computed,
+where the default `r=2.0` yields the Frobenius norm is computed.
 """
-function LinearAlgebra.norm(M::AbstractPowerManifold, p, X)
-    sum_squares = zero(number_eltype(X))
+function LinearAlgebra.norm(M::AbstractPowerManifold, p, X, r::Real = 2.0)
+    isinf(r) && return _norm_Inf(M, p, X)
```
mateuszbaran, line 1192: Technically it makes sense to distinguish positive infinity norm and negative infinity not-really-norm:
```julia
julia> norm([1, 2], Inf)
2.0

julia> norm([1, 2], -Inf)
1.0
```
So for `r` equal to `-Inf` we should either follow LinearAlgebra.jl or leave it undefined.

src/PowerManifold.jl
```diff
@@ -1092,23 +1182,46 @@ function mid_point!(M::PowerManifoldNestedReplacing, q, p1, p2)
...
+function _norm_1(M::AbstractPowerManifold, p, X)
+    s = zero(number_eltype(p))
+    rep_size = representation_size(M.manifold)
+    for i in get_iterator(M)
+        s += norm(M.manifold, _read(M, rep_size, p, i), _read(M, rep_size, X, i))
+    end
+    return s
+end
+function _norm_Inf(M::AbstractPowerManifold, p, X)
+    d = 0.0
```
mateuszbaran, line 1213:
```suggestion
    d = float(zero(eltype(X))
```

src/PowerManifold.jl
```diff
@@ -1092,23 +1182,46 @@ function mid_point!(M::PowerManifoldNestedReplacing, q, p1, p2)
...
+    v = 0.0
```
mateuszbaran, line 1214:
```suggestion
```
We don't need that for anything.

src/ProductManifold.jl
```diff
@@ -372,22 +372,49 @@ function default_vector_transport_method(M::ProductManifold)
...
+
+Compute the distance between `q` and `p` on an [`ProductManifold`](@ref).
+
+First, the componentwise distances are computed. These can be approximated using the
+`norm` of an [`AbstractInverseRetractionMethod`](@ref) `m`.
+Then, the `r`-norm of these elements is computed.
+"""
+
+@doc "$(_doc_distance_prod)"
+function distance(M::ProductManifold, p, q, r::Real = 2.0)
```
mateuszbaran, line 387: Shouldn't this use the r-distance on component-like manifolds? In that case I'd actually prefer to have two separate implementations for default `r` and custom `r` instead of a default value here.

src/ProductManifold.jl
```diff
@@ -372,22 +372,49 @@ function default_vector_transport_method(M::ProductManifold)
...
+        map(
+            distance,
+            M.manifolds,
+            submanifold_components(M, p),
+            submanifold_components(M, q),
+        ),
+        r,
+    )
+end
+function distance(M::ProductManifold, p, q, ::LogarithmicInverseRetraction, r::Real = 2.0)
```
mateuszbaran, line 398:
```suggestion
function distance(M::ProductManifold, p, q, ::LogarithmicInverseRetraction, r::Real = 2)
```

src/ProductManifold.jl
```diff
@@ -372,22 +372,49 @@ function default_vector_transport_method(M::ProductManifold)
...
+    return distance(M, p, q, r)
+end
+
+@doc "$(_doc_distance_prod)"
+function distance(
+    M::ProductManifold,
+    p,
+    q,
+    m::AbstractInverseRetractionMethod,
+    r::Real = 2.0,
```
mateuszbaran, line 408:
```suggestion
    r::Real = 2,
```

src/ProductManifold.jl
```diff
@@ -725,21 +759,15 @@ function mid_point!(M::ProductManifold, q, p1, p2)
...
-    norms_squared = (
-        map(
-            norm,
-            M.manifolds,
-            submanifold_components(M, p),
-            submanifold_components(M, X),
-        ) .^ 2
-    )
-    return sqrt(sum(norms_squared))
+function LinearAlgebra.norm(M::ProductManifold, p, X, r::Real = 2.0)
```
mateuszbaran, line 767:
```suggestion
function LinearAlgebra.norm(M::ProductManifold, p, X, r::Real = 2)
```

mateuszbaran (commented):
A few comments.

src/DefaultManifold.jl
kellertuer: Sure.

src/PowerManifold.jl
kellertuer: Iff, this already now feels like boilerplates of boilerplates, with another case to manage – well, sure. just another 20+ lines down to hack.
And I have never seen that one in action, but sure, that can be done

src/ProductManifold.jl
kellertuer: I do not understand the question nor what you would like to have here, can you elaborate?

kellertuer: This _does_ use the component-r-norm thingy.

kellertuer:
Ufftoday is not my day for coding, I feel. Took me an hour to get the tests to run, now JuliaFormatter errors locally on me on code that previously ran just fine. There is days were I think I should maybe switch to woodworking or so.

src/PowerManifold.jl
kellertuer: added it, though I am not sure this actually is a norm – but maybe there computer scientists (that can be done in code) and mathematicians (this is nor a norm/distance) disagree ;)

src/ProductManifold.jl
mateuszbaran: Regarding the first part, let's take `M` to be SE(3), and `p = ArrayPartition([0, 0, 0], I(3))`, `q = ArrayPartition([3, 4, 0], zeros(3, 3))`. Should `distance(M, p, q, Inf)` return 4 or 5? The current code would return 5 but I think 4 would probably be more appropriate?

kellertuer: Ah, I se what you mean. It should return 5 and that is correct I think.

It does the usual distance on every component. That is the 2-norm on the R3-part and is 5, the second component is 0.
Only _then_, from these components, the Inf norm is taken.

The Inf can not be passed “recursively”, i.e. Inf does not mean “Do Inf norm on all manifolds where this makes sense” – that to me would be magic; and the `r=` parameter does not exist on all manifolds, so even in code getting 4 generically would be a super-hustle.

TLDR: So with the idea “classical distance on the components, r-norm on the vector” – 5 is correct.

src/PowerManifold.jl
mateuszbaran: The `-Inf` function is I think a seminorm but is definitely not a norm but in general. Well, some people seem to find it useful.

src/ProductManifold.jl
kellertuer: And it would be a super huste to test as well. I am already trying to fix/test the corner cases on the existing code and that takes me 2 hours already now. So If you want to have pass-down-inf-magic that would take much longer.

src/PowerManifold.jl
kellertuer: Sure, added it already – just fighting a few remaining inverse-retractions-and-product-manifold stuff now.

src/PowerManifold.jl
```diff
@@ -530,22 +530,122 @@ function default_vector_transport_method(M::PowerManifold, t::Type)
...
+    distance(M::AbstractPowerManifold, p, q, r::Real=2.0)
+    distance(M::AbstractPowerManifold, p, q, m::AbstractInverseRetractionMethod=LogarithmicInverseRetraction(), r::Real=2.0)
+
+Compute the distance between `q` and `p` on an [`AbstractPowerManifold`](@ref).
 
-Compute the distance between `q` and `p` on an [`AbstractPowerManifold`](@ref),
-i.e. from the element wise distances the Forbenius norm is computed.
+First, the componentwise distances are computed. These can be approximated using the
+`norm` of an [`AbstractInverseRetractionMethod`](@ref) `m`.
+Then, the `r`-norm of these elements is computed.
```
mateuszbaran, line 541:
```suggestion
Then, the `r`-norm of these elements is computed.

Note that the `r` argument is *not* passed recursively to components when the distance is computed.
```
Maybe that would make it clear?

src/PowerManifold.jl
```diff
@@ -530,22 +530,122 @@ function default_vector_transport_method(M::PowerManifold, t::Type)
     return default_vector_transport_method(M.manifold, eltype(t))
 end
 
-@doc raw"""
-    distance(M::AbstractPowerManifold, p, q)
+_doc_distance_pow = """
+    distance(M::AbstractPowerManifold, p, q, r::Real=2.0)
+    distance(M::AbstractPowerManifold, p, q, m::AbstractInverseRetractionMethod=LogarithmicInverseRetraction(), r::Real=2.0)
```
mateuszbaran, line 535:
```suggestion
    distance(M::AbstractPowerManifold, p, q, r::Real=2)
    distance(M::AbstractPowerManifold, p, q, m::AbstractInverseRetractionMethod=LogarithmicInverseRetraction(), r::Real=2)
```

src/PowerManifold.jl
```diff
@@ -530,22 +530,122 @@ function default_vector_transport_method(M::PowerManifold, t::Type)
...
-Compute the distance between `q` and `p` on an [`AbstractPowerManifold`](@ref),
-i.e. from the element wise distances the Forbenius norm is computed.
+First, the componentwise distances are computed. These can be approximated using the
+`norm` of an [`AbstractInverseRetractionMethod`](@ref) `m`.
+Then, the `r`-norm of these elements is computed.
 """
+
 function distance(M::AbstractPowerManifold, p, q)
-    sum_squares = zero(number_eltype(p))
+    return _distance_r(M, p, q, 2.0)
```
mateuszbaran, line 545:
```suggestion
    return _distance_r(M, p, q, 2)
```

src/PowerManifold.jl
```diff
@@ -530,22 +530,122 @@ function default_vector_transport_method(M::PowerManifold, t::Type)
...
+)
+    (isinf(r) && r > 0) && return _distance_max(M, p, q, m)
+    (isinf(r) && r < 0) && return _distance_min(M, p, q, m)
+    (r == 1) && return _distance_1(M, p, q, m)
+    return _distance_r(M, p, q, m, r)
+end
+#
+#
+# The three special cases
+function _distance_r(M::AbstractPowerManifold, p, q, m::AbstractInverseRetractionMethod, r)
```
mateuszbaran, line 581:
```suggestion
function _distance_r(M::AbstractPowerManifold, p, q, m::AbstractInverseRetractionMethod, r::Real)
```

src/PowerManifold.jl
```diff
@@ -530,22 +530,122 @@ function default_vector_transport_method(M::PowerManifold, t::Type)
...
+    rep_size = representation_size(M.manifold)
+    values = [
+        distance(M.manifold, _read(M, rep_size, p, i), _read(M, rep_size, q, i), m) for
+        i in get_iterator(M)
+    ]
+    return norm(values, r)
+end
+function _distance_r(M::AbstractPowerManifold, p, q, r)
```
mateuszbaran, line 589:
```suggestion
function _distance_r(M::AbstractPowerManifold, p, q, r::Real)
```

src/PowerManifold.jl
kellertuer: I am not so sure. In 98% of times “passing down” does not even make sense (that is outside of Product/Power of Power/Product – on Manifolds.jl maybe Euclidan).
I was of the impression that “of these elements” should be precise enough.

src/PowerManifold.jl
```diff
@@ -530,22 +530,122 @@ function default_vector_transport_method(M::PowerManifold, t::Type)
...
+function _distance_1(M::AbstractPowerManifold, p, q)
+    s = zero(number_eltype(p))
+    rep_size = representation_size(M.manifold)
+    for i in get_iterator(M)
+        s += distance(M.manifold, _read(M, rep_size, p, i), _read(M, rep_size, q, i))
+    end
+    return s
+end
+function _distance_max(M::AbstractPowerManifold, p, q, m::AbstractInverseRetractionMethod)
+    d = 0.0
```
mateuszbaran, line 614:
```suggestion
    d = float(zero(number_eltype(p))
```

src/PowerManifold.jl
kellertuer: Already in the current commit I am working through :)

src/PowerManifold.jl
```diff
@@ -530,22 +530,122 @@ function default_vector_transport_method(M::PowerManifold, t::Type)
...
+    rep_size = representation_size(M.manifold)
+    for i in get_iterator(M)
+        v = distance(M.manifold, _read(M, rep_size, p, i), _read(M, rep_size, q, i), m)
+        d = (v > d) ? v : d
+    end
+    return d
+end
+function _distance_max(M::AbstractPowerManifold, p, q)
+    d = 0.0
```
mateuszbaran, line 623:
```suggestion
    d = float(zero(number_eltype(p))
```

src/PowerManifold.jl
```diff
@@ -1092,20 +1199,51 @@ function mid_point!(M::PowerManifoldNestedReplacing, q, p1, p2)
 end
 
 @doc raw"""
-    norm(M::AbstractPowerManifold, p, X)
+    norm(M::AbstractPowerManifold, p, X, r::REal=2.0)
```
mateuszbaran, line 1202:
```suggestion
    norm(M::AbstractPowerManifold, p, X, r::Real=2)
```

src/PowerManifold.jl
```diff
@@ -1092,20 +1199,51 @@ function mid_point!(M::PowerManifoldNestedReplacing, q, p1, p2)
...
 
 Compute the norm of `X` from the tangent space of `p` on an
-[`AbstractPowerManifold`](@ref) `M`, i.e. from the element wise norms the
-Frobenius norm is computed.
+[`AbstractPowerManifold`](@ref) `M`, i.e. from the element wise norms `r`-norm is computed,
+where the default `r=2.0` yields the Frobenius norm is computed.
 """
-function LinearAlgebra.norm(M::AbstractPowerManifold, p, X)
-    sum_squares = zero(number_eltype(X))
+function LinearAlgebra.norm(M::AbstractPowerManifold, p, X, r::Real = 2.0)
```
mateuszbaran, line 1208:
```suggestion
function LinearAlgebra.norm(M::AbstractPowerManifold, p, X, r::Real = 2)
```

src/PowerManifold.jl
```diff
@@ -1092,20 +1199,51 @@ function mid_point!(M::PowerManifoldNestedReplacing, q, p1, p2)
...
+    (isinf(r) && r > 0) && return _norm_max(M, p, X)
+    (isinf(r) && r < 0) && return _norm_min(M, p, X)
+    (r == 1) && return _norm_1(M, p, X)
+    return _norm_r(M, p, X, r)
+end
+function _norm_r(M::AbstractPowerManifold, p, X, r)
```
mateuszbaran, line 1214:
```suggestion
function _norm_r(M::AbstractPowerManifold, p, X, r::Real)
```

src/PowerManifold.jl
mateuszbaran: To me passing-down is a sufficiently plausible alternative that we should make this point very clear in the docstring. Not passing it is of course perfectly fine, it's just a matter of making the documentation clear.

kellertuer: Irgs, I was too fast an added this. Meh. My brain starts melting after about 3 hours of debugging corner cases I fear.

src/ProductManifold.jl
mateuszbaran: OK, 5 is fine, let's just make it clear in docs.

src/PowerManifold.jl
kellertuer: The passing down reminds me of the “autobroadcast” in Matlab, where when you accidentally add a row to a column vector, you get the corresponding rank1-1matrix.
Expecially when it is context dependent whether that can happen, this is dangerous magic. We should not do that. And even getting it right would be challenging, For every distance/norm call we would have to check whether we can pass down r? Uff.

mateuszbaran: Sorry, I just wanted to discuss this issue. Not passing it down is perfectly fine.

kellertuer: And about documenting it right, I had hoped That is documented originally already right.
We do _elementwise_ the _classical_ distance. on that we do the R-norm

I would also prefer to document, _what_ we do and not _what we not do_ ;) Because what we not do, we would add a lot of things for sure as well. The method also does not make you dinner ;)

kellertuer:
I rephrased the docs and made more clearer, that `r`-norm _only_ applies to the outer layer.

I do agree it would maybe be nice to “pass in” an `r` to inner layers – but that should be done _explicitly_, would involve more interface work and more checks – since we would have to check whether the “inner” layer (single factors of the product manifold or the manifold in power) do accept the `r`. And what do we do if only 2 of 5 factors of the product accept an r-norm themselves? So for now I think, not handling “inner-r-norms” is fine – when someone needs them and has a good interface, we can add it.
My fear is, that already now we have a code-explosion in list/norm – with the just described features, that would yield quite some chaos.

kellertuer:
Already now I struggle to find test cases for all, I think I am still missing a single-inverse-retraction which is not Log on ProductManifold as a test case. But I think I have to implement an own case to check for that, a new inverse retraction a a test case or so.

kellertuer:
Thanks for already fixing all the types (`::Real`). and defaults (to `2`) – I hope I now found the last few tricky boundary cases, where functions are defined to avoid ambiguities.

Let's postpone the idea `distance(M, p, q, m, Inf; recursive=true)` to a later point. It could surely check whether any inner manifold `has_components` and for those cases pass the `r` down to them as well. But that would involve so many places I think (over 20?) that this would take me a day or two to find all the details. That is maybe not worth the effort for now, when I do not see who could find that useful.

mateuszbaran:
Thanks! Let's not do recursive `r` in this PR, or maybe at all.

kellertuer:
Can I merge this later today? I would like to maybe do the gradient change stopping criterion later today still, which was the origin for this feature.

test/power.jl
```diff
@@ -226,6 +232,10 @@ end
                     @test manifold_dimension(N) == prod(pow_size) * manifold_dimension(M)
                     @test mid_point(N, p, q) == mid_point.(Ref(M), p, q)
                     @test sqrt(inner(N, p, q, q)) ≈ norm(N, p, q)
+                    norms = norm.(Ref(M), p, q)
+                    #@test norm(N, p, q, 1.0) = sum(norms)
+                    #@test norm(N, p, q, 2.0) = sqrt(sum(norms.^2))
+                    #@test norm(N, p, q, Inf) = maximum(norms)
```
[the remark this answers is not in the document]
kellertuer: Oh, one of those lines reports “ LoadError: syntax: "1" is not a valid function argument name around ManifoldsBase.jl/test/power.jl:236” –  I do not understand why, but will investigate a bit.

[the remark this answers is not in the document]
kellertuer: Ah, I must have been a bit tired already, maybe to not test assignments ;) Re-added tests, will push after tests ran locally.

PR #213, opened 2024-11-15
Make ONB default

mateuszbaran:
I think this would be nice for convenience.

### DIFF
--- a/src/bases.jl
+++ b/src/bases.jl
@@ -842,6 +842,8 @@ end
     get_vectors(M::AbstractManifold, p, B::AbstractBasis=get_basis(M, p, DefaultOrthonormalBasis()))
 
 Get the basis vectors of basis `B` of the tangent space at point `p`.
+The function may or may not work if passed a basis other than a [`CachedBasis`](@ref).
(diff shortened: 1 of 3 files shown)


src/bases.jl
```diff
@@ -817,13 +839,16 @@ function get_vector_cached!(M::AbstractManifold, Y, p, X, B::CachedBasis)
 end
 
 """
-    get_vectors(M::AbstractManifold, p, B::AbstractBasis)
+    get_vectors(M::AbstractManifold, p, B::AbstractBasis=get_basis(M, p, DefaultOrthonormalBasis()))
```
kellertuer, line 842: Why do the others just need `DefaultOrthonormalBasis()` but this here the same wrapped in another call? And the code seems to do something different, namely not call `get_basis`?

mateuszbaran: `get_vectors` is generally supposed to work with cached bases, so we need to get a cached basis first.

kellertuer: Then we could maybe mention that in line 844 in the docs here?

mateuszbaran: Sure, I've added a note there.

PR #217, opened 2024-11-23
Update `ValidationManifold`

### LINKED ISSUE #216: Improve `ValidationManifold`
At the community meeting today we had an interesting discussion about the validation manifold

* maybe with its current name it is not as visible as expected / its use not fully clear
* Improce `ValidationTVector` to optionally store the base point of the tangent vector as well; the current constructor would store `Nothing`, the new one a point
* The `ValidationManifold` should store whether new tangent vectors on this manifold store their base point
* The `ValidationManifold` could store a dictionary, that allows to deactivate certain checks
  *  key of the dictionary could be the function where stuff is deactivated
  * the value could be a named tuple or dictionary that might store a `Symbol` (same as the current mode), where `:none` would even deactivate the test. Possible keys for the sub dict might be
    * `:Input`
    * `:point`
    *  `:vector`
    * `:output`
    * and maybe other like norm being positive or such
    * default is always the global `mode`   
* check that all current function always return typed values (unless a number is returned)
* check / update that all interface functions are covered.

(cc @Affie)

kellertuer:
Aims to resolve #216.

# 🛣️🗺️

* [x] introduce a `point` to the vector validation types
* [x] introduce s `store_base_point` to `ValidationManifold`
* [x] add a fine granular dictionary to disable certain validations
* [x] adapt all existing methods
* [x] adapt all `error` calls to now use `_msg` to also respect the `M.mode` in these
* [x] check documentation that the goal of the validation manifold becomes clearer
* [x] check that also newer interface functions are covered by the validation manifold
* [x] fix converts / ambiguities after generalising to accepting all points for validation


kellertuer:
@mateuszbaran can you maybe take a look at the current design of the idea of ignoring parts of the tests?
I just tested it on `distance`, since there we have a “non point, non vector output” that potentially still has a check.

But I already tried documenting it fully as well

If this design is ok I would go through the whole API and
* add functions to the validation manifold that are currently not yet covered
* add these deactivation mechanisms to all checks performed.

src/ValidationManifold.jl
```diff
@@ -127,31 +237,35 @@ function copyto!(
     return p
 end
 
+decorated_manifold(M::ValidationManifold) = M.manifold
+
 function distance(M::ValidationManifold, p, q; kwargs...)
-    is_point(M, p; error = M.mode, kwargs...)
-    is_point(M, q; error = M.mode, kwargs...)
-    return distance(M.manifold, array_value(p), array_value(q))
+    _vMc(distance, (:Point, :Input), M.ignore) && is_point(M, p; error = M.mode, kwargs...)
```
mateuszbaran, line 243: I'd suggest putting `is_point` check in `_vMc` or some other helper function so that less code is duplicated across functions.

src/ValidationManifold.jl
```diff
@@ -1,52 +1,149 @@
...
+    store_base_point::Bool
+    ignore::D
 end
-function ValidationManifold(M::AbstractManifold; error::Symbol = :error)
-    return ValidationManifold(M, error)
+function ValidationManifold(
+    M::AbstractManifold;
+    error::Symbol = :error,
+    store_base_point::Bool = false,
+    ignore::D = Dict{Union{Symbol,Function},Union{Bool,Vector{Symbol}}}(),
```
mateuszbaran, line 70: In which case can the value of `ignore` be a `Bool`? `_vMc` doesn't seem to handle that.

mateuszbaran (commented):
I have two comments for now.

src/ValidationManifold.jl
kellertuer: This case https://github.com/JuliaManifolds/ManifoldsBase.jl/blob/d5712ca9bbe5d6bf4e0e99c740fa0adc231802a9/src/ValidationManifold.jl#L93

The idea is: with `exp => [:Point]` you deactivate `is_points` in all exp calls – but if you want to deactivate _all_ point checks for all functions you do `:Point => false` (or since it is an ignore dict maybe also true instead, have to think about that; one could also ignore all points stuff if the key is present) cf https://github.com/JuliaManifolds/ManifoldsBase.jl/blob/d5712ca9bbe5d6bf4e0e99c740fa0adc231802a9/src/ValidationManifold.jl#L21

src/ValidationManifold.jl
mateuszbaran: The call in `distance` would be reduced to just
```julia
 _vMc_is_point(M, distance, :Input, p; kwargs...)
```
instead of
```julia
_vMc(distance, (:Point, :Input), M.ignore) && is_point(M, p; error = M.mode, kwargs...)
```
which looks much shorter.

kellertuer: yes, but I would implement quite a few such wrappers?

src/ValidationManifold.jl
mateuszbaran: Hm, I'd suggest storing these two things separately for more clarity. Maybe `ignore_f` and `ignore_all`? Then `ignore_all` could also just be a `Vector{Symbol}` instead of `Dict` (you'd just check `(type in ignore_all) && return false`).

src/ValidationManifold.jl
mateuszbaran: Only for points and vectors, so two? Other things wouldn't necessarily need such wrapper because they are more rare.

kellertuer: hm, will think about it, it is a bit less flexible then

src/ValidationManifold.jl
kellertuer: the first could be `ignore_function` and the other `ignore_type`, sure, will do that.

src/ValidationManifold.jl
mateuszbaran: `_vMc` could still remain as is but it would have a less flexible wrapper for easier usage in typical cases.

kellertuer: fair enough

kellertuer:
I found even a nicer way
* overload `is_point(M::ValidationManifold,...)` with two keywords – the context `:Point` is also added automatically there.
* pass `M` also to `_vMc`

the internals stay within `_vMc` that way

I renamed the vector-thing to `ignore_context` because for example with `:Input`  as an area/context to ignore, this sounded good.

Now one only has to put a `function=..., context=...` to a lot of the input/output checks.

kellertuer:
Just adapted all other point/vector checks to this new schemes, feels good in style I think.

1) I want to rework the `error` calls to use `_msg`  (which might turn these into warnings) and adapt them to the function/context modes as well.
2) I want to check which functions from the API are not yet covered in the validation and add them as well.

Besides these two points, the PR should be done, so I already set it to active.

kellertuer:
This is feature complete and hence ready for review.

* Since now all old “errors” that always threw errors now respect both mode and context, several tests have still to be adapted
* Since I also added a few functions from the interface, these also still need test coverage.

Will continue with this the next few days. But overall I like the new features and how they turned out

mateuszbaran:
Nice, I think the updated approach looks better.

src/ValidationManifold.jl
```diff
@@ -260,87 +527,220 @@ function injectivity_radius(M::ValidationManifold, method::AbstractRetractionMet
...
+* `context::NTuple{N,Symbol}=NTuple{0,Symbol}()` to specify one or more contexts, this
+  call was issued in. The context `:Point` is added before checking whether the test
+  should be performed
+
+all other keywords are passed on.
+"""
+function is_point(
+    M::ValidationManifold,
+    p;
+    within::Union{Nothing,<:Function} = nothing,
```
mateuszbaran, line 580: We could also  have `error = M.mode` kwarg here? It would simplify using this method. And then pass `error` to `is_point`.

src/ValidationManifold.jl
```diff
@@ -122,55 +268,136 @@ function copyto!(
...
+    y = embed(M.manifold, _value(p), _value(X))
+    is_point(M, y; error = M.mode, within = embed, context = (:Output,), kwargs...)
+    return ValidationMPoint(y)
+end
+function embed(M::ValidationManifold, p, X; kwargs...)
+    is_point(M, p; error = M.mode, within = embed, context = (:Input,), kwargs...)
+    is_vector(M, p, X; error = M.mode, within = embed, context = (:Input,), kwargs...)
+    q = embed(M.manifold, _value(p), _value(X))
+    is_point(M, q; error = M.mode, within = embed, context = (:Output,), kwargs...)
+    return ValidationMPoint(q)
```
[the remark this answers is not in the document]
kellertuer: Yes, sometimes, AI helps a bit and sometimes its off :D

src/ValidationManifold.jl
kellertuer: nice idea.

kellertuer: did the same for `_msg` and `is_vector`.

kellertuer:
The original tests are back up and running, just one `convert` from a manifold _to_ a variation manifold is broken and I can not see how to fix it – in other words: If that code worked before I do not see why it does now no longer work.

The remainder will just be to also cover the new functions I implemented.

mateuszbaran:
The main reason why the old test doesn't work is that that `==` in Julia is comparing by default in a non-intuitive way when a struct has mutable fields. See here: https://discourse.julialang.org/t/surprising-struct-equality-test/4890 . Previously there were no mutable fields in `ValidationManifold` and now there are.

kellertuer:
Now just
* embed embed_project
* riemannian_tensor
* and a few text message prints (which is rare now since most are Errors) are missing.

NEWS.md
```diff
@@ -5,6 +5,18 @@ All notable changes to this project will be documented in this file.
...
+## [0.15.23]
+
+### Added
+
+* a field `point` to `ValidationFibreVector` to store potentially store the point of the vector.
+* a field `store_base_point` to `ValidationManifold` to indicate whether for new fibre vectors the base point should be stored.
+
+### Changed
+
+* the internal function `_array_value` was renamed to `_value`,
```
mateuszbaran, line 17: Manifolds.jl currently uses and extends `array_value`. It would be easier to just keep the name, even if it isn't entirely appropriate.

kellertuer: Since we have an `_` I would consider that internal and since it also works for arbitrary points and not only for array values (now even more precise), I think it is worth renaming.

That it breaks Manifolds.jl is a bit of a problem, but we were maybe not careful enough splitting both packages there.
I can do the PR that fixes that on Manifolds, if you prefer.

mateuszbaran: Actually, the old name is `array_value` (without the leading underscore). And, as you can see, making it internal doesn't exactly work out since there is no way to extend `ValidationManifold` without using that function.

Renaming to `_value` IMO overly emphasizes the level of "being internal" this function has. There is not much reason to extend it and its direct usage is rarely the right choice, but it's not "do not touch, copy if you must" level of "internal".

kellertuer: Hm, right to extend that to other functions is indeed complicated from outside (new types/manifolds should work directly). But the name is super misleading, it does not provide the array value but the internally stored thingy – however another user typed that, maybe an array, maybe an `AbstractManifoldPoint´ or – whatever..

can we then maybe (for non-breakingness) keep but deprecate `array_value` and use a new name?
`internal_value` maybe?
The next breaking release of ManifoldsBase (or the second-to-next) could then remove it

mateuszbaran: OK, `internal_value` sounds fine, and `array_value` can be a deprecated alias.

kellertuer:
I hope this PR should now be finished.

kellertuer:
This PR'does not yet introduce interlinks, but the CSS for the prefixes is also useful here as well already, so I felt adding it wouldn't be too bad. But with that the PR is indeed finished.

src/ValidationManifold.jl
```diff
@@ -122,55 +295,143 @@ function copyto!(
...
+    MEV = ValidationManifold(get_embedding(M.manifold), M)
+    is_vector(
+        MEV,
+        embed(M.manifold, internal_value(p)),
+        Y;
+        within = embed,
+        context = (:Output,),
+        kwargs...,
+    )
+    return ValidationTVector(Y)
```
kellertuer, line 338: Oh, I have to check this carefully and add the base point (a copy of it) is active.

kellertuer:
I fixed the addition, I hope now all works fine as intended also with base points in TVectors.

mateuszbaran (approved):

PR #218, opened 2024-11-27
Small typos

lkdvos:
Hi,

I was reading through the manual and simultaneously fixed a couple typos.
Just as a small remark, one of the changes is not just a typo, but I slightly reworded the sentence because it was a bit confusing to me. If you prefer not to have this, let me know and I'll change it back


kellertuer:
Thanks, this looks great. And yes, we only slowly work through old code and use more spellcheckers, so any fix of typos is super welcome – thanks for reporting them!

lkdvos:
My pleasure, I have the exact same problem for my own packages and always really appreciate when someone just fixes typos as they go

PR #223, opened 2025-02-05
Fix exp_fused allocation

mateuszbaran:
This should fix the issue with ManifoldDiff.jl.


kellertuer:
Suddenly we also have the one line back I was missing yesterday – that's where it went :D

PR #225, opened 2025-03-12
Allow an `allocate(M, f, T)` case

kellertuer:
This was an interims step anyways when calling `allocate_result(M; f)` but I would like to have it as an extra step especially for functions like `zero_vector(G, T)` where we do not have a `x` variable to “allocate from” but also do not want to use the “default type” `allocate_result_type` would return.

The idea _should_ work, I just got a bit lost in fighting ambiguities for now

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "ManifoldsBase"
 uuid = "3362f125-f0bb-47a3-aa74-596ffd7ef2fb"
 authors = ["Seth Axen <seth.axen@gmail.com>", "Mateusz Baran <mateuszbaran89@gmail.com>", "Ronny Bergmann <manopt@ronnybergmann.net>", "Antoine Levitt <antoine.levitt@gmail.com>"]
-version = "1.0.1"
+version = "1.1.0"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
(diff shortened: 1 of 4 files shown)


kellertuer:
Hmpf. And it wraps once too much in Array. The use case is that I want to allocate a 0-dim Array and know its type, so

```
ManifoldsBase.allocate_result(Circle(), exp, Array{Float64,0})
```
should “materialize” that array. But as for so often in these technical details I seem to miss something. Can you take a look @mateuszbaran ?

kellertuer:
Would it then maybe make more sense to continue to overwrite `identity_element(G,T)` ? This was the detour I took until now but it requires to see all types `T` that might come along in the future.

mateuszbaran:
Extending `allocate` to a new signature is a bad idea. Would `allocate_on(G, TangentSpaceType(), X)` work for you? In that example `X` is the type of tangent vector to allocate.

> ah! Is `allocate_result_type` more like `allocate_result_eltype`? Then I am mainly struggling with a bit of missing docs :/

Yes, it's even documented: https://github.com/JuliaManifolds/ManifoldsBase.jl/blob/85b42907c26df0463f3fc91ba7dafd3fa534f800/src/ManifoldsBase.jl#L197-L200 .

mateuszbaran:
I can help a bit more if you show me the code that isn't working where this allocation is needed :slightly_smiling_face:

PR #226, opened 2025-04-07
Loosen Fiber type constraints

mateuszbaran:


src/ManifoldsBase.jl
```diff
@@ -77,6 +77,7 @@ allocate(a) = similar(a)
 allocate(a, dim1::Integer, dims::Integer...) = similar(a, dim1, dims...)
 allocate(a, dims::Tuple) = similar(a, dims)
 allocate(a, T::Type) = similar(a, T)
+allocate(::Number, T::Type) = fill(zero(T))
```
kellertuer, line 80: Is this deliberate that this would map an uncuttable number to a mutable 0-dim array?

kellertuer (approved):
I have one small question but even independent of that we can surely do this.

src/ManifoldsBase.jl
mateuszbaran: What's an uncattable number?

Anyway, yes, this line is deliberate -- when you try to allocate "something like a number", allocating a 0-dim array makes sense to me.

kellertuer: Damn it autocorrect on the phone ;) immutable.

You are right, that that is what one should get – just that we should still be careful that the immutable manifold thingies stay immutable in general (overwriting also `exp` not only `exp!` – as we indeed do).
So the only danger here is that someone expecting to “stay immutable” indeed does not.

PR #227, opened 2025-04-08
Relax VectorSpaceFiber type

mateuszbaran:
Without this some Manifolds.jl tests are failing.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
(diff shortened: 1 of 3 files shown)


kellertuer:
Interesting, that loosening a type made things fail.

kellertuer (approved):

mateuszbaran:
That's because I've also changed the default behavior a little: the default number system is now $\mathbb{R}$ instead of the one from the manifold.

kellertuer:
Hm, is that a good idea? I would have thought by default the one from the manifold is probably the best idea, at least for both TpM and T*pM?

mateuszbaran:
There are no good ideas because we don't have a consistent interpretation for the number system in `AbstractManifold`. The benefit of $\mathbb{R}$ is that real fibers always exist and that's what we almost always work with.

kellertuer:
Ok, that sounds reasonable.

PR #228, opened 2025-04-29
Stabilized retraction

mateuszbaran:
To address the issue from https://github.com/JuliaManifolds/Manopt.jl/pull/454 . IIRC there were some other similar issues reported previously.


.github/workflows/ci.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        julia-version: ["lts", "1", "pre"]
+        julia-version: ["lts", "1"]
```
kellertuer, line 14: why do we remove `"pre"`?

kellertuer (commented):
Nice test example, thanks for the docs.

Just two smaller points
* Maybe check that you cover all cases of mutating and allocating variants?
* And – since we are at it, we could do the same for vector transports as well?

.github/workflows/ci.yml
mateuszbaran: Something is broken again, check the failing test for the first commit.

mateuszbaran:
> * Maybe check that you cover all cases of mutating and allocating variants?

OK, I will add more tests

> * 
> * And – since we are at it, we could do the same for vector transports as well?

I don't recall any issues related to that so maybe let's add it when it turns out to be needed?

kellertuer:
Sure we can also only add it if we notice it is necessary for the VT

.github/workflows/ci.yml
kellertuer: *nirgs* DiffEq again.

Hm, can we maybe add them in a form that they are allowed to fail?

mateuszbaran: We had nightly tests running on a separate CI job before, maybe it should be brought back.

kellertuer (approved):
This by itself looks nice, but check whether you could add `"pre"` as a separate test step on CI that is allowed to fail. Can we maybe also report the DiffEq error?

.github/workflows/ci.yml
kellertuer: Yes maybe we could reactivate that and use pre there.

kellertuer:
Nice! Now the other tests run still when pre fails and the other tests still report code cov. Then this PR is good to go :)

PR #230, opened 2025-05-08
tangent_vector_type

mateuszbaran:
Required here: https://github.com/JuliaManifolds/LieGroups.jl/pull/37 .


kellertuer (approved):
The default might lead to a few strange errors for the non-implemented pairs, but I do not have a better idea how to provide a good other detaulf

PR #232, opened 2025-07-25
Switch to runic

kellertuer:
Also illustrates where runic is nicer in allowing multiple arguments in a line and still have multiline arguments.


kellertuer:
I already reformatter more lines to multiline args to reduce vertical spacing at it seems to me that this makes stuff more readable, when one can carefully group the arguments one selves; might work through more maybe on my travel back.

PR #254, opened 2026-02-24
Faster get_vector! on power manifolds

mateuszbaran:
I'm surprised I haven't encountered this issue sooner.


kellertuer (approved):
Interesting. That is even a point I understand in optimising the code ;)

LGTM.

PR #258, opened 2026-03-24
Faster `submanifold_component` on `ArrayPartition` with integer index

mateuszbaran:
It's really hard to go the Val-route in the GCD code, so let's make the Integer-route less punishing.


kellertuer (approved):
Sure. I trust you on this :)

PR #260, opened 2026-04-03
Move to a workspace approach and dependabot … bump RecursiveArrayTools

kellertuer:
Somehow the CompatHelper did not trigger here. But that is a good chance to move to dependant. I also introduced a Workspace for the Project.toml

I think the tests still run on RecursiveArrayTools 3 this way, since this needs to be registered and Manifolds.jl to be updated before that changes.


kellertuer:
I moved all test equipment into `ManifoldsBase.Test`, noticed that our macros automatically introduce two ambiguities that way, fixed that, adapted all tests and started a CI to avoid typos. I also fixed all typos 

I also separated all test dependencies from `[extras]` to a separate `Project.toml` and hopefully did not forget any compat entry.

Now this should hopefully be fine.

_edit:_ of course it is not. Would have been a surprise if any CI would act like running tests locally.

mateuszbaran (approved):
Thanks for taking care of this, LGTM 👍

kellertuer:
Takk! It was a bit of work, but mainly because we mixed 2-3 approaches to modularise the test helpers previously, thanks for the approval I'll go ahead and merge this and register it to then check how Manifolds.jl handles the new RecursiveArrayTools (since their breaking release has zero changelog of what broke).

kellertuer:
Let's just hope I did not miss a compat ;)

PR #270, opened 2026-05-25
Fix injectivity_radius documentation

sblelong:
The former version used the letter `R` that was not involved in the call `_injectivity_radius(M[, p], m::T)`.

As I understand, `m` is the `AbstractRetractionMethod` for which the injectivity radius is expected to be implemented.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [2.3.6] 25/05/2026
+## [2.3.6] unreleased
 
 ### Fixed
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [2.3.6] 25/05/2026
+
+### Fixed
+
+* documentation of the `injectivity_radius` function now displays clearer instructions for the implementation of a radius linked to a user's own `AbstractRetractionMethod`.
+
 ## [2.3.5] 03/04/2026
 
 ### Changed
--- a/NEWS.md
+++ b/NEWS.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-* (#270) documentation of the `injectivity_radius` function now displays clearer instructions for the implementation of a radius linked to a user's own `AbstractRetractionMethod`.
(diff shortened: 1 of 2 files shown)


sblelong:
Last thing to mention: in the Changelog, I switched from 2.3.5 to 2.3.6 to follow the pattern. I wonder if a doc fix deserves this changement in the numbering?

kellertuer:
Thanks for both spotting and fixing it :)

Concerning the Changelog:
* you could also but this PR in brackets behind ( so `(#270)` ) then it links to this PR in the rendered version of the changelog in the Documentation
* we could remove the date on the changelog entry and only fill that when we do a next version
* ...because for a next version you would also have to adapt the `Project.toml`

but I agree that this is neither a bug fix nor a new feature but merely a typo, so it could also just be included in the next version we do, whenever that is happening.
So I would prefer if you just remove the date and write `unreleased` there for now, so in the changelog we see, that this version is on master but for now “too small” to be registered yet.

Thanks again!

NEWS.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [2.3.6] 25/05/2026
```
kellertuer:
```suggestion
## [2.3.6] unreleased
```

NEWS.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [2.3.6] unreleased
+
+### Fixed
+
+* (#270) documentation of the `injectivity_radius` function now displays clearer instructions for the implementation of a radius linked to a user's own `AbstractRetractionMethod`.
```
kellertuer:
```suggestion
* documentation of the `injectivity_radius` function now displays clearer instructions for the implementation of a radius linked to a user's own `AbstractRetractionMethod`.  (#270) 
```

Sorry I was not precise enough where to put the reference here.

kellertuer: Also: No link necessary: Watching this rendered on GitHub makes it a link, in the docs we do that in processing the file anyways. Just this is fine.

sblelong: That is fine, you said "behind" and I realized that after the commit... so I fixed it in the following commit. I'll remove the link right away. Looking at the rendered file directly in Github did not display it as a link, that is why I added it manually.

sblelong:
It is my pleasure :)

Thank you for your suggestions. I think I've applied all of them!

kellertuer:
Nice!

Yeah we have this small function in the docs
https://github.com/JuliaManifolds/ManifoldsBase.jl/blob/3544e47ee7ceb950817628cb6134d8c41b3d6e24/docs/make.jl#L91-L104

which turns those marks (see the block after) into nice links when bringing the changelog to the docs. The reason is exactly, because on GitHub the current form is often turned into a link automatically, so we follow that idea and turn it into a link for the docs as well.

Thanks!

kellertuer (approved):

sblelong:
Interesting! I'll remember that for future contributions here or to related packages. Thank you for your help!

kellertuer:
Nice, and no worries. I know that for me this is all “how it is” – and has grown into this over a few years, while for you all of the stuff is new. So I am surely there to help and guide you on these specifics.

And it sounds great that you plan to do future contributions. Looking forward to that!

PR #272, opened 2026-06-01
Add a ManifoldsBaseMakie extension

kellertuer:
so that the checks, e.g. for retractions or geodesics – or the gradient in Manopt, can also be done with Makie instead of Plots.


mateuszbaran:
Judging by the error I think we need to use CairoMakie on CI instead of GLMakie.

kellertuer:
Nice, the concept and the test ideas even work. 

The remaining 4 lines are currently impossible to cover, since that method is overwritten (actually by both the Plots and the combined extension) which causes warnings anyways, so I have to change that a bit;
will try the idea that the plot functions (`plot_slope` and `plot_check_geodesic`) change their interface and have a `Val{Symbol}` as first parameter). This is fine, since that function is anyways an internal function.

kellertuer:
Uff, this is quite tricky to get right, that loading a plotting package it “just works” we do not have exponential growth in extensions and we do not overwrite existing methods. Somewhere I might still have a quadruple knot in my thoughts by now. But I tried for several hours to get it right.

Ah. Changing variables within therein package from the extension does not have a persistent effect, it is basically not set. Interesting to know.
Then I have to rethink the logic a bit, since it is then a bit harder to have in mind which of the extensions to use.

mateuszbaran (approved):

PR #280, opened 2026-07-03
Introduce `ZeroVector`

### LINKED ISSUE #279: Move ZeroVector from Manopt.jl
See here: https://github.com/JuliaManifolds/Manopt.jl/pull/617#discussion_r3466125793 .

kellertuer:
Here is a first draft to close #279 .

There are still quite a few `TODO`s in the `zero_vector.jl` file – mainly which functions to implement and which not.
For me `Riemann_tensor`, `sectional_curvature` and `angle` do not make much sense, but maybe one could “error more gracefully” there still.

The general modelling here is to do all these on “level 1” of our design scheme – for a few cases one could check whether level 2 makes more sense. A main challenge is probably to avoid ambiguities.

I also refactored the code a bit in places where all variables would fit on one line – since Runic allows for that while JuliaFormatter did not and put every variable on its own line.

Lets discuss also whether further methods are missing – and check whether it breaks something in Manifolds.jl or Manopt.jl.


kellertuer:
Documentation is also still missing, though there it is maybe enough to add the docs of the type to the vector space page.

kellertuer:
That is what I feared. A “straight-forward” plain approach introduces 200+ ambiguities. So we probably need something more clever than this PR.

mateuszbaran:
I fear properly adding it everywhere would require a very careful and time-consuming ambiguity hunting. I personally think `ZeroVector` has a fairly narrow utility and I would be fine with it not being supported by most (or even nearly all) of our functions. Solving all of these ambiguities will most likely not be worth the small gain it would give.

kellertuer:
That is why I stopped at that point yesterday. I am not sure which narrow utility you have in mind. Could you then continue here and minimise it again?

mateuszbaran:
So far there is exactly one example of actual utility, in Manopt.jl :smile: . To me it's more about "let's not make a separate `ZeroVector` type in case we notice a second example in a different package".

kellertuer:
That is very narrow 😄

But then lets document that carefully that it is clear that the zero vector does not (necessarily, automatically) work in “any tangent vector position”.

kellertuer:
I now added a careful technical note, that the type has to be considered on every manifold separately since it has to be implemented on layer 3, and kept those functions that do not cause ambiguities (`+, -, *, allocate`, and `zero_vector` with an optional `allocate=true` positional third argument).

src/vector_spaces.jl
```diff
@@ -134,3 +163,15 @@ vector_space_dimension(::AbstractManifold, ::VectorSpaceType)
...
+    zero_vector(M::AbstractManifold, p, allocate::Bool = true)
+
+Return the tangent vector from the tangent space ``T_p\mathcal M`` at `p` on the
+[`AbstractManifold`](@ref) `M`, that represents the zero vector.
+Setting `allocate = false` avoids allocation of memory for this vector
+and returns a [`ZeroVector`](@ref)`()`.
+"""
+function zero_vector(M::AbstractManifold, p, allocate::Bool)
+    return allocate ? zero_vector(M, p) : ZeroVector()
+end
```
mateuszbaran: I don't think this method is particularly useful so I'd prefer to not have it. It should be a design decision to use either the fully allocated zero vector or `ZeroVector`, not a switch like here.

src/vector_spaces.jl
```diff
@@ -102,15 +102,44 @@ manifold.
 """
 const AbstractCotangentVector = AbstractFibreVector{CotangentSpaceType}
 
+
+"""
+    ZeroVector <: AbstractTangentVector
+
+A type to represent the zero vector. This can be used generically
+when it is beneficial to not allocate memory in certain places of computations.
```
mateuszbaran:
```suggestion
A type to represent the zero vector. This can be used generically
when it is beneficial to statically dispatch on a zero vector.
```
I think this is a more important benefit.

src/vector_spaces.jl
kellertuer: Ok.

src/vector_spaces.jl
```diff
@@ -102,15 +102,44 @@ manifold.
...
+
+!!! note "Technical Note"
+    Following our [Design principles](design.md) the dispatch on tangent vectors
+    only happens on the most inner (third) layer, providing generic implementations
+    with this type is a very manifold-specific task and should be considered carefully.
+
+    Therefore there are also only a few generic implementations available, e.g. [`zero_vector`](@ref zero_vector(::AbstractManifold, ::Any, ::Bool))
```
mateuszbaran: It's actually `+`, `-`, `*`, `allocate` and `copy`, and I think out of these `allocate` isn't particularly useful.

mateuszbaran:
OK, everything else looks fine.

src/vector_spaces.jl
kellertuer: Yes this line should now be changed after you did not like the function 😄 

And that allocate was still there is more like random, I just removed all functions that caused ambiguities. I will adapt that line

kellertuer:
Should this also be a new version?

mateuszbaran (approved):
I think we can tag a new release after merging this.

NEWS.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [2.5.0] unreleased
+
+## Added
+
+* a `ZeroVector` type, to avoid allocating unnecessary zeros similar to the [`Identity`](https://juliamanifolds.github.io/LieGroups.jl/stable/interface/operations/#LieGroups.Identity) on [`LieGroups.jl`](https://github.com/JuliaManifolds/LieGroups.jl).
```
mateuszbaran:
```suggestion
* a `ZeroVector` type, to allow for static dispatch and avoid allocating unnecessary zeros similar to the [`Identity`](https://juliamanifolds.github.io/LieGroups.jl/stable/interface/operations/#LieGroups.Identity) on [`LieGroups.jl`](https://github.com/JuliaManifolds/LieGroups.jl).
```

kellertuer:
Great. I set the date in the news and will merge when the tests pass – and register after merge.

PR #287, opened 2026-09-02
Fix a small bug where we forgot to pass down the slope

kellertuer:
...and a bit of stability – in the checks for gradient/differential and their slope checks, which are actually here in ManifoldsBase.


mateuszbaran (approved):
OK, LGTM.

PR #290, opened 2026-09-09
Add Landing page and code blocks

kellertuer:
* add `DocumenterCodeBlocks.jl` plugin to the docs
* add a landing page with `DocumenterLandingPage.jl`
* extend runic to also check Markdown (md) and Quarto (qmd) files for proper code block formatting.
* introduce two new labels here to skip docs (`skip documentation`) or changelog (`no changelog necessary`) – that the dependabot also uses now, since it can not run docs and does not need a changelog.

and fix a few doc strings to adapt to code blocks. It still seems to not find `isapprox` and `norm` though I am loading the Julia interlinks. That one I could not figure out. Loading the Julia interlinks file resolved this for Manifolds.jl


mateuszbaran (approved):
LGTM

kellertuer:
Is this worth a new version or do we leave it on master until something new comes along?

mateuszbaran:
I don't have anything to do in ManifoldsBase.jl in the near future and I'd prefer to not have unreleased changes on master for a longer time. If you want to unleash Claude on ManifoldsBase.jl soonish then I think both things can be released together but if not then I think it's better to tag a release with just these changes.

kellertuer:
Fair enough.

Yes, since now in the semester I have a bit less time, I was thinking to do a bit of code review on our smallest package - so here - next.
And also apply the idea from this PR to LieGroups (but that is not much work actually).

Ok then we can combine this one with a bit of code review fixes.

PR #291, opened 2026-09-10
A(n AI assisted/annoyed) Code review

kellertuer:
I currently have the opportunity to get a bit annoyed by AI.

It at least finds a few loopholes and bugs along the way. 
This is the same approach already done in https://github.com/JuliaManifolds/Manopt.jl/pull/643, that I plan to do at least once (maybe not 3-4 times like there) here.

But I want to start even more transparent here, so I will post the current progress and the initial report already in this initial comment.

I did ask Claude to check the code for inaccuracies, typos, bugs – and to check the reports posted on the PR above for the general approach that it only writes reports and does not modify code itself – only if I specifically ask it to apply a change for example. That produced some 18k lines of a report – so I asked why it is so long and it reduced several self checks and so on.
That report is attached here for transparency.

From the first 22 (called severe) it claims one Manifold (Unitary) might fail due to the fix, since it implements an allocating variant somewhere. We should maybe check that – the fix proposed (and applied) seems correct to always first allocate – call the mutating variant – and only then pass from OGB to ONB.

The current status is

| severity | items | fixed | rejected | open | range |
|---|---|---|---|---|---|
| high | 22 | 22 | 0 | 0 | 1–22 |
| medium | 120 | 37 | 1 | 82 | 23–142 |
| low | 191 | 0 | 0 | 191 | 143–333 |

Fixed so far: 1–45, 47–60. Rejected: 46.

In my experience both medium and low are usually mainly small typos and are even faster to apply than the 22 severe noes (which now took me about 90 minutes to carefully read, assess, rephrase slightly and let them be applied).

[report_orig.md](https://github.com/user-attachments/files/32059844/report_orig.md)

### DIFF
--- a/src/EmbeddedManifold.jl
+++ b/src/EmbeddedManifold.jl
@@ -72,6 +72,13 @@ function project(M::EmbeddedManifold, p)
     return q
 end
 
+function project(M::EmbeddedManifold, p, X)
+    # the order of args switched, since the allocation by default takes the type of the first.
+    Y = allocate_result(M, project, X, p)
+    project!(M, Y, p, X)
+    return Y
+end
+
 function show(
         io::IO, M::EmbeddedManifold{𝔽, MT, NT},
     ) where {𝔽, MT <: AbstractManifold{𝔽}, NT <: AbstractManifold}
--- a/src/PowerManifold.jl
+++ b/src/PowerManifold.jl
@@ -155,43 +155,29 @@ const PowerManifoldNestedReplacing = AbstractPowerManifold{
 @inline _access_nested(x, i::Tuple) = x[i...]
 
 function Base.:^(
-        M::PowerManifold{
-            𝔽, TM, TSize, <:Union{NestedPowerRepresentation, NestedReplacingPowerRepresentation},
-        }, size::Integer...,
+        M::PowerManifold{𝔽, TM, TSize, <:Union{NestedPowerRepresentation, NestedReplacingPowerRepresentation}}, size::Integer...,
     ) where {𝔽, TM <: AbstractManifold{𝔽}, TSize}
     return PowerManifold(M, size...)
 end
 
 function allocate_on(
-        M::PowerManifold{
-            𝔽, TM, TSize, <:Union{NestedPowerRepresentation, NestedReplacingPowerRepresentation},
-        },
+        M::PowerManifold{𝔽, TM, TSize, <:Union{NestedPowerRepresentation, NestedReplacingPowerRepresentation}},
     ) where {𝔽, TM <: AbstractManifold{𝔽}, TSize}
     return [allocate_on(M.manifold) for _ in get_iterator(M)]
 end
 function allocate_on(
-        M::PowerManifold{
-            𝔽, TM, TSize, <:Union{NestedPowerRepresentation, NestedReplacingPowerRepresentation},
-        },
-        ::Type{<:Array{U}},
+        M::PowerManifold{𝔽, TM, TSize, <:Union{NestedPowerRepresentation, NestedReplacingPowerRepresentation}}, ::Type{<:Array{U}},
     ) where {𝔽, TM <: AbstractManifold{𝔽}, TSize, U}
     return [allocate_on(M.manifold, U) for _ in get_iterator(M)]
 end
 
 function allocate_on(
-        M::PowerManifold{
-            𝔽, TM, TSize, <:Union{NestedPowerRepresentation, NestedReplacingPowerRepresentation},
-        },
-        ft::TangentSpaceType,
+        M::PowerManifold{𝔽, TM, TSize, <:Union{NestedPowerRepresentation, NestedReplacingPowerRepresentation}}, ft::TangentSpaceType,
     ) where {𝔽, TM <: AbstractManifold{𝔽}, TSize}
     return [allocate_on(M.manifold, ft) for _ in get_iterator(M)]
 end
 function allocate_on(
-        M::PowerManifold{
-            𝔽, TM, TSize, <:Union{NestedPowerRepresentation, NestedReplacingPowerRepresentation},
-        },
-        ft::TangentSpaceType,
-        ::Type{<:Array{U}},
+        M::PowerManifold{𝔽, TM, TSize, <:Union{NestedPowerRepresentation, NestedReplacingPowerRepresentation}}, ft::TangentSpaceType, ::Type{<:Array{U}},
     ) where {𝔽, TM <: AbstractManifold{𝔽}, TSize, U}
     return [allocate_on(M.manifold, ft, U) for _ in get_iterator(M)]
 end
@@ -248,16 +234,14 @@ function allocate_result(M::PowerManifoldNestedReplacing, f, x...)
     end
 end
 function allocate_result_embedding(
-        ::PowerManifoldNestedReplacing,
-        ::typeof(project),
-        x,
-        args...
+        ::PowerManifoldNestedReplacing, ::typeof(project), x, args...,
     )
     return copy(x)
 end
 # the following is not used but necessary to avoid ambiguities
 function allocate_result(
-        M::PowerManifoldNestedReplacing, f::typeof(get_coordinates), p, X, B::AbstractBasis,
+        M::PowerManifoldNestedReplacing, f::typeof(get_coordinates), p, X,
+        B::AbstractBasis,
     )
     return invoke(
         allocate_result,
@@ -550,8 +534,7 @@ end
 
 @doc "$(_doc_distance_pow)"
 function distance(
-        M::AbstractPowerManifold,
-        p, q, m::AbstractInverseRetractionMethod, r::Real = 2,
+        M::AbstractPowerManifold, p, q, m::AbstractInverseRetractionMethod, r::Real = 2,
     )
     (isinf(r) && r > 0) && return _distance_max(M, p, q, m)
     (isinf(r) && r < 0) && return _distance_min(M, p, q, m)
@@ -771,8 +754,7 @@ end
 _get_field(::AbstractManifold{𝔽}) where {𝔽} = 𝔽
 
 function get_embedding(
-        M::PowerManifold{𝔽, TM, TSW, TPR},
-        P::Type,
+        M::PowerManifold{𝔽, TM, TSW, TPR}, P::Type,
     ) where {𝔽, TM <: AbstractManifold{𝔽}, TSW, TPR <: Union{NestedPowerRepresentation, NestedReplacingPowerRepresentation}}
     ME = get_embedding(M.manifold, eltype(P))
     return PowerManifold{_get_field(ME), typeof(ME), TSW, TPR}(ME, M.size)
@@ -1026,16 +1008,14 @@ retraction method has to be one that is available on the base [`AbstractManifold
 inverse_retract(::AbstractPowerManifold, ::Any...)
 
 function inverse_retract(
-        M::AbstractPowerManifold, p, q,
-        m::AbstractInverseRetractionMethod = default_inverse_retraction_method(M, typeof(p)),
+        M::AbstractPowerManifold, p, q, m::AbstractInverseRetractionMethod = default_inverse_retraction_method(M, typeof(p)),
     )
     X = allocate_result(M, inverse_retract, p, q)
     return inverse_retract!(M, X, p, q, m)
 end
 
 function inverse_retract!(
-        M::AbstractPowerManifold, X, p, q,
-        m::AbstractInverseRetractionMethod = default_inverse_retraction_method(M, typeof(p)),
+        M::AbstractPowerManifold, X, p, q, m::AbstractInverseRetractionMethod = default_inverse_retraction_method(M, typeof(p)),
     )
     rep_size = representation_size(M.manifold)
     for i in get_iterator(M)
@@ -1046,8 +1026,7 @@ function inverse_retract!(
     return X
 end
 function inverse_retract!(
-        M::PowerManifoldNestedReplacing, X, p, q,
-        m::AbstractInverseRetractionMethod = default_inverse_retraction_method(M, typeof(p)),
+        M::PowerManifoldNestedReplacing, X, p, q, m::AbstractInverseRetractionMethod = default_inverse_retraction_method(M, typeof(p)),
     )
     rep_size = representation_size(M.manifold)
     for i in get_iterator(M)
@@ -1380,8 +1359,7 @@ function retract(
 end
 
 function retract!(
-        M::AbstractPowerManifold, q, p, X,
-        m::AbstractRetractionMethod = default_retraction_method(M, typeof(p)),
+        M::AbstractPowerManifold, q, p, X, m::AbstractRetractionMethod = default_retraction_method(M, typeof(p)),
     )
     rep_size = representation_size(M.manifold)
     for i in get_iterator(M)
@@ -1402,8 +1380,7 @@ function retract!(
 end
 
 function retract_fused!(
-        M::AbstractPowerManifold, q, p, X, t::Number,
-        m::AbstractRetractionMethod = default_retraction_method(M, typeof(p)),
+        M::AbstractPowerManifold, q, p, X, t::Number, m::AbstractRetractionMethod = default_retraction_method(M, typeof(p)),
     )
     rep_size = representation_size(M.manifold)
     for i in get_iterator(M)
@@ -1414,8 +1391,7 @@ function retract_fused!(
     return q
 end
 function retract_fused!(
-        M::PowerManifoldNestedReplacing, q, p, X, t::Number,
-        m::AbstractRetractionMethod = default_retraction_method(M, typeof(p)),
+        M::PowerManifoldNestedReplacing, q, p, X, t::Number, m::AbstractRetractionMethod = default_retraction_method(M, typeof(p)),
     )
     rep_size = representation_size(M.manifold)
     for i in get_iterator(M)
@@ -1540,8 +1516,7 @@ Base.@propagate_inbounds function Base.setindex!(
 end
 
 function Base.show(
-        io::IO,
-        M::PowerManifold{𝔽, TM, TSize, TPR},
+        io::IO, M::PowerManifold{𝔽, TM, TSize, TPR},
     ) where {𝔽, TM <: AbstractManifold{𝔽}, TSize, TPR <: AbstractPowerRepresentation}
     size = get_parameter(M.size)
     return print(io, "PowerManifold($(M.manifold), $(TPR()), $(join(size, ", ")))")
@@ -1569,8 +1544,7 @@ function Base.show(
 end
 
 function vector_transport_direction!(
-        M::AbstractPowerManifold, Y, p, X, d,
-        m::AbstractVectorTransportMethod = default_vector_transport_method(M, typeof(p)),
+        M::AbstractPowerManifold, Y, p, X, d, m::AbstractVectorTransportMethod = default_vector_transport_method(M, typeof(p)),
     )
     rep_size = representation_size(M.manifold)
     for i in get_iterator(M)
@@ -1582,16 +1556,14 @@ function vector_transport_direction!(
     return Y
 end
 function vector_transport_direction(
-        M::AbstractPowerManifold, p, X, d,
-        m::AbstractVectorTransportMethod = default_vector_transport_method(M, typeof(p)),
+        M::AbstractPowerManifold, p, X, d, m::AbstractVectorTransportMethod = default_vector_transport_method(M, typeof(p)),
     )
     Y = allocate_result(M, vector_transport_direction, p, X, d)
     return vector_transport_direction!(M, Y, p, X, d, m)
 end
 
 function vector_transport_direction!(
-        M::PowerManifoldNestedReplacing, Y, p, X, d,
-        m::AbstractVectorTransportMethod = default_vector_transport_method(M, typeof(p)),
+        M::PowerManifoldNestedReplacing, Y, p, X, d, m::AbstractVectorTransportMethod = default_vector_transport_method(M, typeof(p)),
     )
     rep_size = representation_size(M.manifold)
     for i in get_iterator(M)
@@ -1602,8 +1574,7 @@ function vector_transport_direction!(
     return Y
 end
 function vector_transport_direction(
-        M::PowerManifoldNestedReplacing, p, X, d,
-        m::AbstractVectorTransportMethod = default_vector_transport_method(M, typeof(p)),
+        M::PowerManifoldNestedReplacing, p, X, d, m::AbstractVectorTransportMethod = default_vector_transport_method(M, typeof(p)),
     )
     Y = allocate_result(M, vector_transport_direction, p, X, d)
     rep_size = representation_size(M.manifold)
@@ -1631,15 +1602,13 @@ vector_transport_to(
     ::AbstractVectorTransportMethod,
 )
 function vector_transport_to(
-        M::AbstractPowerManifold, p, X, q,
-        m::AbstractVectorTransportMethod = default_vector_transport_method(M, typeof(p)),
+        M::AbstractPowerManifold, p, X, q, m::AbstractVectorTransportMethod = default_vector_transport_method(M, typeof(p)),
     )
     Y = allocate_result(M, vector_transport_to, p, X)
     return vector_transport_to!(M, Y, p, X, q, m)
 end
 function vector_transport_to!(
-        M::AbstractPowerManifold, Y, p, X, q,
-        m::AbstractVectorTransportMethod = default_vector_transport_method(M, typeof(p)),
+        M::AbstractPowerManifold, Y, p, X, q, m::AbstractVectorTransportMethod = default_vector_transport_method(M, typeof(p)),
     )
     rep_size = representation_size(M.manifold)
     for i in get_iterator(M)
@@ -1651,8 +1620,7 @@ function vector_transport_to!(
     return Y
 end
 function vector_transport_to!(
-        M::PowerManifoldNestedReplacing, Y, p, X, q,
-        m::AbstractVectorTransportMethod = default_vector_transport_method(M, typeof(p)),
+        M::PowerManifoldNestedReplacing, Y, p, X, q, m::AbstractVectorTransportMethod = default_vector_transport_method(M, typeof(p)),
     )
     rep_size = representation_size(M.manifold)
     for i in get_iterator(M)
--- a/src/PowerManifold.jl
+++ b/src/PowerManifold.jl
@@ -59,7 +59,7 @@ would not be represented by statically-sized arrays.
 
 # Constructor
 
-    PowerManifold(M::PowerManifold, N_1, N_2, ..., N_d; parameter::Symbol=:field)
+    PowerManifold(M::PowerManifold, N_1, N_2, ..., N_d; parameter::Symbol=_parameter_symbol(M))
     PowerManifold(M::AbstractManifold, NestedPowerRepresentation(), N_1, N_2, ..., N_d; parameter::Symbol=:field)
     M^(N_1, N_2, ..., N_d)
 
@@ -79,7 +79,8 @@ Since there is no default [`AbstractPowerRepresentation`](@ref) within this inte
 `^` operator is only available for `PowerManifold`s and concatenates dimensions.
 
 `parameter`: whether a type parameter should be used to store `n`. By default size
-is stored in a field. Value can either be `:field` or `:type`.
+is stored in a field, unless `M` is a [`PowerManifold`](@ref), then its storage is
+inherited. Value can either be `:field` or `:type`.
 """
 struct PowerManifold{𝔽, TM <: AbstractManifold{𝔽}, TSize, TPR <: AbstractPowerRepresentation} <:
     AbstractPowerManifold{𝔽, TM, TPR}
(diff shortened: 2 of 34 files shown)


kellertuer:
I got a bit lost in trying to fix what looks like a strange artefact, while trying to fix it, Claude was arguing for and against actually fixing it, changing its resulting point about 20 times. 

Here is the point of confusion I ran into and the reason for the half-fixed commit https://github.com/JuliaManifolds/ManifoldsBase.jl/pull/291/commits/5affccabb6058b54081782e7fafe03e5a57c0524

1. direct embedding claims it can directly embed without having to call embed (array stays the same like in sphere?) cf.

https://github.com/JuliaManifolds/ManifoldsBase.jl/blob/5affccabb6058b54081782e7fafe03e5a57c0524/src/decorator_trait.jl#L23-L28

2. In processing that type, it somehow gets “switched” on master (now fixed in the commit above) we have lines like

https://github.com/JuliaManifolds/ManifoldsBase.jl/blob/104566a1006ad84925d18fbb37e36def65d573f8/src/decorator_trait.jl#L906-L911

or (implicitly since the default is indirect: 

https://github.com/JuliaManifolds/ManifoldsBase.jl/blob/104566a1006ad84925d18fbb37e36def65d573f8/src/decorator_trait.jl#L900-L905

So these two exactly invert the binary logic? Why? I am lost what this is meant to mean, so I can also not fix it.

I am also surprised that in the tests we have something like 

https://github.com/JuliaManifolds/ManifoldsBase.jl/blob/5affccabb6058b54081782e7fafe03e5a57c0524/test/embedded_manifold.jl#L83-L93

So that states a direct embedding (no embed necessary to call) but still implements an embed.
How did all of this come along and what is the correct fix? I simply do not understand the current state and what it is supposed to mean. I do understand the doc string and that direct means direct. 
But the implementation seems to (a) switch happily between both and (b) treat direct as if it was indirect?

kellertuer:
I have a fix, but we should discuss whether this was a bug to switch the behaviour (in the code blocks above) or not.

mateuszbaran:
I'm 99% sure the embedding thing is just a bug, maybe we were a bit too tired when we did the rework. There should be no direct/indirect inversion and `AnotherHalfPlaneManifold` should be an indirect embedding manifold.

kellertuer:
For me it also looks like a bug that we just switch that. I just wanted to be sure and check, before continuing. Mainly because it is a bit technical and tricky ;)

kellertuer:
This now also includes JET after I fixed 5 issues it reported.
I can not understand JET error messages but yet again the LLM build proposed changes. Those are the 334-339 things that I will also add in the final report at the end of this PR.

src/ProductManifold.jl
```diff
@@ -1262,46 +1243,15 @@ submanifold_components(::Any...)
...
+function ziptuples(a::Tuple, b::Tuple, c::Tuple)
+    return ntuple(i -> (a[i], b[i], c[i]), min(length(a), length(b), length(c)))
+end
+function ziptuples(a::Tuple, b::Tuple, c::Tuple, d::Tuple)
+    n = min(length(a), length(b), length(c), length(d))
+    return ntuple(i -> (a[i], b[i], c[i], d[i]), n)
+end
+function ziptuples(a::Tuple, b::Tuple, c::Tuple, d::Tuple, e::Tuple)
+    n = min(length(a), length(b), length(c), length(d), length(e))
+    return ntuple(i -> (a[i], b[i], c[i], d[i], e[i]), n)
```
mateuszbaran, line 1256: What's the reason to change this? The old code was specifically designed to make sure it's fast. Maybe Julia is smarter now than back then but this needs a very careful verification with benchmarks if the change is to stay.

src/bases.jl
```diff
@@ -575,7 +573,10 @@ function _get_coordinates(M::AbstractManifold, p, X, B::DefaultOrthogonalBasis)
     return get_coordinates_orthogonal(M, p, X, number_system(B))
 end
 function get_coordinates_orthogonal(M::AbstractManifold, p, X, N)
-    return get_coordinates_orthonormal(M, p, X, N)
+    # arguments X and p for allocate_result are intentionally reversed
+    # to make ManifoldDiff.jl tests pass
+    c = allocate_result(M, get_coordinates, X, p, DefaultOrthogonalBasis(N))
+    return get_coordinates_orthogonal!(M, c, p, X, N)
 end
```
mateuszbaran, line 580: I thought this was intentionally done the old way to handle the `SArray` case easier where relevant (i.e. if someone implements `SArray`-based `get_coordinates` for a ONB, it's also picked when the default orthogonal basis is requested).

src/bases.jl
```diff
@@ -1006,12 +1014,12 @@ end
...
+        _show_basis_vector_range(io, Ξ, 1:nv; pre = pre, sym = sym)
     else
         halfn = div(max_vectors, 2)
-        _show_basis_vector_range(io, Ξ, 1:halfn; pre = "  ", sym = " E")
+        _show_basis_vector_range(io, Ξ, 1:halfn; pre = pre, sym = sym)
         print(io, "\n ⋮")
-        _show_basis_vector_range(io, Ξ, (nv - halfn + 1):nv; pre = "  ", sym = " E")
+        _show_basis_vector_range(io, Ξ, (nv - halfn + 1):nv; pre = pre, sym = sym)
     end
 end
```
mateuszbaran, line 1024: This looks half-reasonable: I think the defaults of `pre` and `sym` need to be adjusted to match previously used values (respectively `" "` and `" E"`).

src/decorator_trait.jl
```diff
@@ -268,17 +265,13 @@ end
...
-        M::AbstractManifold,
-        f::typeof(embed),
-        x::Vararg{Any, N},
+        M::AbstractManifold, f::typeof(embed), x::Vararg{Any, N},
     ) where {N}
-    T = allocate_result_type(get_embedding(M, typeof(x[1])), f, x)
-    return allocate(M, x[1], T, representation_size(get_embedding(M, typeof(x[1]))))
+    T = allocate_result_type(get_embedding(M, typeof(x[end])), f, x)
+    return allocate(M, x[1], T, representation_size(get_embedding(M, typeof(x[end]))))
 end
```
mateuszbaran, line 272: This is just messing with a pre-existing convention. `allocate_result_*` methods have rearranged order of arguments so that the first one is the main template.

src/EmbeddedManifold.jl
```diff
@@ -72,6 +72,13 @@ function project(M::EmbeddedManifold, p)
     return q
 end
 
+function project(M::EmbeddedManifold, p, X)
+    # the order of args switched, since the allocation by default takes the type of the first.
+    Y = allocate_result(M, project, X, p)
+    project!(M, Y, p, X)
+    return Y
+end
```
mateuszbaran, line 80: I'm surprised this method is necessary.

src/exp_log_geo.jl
```diff
@@ -188,7 +188,7 @@ Get a [`geodesic`](@ref) $γ_{p,q}(t)$ whose length is the shortest path between
 points `p`and `q`, where $γ_{p,q}(0)=p$ and $γ_{p,q}(1)=q$.
 When there are multiple shortest geodesics, a deterministic choice will be returned.
 
-This function returns a function of time, which may be a `Real` or an `AbstractVector`.
+This function returns a function of (time) `t`.
```
mateuszbaran, line 191: Was the old text wrong?

src/ProductManifold.jl
kellertuer: The LLM claims that on product manifolds something is not always known so then generated does not work, and it also claimed that without the generated on 1.10 and 1.13 it is of same speed.

But thanks for checking, I was very unsure about this myself. But keeping the generated in (the LLM claims) JET errors. I can check this later

Here is the report part (not yet included above – and sorry for posting LLM slop here)

> ### X 338. `ziptuples` compares tuple lengths while the code is being generated, which fails whenever a length is not yet known
> 
> `src/ProductManifold.jl:1265` · **low** · JET
> 
> The four `ziptuples` methods are generated functions that compute `min(N, M)` from the lengths in the
> tuple types. For a product manifold whose number of factors is not known, the lengths are not numbers
> and the comparison fails. At run time the lengths are always known, so no call fails; the analysis,
> which looks at the general case, reports five failures.
> 
> **Fix.** Plain `ntuple` calls give the same result and need no generated code: for tuples of known
> length `ntuple` with a constant count is unrolled by the compiler. No changelog entry.
> 
> **Check.** Both versions return the same tuples, are type-stable (`@inferred`) and allocate the same
> on Julia 1.13 and 1.10, measured on two and on five tuples. On a copy with the change the analysis
> gives no report and `test/product_manifold.jl` passes, 218 of 218; `runic --check` exits 0.
> 
> ```diff
> --- src/ProductManifold.jl:1264
>  """
> -@generated function ziptuples(a::NTuple{N, Any}, b::NTuple{M, Any}) where {N, M}
> -    ex = Expr(:tuple)
> -    for i in 1:min(N, M)
> -        push!(ex.args, :((a[$i], b[$i])))
> -    end
> -    return ex
> +ziptuples(a::Tuple, b::Tuple) = ntuple(i -> (a[i], b[i]), min(length(a), length(b)))
> +function ziptuples(a::Tuple, b::Tuple, c::Tuple)
> +    return ntuple(i -> (a[i], b[i], c[i]), min(length(a), length(b), length(c)))
>  end
> -@generated function ziptuples(
> -        a::NTuple{N, Any},
> -        b::NTuple{M, Any},
> -        c::NTuple{L, Any},
> -    ) where {N, M, L}
> -    ex = Expr(:tuple)
> -    for i in 1:min(N, M, L)
> -        push!(ex.args, :((a[$i], b[$i], c[$i])))
> -    end
> -    return ex
> +function ziptuples(a::Tuple, b::Tuple, c::Tuple, d::Tuple)
> +    n = min(length(a), length(b), length(c), length(d))
> +    return ntuple(i -> (a[i], b[i], c[i], d[i]), n)
>  end
> -@generated function ziptuples(
> -        a::NTuple{N, Any},
> -        b::NTuple{M, Any},
> -        c::NTuple{L, Any},
> -        d::NTuple{K, Any},
> -    ) where {N, M, L, K}
> -    ex = Expr(:tuple)
> -    for i in 1:min(N, M, L, K)
> -        push!(ex.args, :((a[$i], b[$i], c[$i], d[$i])))
> -    end
> -    return ex
> +function ziptuples(a::Tuple, b::Tuple, c::Tuple, d::Tuple, e::Tuple)
> +    n = min(length(a), length(b), length(c), length(d), length(e))
> +    return ntuple(i -> (a[i], b[i], c[i], d[i], e[i]), n)
>  end
> -@generated function ziptuples(
> -        a::NTuple{N, Any},
> -        b::NTuple{M, Any},
> -        c::NTuple{L, Any},
> -        d::NTuple{K, Any},
> -        e::NTuple{J, Any},
> -    ) where {N, M, L, K, J}
> -    ex = Expr(:tuple)
> -    for i in 1:min(N, M, L, K, J)
> -        push!(ex.args, :((a[$i], b[$i], c[$i], d[$i], e[$i])))
> -    end
> -    return ex
> -end
> ```

src/PowerManifold.jl
```diff
@@ -59,7 +59,7 @@ would not be represented by statically-sized arrays.
 
 # Constructor
 
-    PowerManifold(M::PowerManifold, N_1, N_2, ..., N_d; parameter::Symbol=:field)
+    PowerManifold(M::PowerManifold, N_1, N_2, ..., N_d; parameter::Symbol=_parameter_symbol(M))
```
mateuszbaran, line 62: That's wrong, and would be a bad breaking change if done.

src/bases.jl
kellertuer: I do not fully follow your argument? This switches to “allocate first/early” sure.

src/bases.jl
kellertuer: Sure we can adapt the defaults of the keywords :) I just noticed that passing them down is reasonable.

src/decorator_trait.jl
kellertuer: I have to check the report point for this. You are the allocation expert – so if you do not like this, you are more correct of course :)

src/PowerManifold.jl
kellertuer: I think for the docs – not that this is docs – it only documents what's in the code already?

src/ProductManifold.jl
kellertuer: It also first says one needs two implementations, then three, then reduced it to this.

src/EmbeddedManifold.jl
kellertuer: indeed I am surprised by now and maybe was not careful enough. Will check the item again.

Overall it is quite challenging to go through 300+ items and keep the concentration hold up ;)

src/ProductManifold.jl
mateuszbaran: Yeah, LLMs need to be checked.

We don't support anything like "product manifold whose number of factors is not known". IIRC there were some instances where my `ziptuples` could fail but they were just some odd corner cases or maybe even compiler bugs. I'd suggest just reverting it unless the LLM can generate a concrete counterexample.

kellertuer: I think that here is JET erroring actually.

src/bases.jl
mateuszbaran: I've re-checked it and we don't really make use of it anyway right now so it can stay.

src/PowerManifold.jl
mateuszbaran: Ah, it's a mess after re-checking. I didn't notice the `::PowerManifold` in this line, in which case this is correct, but then this docstring is missing the `:field`-defaulting constructor with arbitrary base manifolds which arguably should be first on the list here as it's the one most people would need, not the power manifold of a power manifold special case.

kellertuer: The one for Manifold and :field is line 63.

I also had to check – so for power manifolds we can inherit the field, for manifolds we don't

src/retractions.jl
```diff
@@ -78,7 +78,7 @@ for point and tangent vectors.
 
 !!! note "Technical Note"
     Though you would call e.g. [`retract`](@ref)`(M, p, X, PolarRetraction())`,
-    to implement a polar retraction, define [`retract_polar!`](@ref)`(M, q, p, X, t)`
```
mateuszbaran: `t` was mentioned because then the code can be faster in some cases though I don't remember if the variant without `t` also needs to be implemented in that case.

src/ProductManifold.jl
kellertuer: So the actual error JET complains about in code analysis is

```
 ManifoldsBase.ziptuples(::Tuple, ::Tuple, ::Tuple)
: ManifoldsBase.ziptuples(::Tuple, ::Tuple, ::Tuple)
││││││└────────────────────
│││││┌ get_vectors(M::ProductManifold, p::ArrayPartition, B::CachedBasis{…}) where 𝔽 @ ManifoldsBaseRecursiveArrayToolsExt ManifoldsBase.jl/ext/ManifoldsBaseRecursiveArrayToolsExt/ProductManifoldRecursiveArrayToolsExt.jl:144
││││││┌ ziptuples(a::Tuple, b::Tuple) @ ManifoldsBase ManifoldsBase.jl/src/ProductManifold.jl:1300
│││││││ MethodError: no method matching isless(::Core.TypeofVararg, ::Core.TypeofVararg)
The function `isless` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  isless(!Matched::Missing, ::Any)
   @ Base missing.jl:87
  isless(::Any, !Matched::Missing)
   @ Base missing.jl:88
  isless(!Matched::Base.BinaryPlatforms.CPUID.ISA, !Matched::Base.BinaryPlatforms.CPUID.ISA)
   @ Base cpuid.jl:22
  ...
```

which is within the `min` I think.

src/retractions.jl
kellertuer: but that would be the `_fused` variant. We split that ;)

src/test_suite.jl
```diff
@@ -192,8 +200,8 @@ function Random.rand!(
     return pX
 end
 ManifoldsBase.representation_size(::TestSphere{N}) where {N} = (N + 1,)
-function ManifoldsBase.retract_project_fused!(M::TestSphere, q, p, X, t::Number)
-    q .= p .+ t .* X
+function ManifoldsBase.retract_project!(M::TestSphere, q, p, X; kwargs...)
+    q .= p .+ X
```
mateuszbaran: The old method is fine as it tests that we can just overwrite the fused variant.

src/bases.jl
kellertuer: I fixed this in https://github.com/JuliaManifolds/ManifoldsBase.jl/pull/291/commits/4e62fdce326f18082217493e81563fb3a5657f21

src/retractions.jl
mateuszbaran: Good point, I forgot we split it.

mateuszbaran: Anyway I'd actually recommend implementing the `fused` variant.

src/ProductManifold.jl
kellertuer: Here is a proposal. Since the ziptuples are unhappty for generated with nonInts, we could for example add a safeguard as in https://github.com/JuliaManifolds/ManifoldsBase.jl/pull/291/commits/84aed5d79e0406c1421e366daed4f70dde33e871

what do you think? This at least makes JET happy

src/decorator_trait.jl
kellertuer: ah right, I checked and will revert this; because yes we do switch the order every now and then but it is much easier to check the first here.

src/EmbeddedManifold.jl
kellertuer: If you look into the code, it also has the project(M, p) before with a similar code.

The analysis by the LLM claimed (and I verified):

```
using ManifoldsBase
using ManifoldsBase: DefaultManifold
E = EmbeddedManifold(DefaultManifold(2), DefaultManifold(3))
ManifoldsBase.project!(::typeof(E), q, p) = copyto!(q, p[1:2])
ManifoldsBase.project!(::typeof(E), Y, p, X) = copyto!(Y, X[1:2])
p = [1.0, 2.0, 3.0]
X = [4.0, 5.0, 6.0]
project(E, p)     # [1.0, 2.0]
project(E, p, X)  # [4.0, 5.0]
```

The second to last line was like that all the time due to the overwriting before, the last has `[4.0, 5.0, 0.0]` (or even initialised in the last) which is not on D2.

The reason is that by default project “passes through” a decorator like e.g. th Metric, but here it should not. I think this fix is correct, especially with the other project directly in front.

src/retractions.jl
kellertuer: yes we should recommend implementing the fused variant here, correct. Will fix.

src/ProductManifold.jl
mateuszbaran: The `N isa Int` solution is acceptable to me but I've checked with Claude and it looks like it's JET's false positive.

src/test_suite.jl
kellertuer: The LLM says (item 92)

```
`TestSphere` implements only the fused projection retraction, so `retract(M, p, X, ProjectionRetraction())` is a `MethodError`
```

and I am never sure which one falls back to what.

src/retractions.jl
mateuszbaran: I think recommending the fused variant would be preferable everywhere.

kellertuer: ah, I also always forget which one falls back to what :D

kellertuer: But for now I fear the fallback is the other way around.

* retract fused by default calls retract
* retract is the one you have to implement first (ease of use easier to vconvey)
* retract fused implements is for speed / faster computations

So the first to implement is always retract not the fused one

src/test_suite.jl
kellertuer: See explanation on the last point. retract is the first to implement, fused always only the additional “speedup” variant and it currently “falls back” to retract by default. So this fix makes both variants work, while only implementing fused makes the non-fused one stay not-implemented.

mateuszbaran: `TestSphere` works on `master` branch with only `_fused` defined but maybe we need both implemented for some non-tested cases to work, I'm not sure.

src/ProductManifold.jl
kellertuer: Maybe it is,m but I am not sure how to fix JET.

src/EmbeddedManifold.jl
kellertuer: Ok, will add :)

src/retractions.jl
mateuszbaran: I see, then I guess ideally both `fused` and non-`fused` should ideally be implemented.

src/test_suite.jl
kellertuer: yes, currently retract alone is not tested, that is why it works on master. Will check and add a test.

In short:
only implementing retract makes both work (fused falls back to maybe an inefficient default)
implementing fused with a more efficient add-on is meant to be the improvement.

I think I argued for that back then to say “less technical people start with just implementing retract so that route should be the default”?

src/retractions.jl
kellertuer: yes for best performance that is the case. I can add that to all the docs and mention both.

Maybe even in the form: Implementing retract! “makes things work” implementing fused “potentially makes things fast”

I just have to think how to best phrase that.

src/EmbeddedManifold.jl
kellertuer: https://github.com/JuliaManifolds/ManifoldsBase.jl/pull/291/commits/3803b91fa907845ccdf8941790cb8ebb26a2db56

src/test_suite.jl
mateuszbaran: I see, OK.

src/exp_log_geo.jl
kellertuer: This is the case with 3 arguments `shortest_geodesic` which after a log passes to

```
geodesic(M::AbstractManifold, p, X) = t -> exp_fused(M, p, X, t)
```

that only accepts a real (time) t – no longer vectors (to evaluate point wise) or an integer number to sample. So the abstract vector does not work.
For geodesic this is documented the same.

And it only affects this function-returned case of course.

src/retractions.jl
kellertuer: I also added doc strings to the internal functions to be implemented, so they can be linked, see

https://github.com/JuliaManifolds/ManifoldsBase.jl/pull/291/commits/1494ebe663f293527045df6382684a1481b31688

mateuszbaran: Nice, I think this is good now.

src/ProductManifold.jl
```diff
@@ -982,10 +938,13 @@ function sectional_curvature(M::ProductManifold, p, X, Y)
...
         if are_linearly_independent(M_i, p_i, X_i, Y_i)
-            curvature += sectional_curvature(M_i, p_i, X_i, Y_i)
+            w_i =
+                inner(M_i, p_i, X_i, X_i) * inner(M_i, p_i, Y_i, Y_i) -
+                inner(M_i, p_i, X_i, Y_i)^2
+            curvature += w_i * sectional_curvature(M_i, p_i, X_i, Y_i)
         end
     end
-    return curvature
+    return curvature / (inner(M, p, X, X) * inner(M, p, Y, Y) - inner(M, p, X, Y)^2)
```
mateuszbaran: This is a good catch but I think power manifold has the same bug (and the same fix should be applied).

kellertuer: Oh, indeed. I think the LLM did not spot that, since I am already in the “low” priority ones and if I remember correctly, this one was medium. Will check tomorrow then.

mateuszbaran:
I think this looks good up to fixing coverage and making sure Manifolds.jl/Manopt.jl still work fine.
Currently at least one Manifolds.jl test fails. Maybe AI can figure it out :wink: .
```
Forwarding: Error During Test at /home/mateusz/.julia/dev/Manifolds/test/test_statistics.jl:525
  Test threw exception
  Expression: mean(M, [p1, p2]) == mean([p1, p2])
  MethodError: no method matching var"_Statistics.mean_forwarding"(::EmbeddedForwardingType{ManifoldsBase.IndirectEmbedding}, ::CenteredMatrices{ℝ, ManifoldsBase.TypeParameter{Tuple{3, 2}}}, ::Vector{Matrix{Float64}})
  The function `_Statistics.mean_forwarding` exists, but no method is defined for this combination of argument types.
  Closest candidates are:
    var"_Statistics.mean_forwarding"(::EmbeddedForwardingType{ManifoldsBase.DirectEmbedding}, ::AbstractDecoratorManifold, ::AbstractVector)
     @ Manifolds ~/.julia/dev/ManifoldsBase/src/nested_trait.jl:157
    var"_Statistics.mean_forwarding"(::SimpleForwardingType, ::AbstractDecoratorManifold, ::AbstractVector)
     @ Manifolds ~/.julia/dev/ManifoldsBase/src/nested_trait.jl:175
    var"_Statistics.mean_forwarding"(::StopForwardingType, ::AbstractDecoratorManifold, ::AbstractVector)
     @ Manifolds ~/.julia/dev/ManifoldsBase/src/nested_trait.jl:193
  Stacktrace:
   [1] mean(M::CenteredMatrices{ℝ, ManifoldsBase.TypeParameter{Tuple{3, 2}}}, x::Vector{Matrix{Float64}})
     @ Manifolds ~/.julia/dev/ManifoldsBase/src/nested_trait.jl:151
   [2] top-level scope
     @ ~/.julia/dev/Manifolds/test/test_statistics.jl:351
   [3] macro expansion
     @ ~/.julia/juliaup/julia-1.13.0+0.x64.linux.gnu/share/julia/stdlib/v1.13/Test/src/Test.jl:1962 [inlined]
   [4] macro expansion
     @ ~/.julia/dev/Manifolds/test/test_statistics.jl:522 [inlined]
   [5] macro expansion
     @ ~/.julia/juliaup/julia-1.13.0+0.x64.linux.gnu/share/julia/stdlib/v1.13/Test/src/Test.jl:1962 [inlined]
   [6] macro expansion
     @ ~/.julia/dev/Manifolds/test/test_statistics.jl:525 [inlined]
   [7] macro expansion
     @ ~/.julia/juliaup/julia-1.13.0+0.x64.linux.gnu/share/julia/stdlib/v1.13/Test/src/Test.jl:750 [inlined]
 20.549594 seconds (68.17 M allocations: 3.140 GiB, 1.71% gc time, 86.47% compilation time)

```

kellertuer:
Oh, I did not expect that. After today I first have to take a break from LLMs. They are exhausting.
But good to know, I will keep that in mind to fix.

kellertuer:
On Mac OS with the master before your newest version, the statistics tests run fine (Julia 1.13).

Reran them after getting the newest master branch [Manifolds 0.11.30), but also there `test_statistics` ran just fine.

Also the centred matrices that are the manifold mentioned run fine

```
[ Info: Running test_atlases.jl
324.855982 seconds (809.05 M allocations: 43.325 GiB, 15.38% gc time, 98.17% compilation time)
[ Info: Running approx_inverse_retraction.jl
  3.911331 seconds (7.03 M allocations: 387.971 MiB, 76.18% compilation time)
[ Info: Running manifolds-old/embedded_torus.jl
 93.493069 seconds (507.35 M allocations: 27.650 GiB, 9.60% gc time, 95.92% compilation time: 1% of which was recompilation)
[ Info: Running manifolds/test_centered_matrices.jl
 10.810270 seconds (49.34 M allocations: 2.849 GiB, 8.52% gc time, 99.68% compilation time)
```
(though atlases take a bit long-ish)

Overall I get

```
Test Summary: |  Pass  Broken  Total      Time
Manifolds.jl  | 21084       1  21085  27m07.7s
     Testing Manifolds tests passed 
```

(and I even have a next analysis that finds the reason and fixes the broken test)

So without reproducing the error locally I would first need more details on how you got this error.

mateuszbaran:
I've found the cause of the error but I have no idea how it could work for you, other than by not using this branch of ManifoldsBase.jl. MWE is
```
M = CenteredMatrices(3, 2)
p1 = [1.0 2; 4 5; -5 -7]
p2 = [3.0 1; 2 5; -5 -6]
mean(M, [p1, p2]) # this call fails
```

kellertuer:
I also have no clue why that test then ran fine I really just opened Julia 1.13 and ran the tests.

But I can confirm your MWE fails for me, I just have no clue why, but it might indeed be due to the switch in the embedding/forwarding at item 60 where I asked above. I think not many use this but the centered matrices indeed might.

mateuszbaran:
I guess your tests were running from a Julia env that didn't have `ManifoldsBase.jl` `] dev`-ed. For the future, the simplest way of making sure you're in the right environment is activating it manually and then doing `include("runtests.jl")`. Using any automation like `] test` may run in a different environment.

kellertuer:
Well, my current way: All packages are dev'ed in my main environment. I ran the example above in that and I ran the tests in that. For the last 10+ years that worked well.

kellertuer:
It is how statistics functions in Manifoldds.jl do the same “mixup” as I described above with code (cf. item 60 in the report).
So the fix here make the bug in Manifolds.jl appear. I can do a PR there as well that should fix it then.

mateuszbaran:
I've already made a PR there: https://github.com/JuliaManifolds/Manifolds.jl/pull/930 :slightly_smiling_face:

kellertuer:
This one is finished up to two DocumenterCodeBlock warnings in the docs rendering. These maybe can not be avoided, since it is within code of the tutorial rendered into the docs where the code calls `Base.approx` and `LinearAlgebra.norm` from Base and LA, respectively – so that is not easy to link I think?

I will check the changelog one last time tomorrow – since I have not read that one since the start – and then propose to merge and register here.

Here is already the final version of the report/slop

[report.md](https://github.com/user-attachments/files/32616534/report.md)

src/ProductManifold.jl
kellertuer: Oh I nearly forgot this. Will check next time I work on this :)

mateuszbaran (approved):
LGTM (up to the one line that lost coverage).

We might also add the "tested with JET" badge here:

```
[![](https://img.shields.io/badge/%F0%9F%9B%A9%EF%B8%8F_tested_with-JET.jl-233f9a)](https://github.com/aviatesk/JET.jl)
```

kellertuer:
Good point, will add; and sure coverage I will check, too.

